PR #11, opened 2022-07-04
Building ncdu for testing purpose

kcubeterm:


kcubeterm:
Just for testing purpose. See if build system is able to find dependencies or not

PR #35, opened 2022-07-18
New package: termux-install

KrutosVIP:
This is my first package on Termux.

I wanted to make life easier with this package, but I dont have my own public Repo


kcubeterm:
`TERMUX_PKG_PLATFORM_INDEPENDENT=true` 
You can add, since all files are nothing but bash scripts that could run on all architecture wherever bash is available

PR #46, opened 2022-08-05
new package: ndk-sysroot-gcc-compact, gcc-11

licy183:
Tested locally on all architectures with some hello-world codes and it seems working fine.


kcubeterm:
Nice work!

LqkUWp:
When installing ndk-sysroot-gcc-compact, ndk-sysroot is auto removed. But dpkg complains that clang depends on ndk-sysroot. Is that fine? is ndk-sysroot-gcc-compact a superset of ndk-sysroot? does ndk-sysroot-gcc-compact support clang?

licy183:
> When installing ndk-sysroot-gcc-compact, ndk-sysroot is auto removed. But dpkg complains that clang depends on ndk-sysroot. Is that fine? is ndk-sysroot-gcc-compact a superset of ndk-sysroot? does ndk-sysroot-gcc-compact support clang?

Yes. As you point out, `ndk-sysroot-gcc-compact` is a superset of `ndk-sysroot`, so `ndk-sysroot-gcc-compact` is compatible with both `clang` and `gcc`. Actually, `ndk-sysroot-gcc-compact` defines some macros that are clang-specified and revert some commits that break the compiling process of gcc. If you find any bug, feel free to report it.

PR #54, opened 2022-08-11
New package: arpack-ng

ifurther:


T-Dynamos:
```
22-08-13T00:46:31.7586771Z   Tell CMake where to find the compiler by setting either the environment
2022-08-13T00:46:31.7587129Z   variable "FC" or the CMake cache entry CMAKE_Fortran_COMPILER to the full
2022-08-13T00:46:31.7587467Z   path to the compiler, or to the compiler name if it is in the PATH.
```
set `FC` or setting `PATH`

licy183:
It seems that a toolchain with fortran support is needed to compile this package. I will move the host-build step in gcc, which compiles a gcc toolchain with fortran support, to a new git repository if I have some time. Or try to use android-gfortran https://github.com/buffer51/android-gfortran.

ifurther:
Maybe it should need the new repository

IntinteDAO:
@licy183 

> It seems that a toolchain with fortran support is needed to compile this package.

Isn't flang a Fortran Compiler? Or it is not enough?

licy183:
> Isn't flang a Fortran Compiler? Or it is not enough?

`flang` from `llvm-project` cannot be used for cross-compiling currently, as it cannot produce `PIC (position-independent code)` but `PIC` is always needed for Android 7.0. `classic-flang` is also difficult to cross-compile, as it doesn't support cross-compile natively.

licy183:
@ifurther Hello, please test whether this package works fine if it is convenient for you. Thanks!

PR #57, opened 2022-08-14
Gitmodule

ifurther:
add run-docker and git submodule , it could be easy run in local using docker container.

### DIFF
--- a/.gitignore
+++ b/.gitignore
@@ -1,16 +1,16 @@
 # These files are from termux/termux-packages
-/termux-packages/
-/disabled-packages/
-/ndk-patches/
-/packages/
-/root-packages/
-/scripts/
-/x11-packages/
-/CODEOWNERS
-/CONTRIBUTING.md
-/build-all.sh
-/build-package.sh
-/clean.sh
+#/termux-packages/
+#/disabled-packages/
+#/ndk-patches/
+#/packages/
+#/root-packages/
+#/scripts/
+#/x11-packages/
+#/CODEOWNERS
+#/CONTRIBUTING.md
+#/build-all.sh
+#/build-package.sh
+#/clean.sh
 
 # Vim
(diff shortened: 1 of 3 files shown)


kcubeterm:
Basically we dont need submodule. our starting approaches was the same but after that we realised it will add extra overhead of updating submodule and all. instead we choose `setup-environment.sh` which do the same job.  and in case we need to override the Build system variable, termuxrc is always there. 

any good reason to have git submodule over `setup-environment.sh` ?

ifurther:
Maybe doesn't use git submodule, but the `setup-environment.sh` sometimes doesn't update build-package.sh.
Another thing is termuxrc setting in `setup-environment.sh` should check in build container

licy183:
Actually, I still don't understand where the necessity of git-submodule is. If we use git-submodule, we will need to manually update it, and I think it is not necessary.

> Maybe doesn't use git submodule, but the `setup-environment.sh` sometimes doesn't update build-package.sh.

I'm not sure why `build-package.sh` cannot be updated with `setup-environment.sh`, so could you please tell me how to reproduce it?

> Another thing is termuxrc setting in `setup-environment.sh` should check in build container, so the `run-docker.sh` is simply bind the 'termuxrc` in container.

You could update `.termuxrc` in the container manually if you need to change it. Or I think we could just override the `termuxrc` file without checking the old one in `setup-environment.sh`.

ifurther:
If `setup-environment.sh` move the termux-package dir to tur dir , the termux-package git is not upgrade.

licy183:
> If `setup-environment.sh` move the termux-package dir to tur dir , the termux-package git is not upgrade.

I'm not sure why this happens, and I can't reproduce it on my machine... I think `git reset --hard origin/master` then `git pull --rebase` can upgrade termux-packages successfully...

ifurther:
let me try

PR #69, opened 2022-09-04
feat: add tur to package list

2096779623:
This way we can save some time and download the package directly from tur instead of building package.

https://github.com/2096779623/tur/runs/8174124791?check_suite_focus=true#step:7:722


licy183:
I think this issue has been fixed by https://github.com/termux-user-repository/tur/commit/0f0fd4391b4f7f835bb358ae3358762e5337571f. Anyway, thanks for reporting!

PR #88, opened 2022-09-21
new package: sqawk

flosnvjx:
~~tcllib is the standard library of tcl.  Some module like `ftpd` use hardcoded path (we have busybox anyway).~~

sqawk is a maintained CLI utility depends on tcllib -- have to say that only [a few](https://github.com/search?q=language%3Atcl+stars%3A%3E50+pushed%3A%3E2020-01-01&ref=simplesearch) maintained project on GitHub still use tcl, too few to create a `termux_setup_tcl` for it.


licy183:
Thanks for your contribution! How about adding `tcllib` to the main repository? I don't think it breaks the package policy of termux-packages.

flosnvjx:
I'll try to send pr to upstream for tcllib, will take some time for patch the hardcoded patch.

flosnvjx:
tcllib PR to upstream has been sent: https://github.com/termux/termux-packages/pull/12055

PR #143, opened 2022-11-08
chore(hacking): remove some hacking packages to another repository

licy183:


2096779623:
Also need update https://github.com/termux/termux-packages/blob/master/packages/tur-repo/build.sh ?

licy183:
> Also need update https://github.com/termux/termux-packages/blob/master/packages/tur-repo/build.sh ?

I don't think we need to update it. Actually, one reason to split packages to another TUR repository is to make it hard to install hacking-related packages.

kcubeterm:
how will one install hacking packages then ?

licy183:
> how will one install hacking packages then ?

Someone must manually add `tur-hacking` to `$PREFIX/etc/apt/sources.list.d/tur.list`. Should we update `README`?

kcubeterm:
if the point is to make installing hard such packages for skids. then its seems good, then why even add that in `README` then.  Keep it secret :)

PR #144, opened 2022-11-08
code-server : update to 4.8.3

T-Dynamos:


T-Dynamos:
@licy183 can't we enable auto update?

licy183:
> @licy183 can't we enable auto update?

It's hard. We don't use the git repository but we use the source tarball to build. We can update coder-server's version but hard to update the one of vscode's.

PR #162, opened 2022-11-24
arpack-ng use termux-repo's libopenblas

amuramatsu:


licy183:
The build for `x86_64` seems a bug of the gcc toolchain... I will try to solve it when I have more free time. At this time, you could disable the arch for `x86_64`.

licy183:
Thanks!

PR #165, opened 2022-11-26
add sundials 6.4.1 for octave

amuramatsu:


licy183:
When building `sundials` for 32-bit platforms, some warnings occurs.
```
[8/520] Building C object src/sundials/CMakeFiles/sundials_generic_obj_static.dir/sundials_logger.c.o
In file included from /home/builder/.termux-build/sundials/src/src/sundials/sundials_logger_impl.h:24,
                 from /home/builder/.termux-build/sundials/src/src/sundials/sundials_logger.c:26:
/home/builder/.termux-build/sundials/src/src/sundials/sundials_hashmap.h:26:48: warning: conversion from 'long long unsigned int' to 'long unsigned int' changes value from '14695981039346656037' to '2216829733' [-Woverflow]
   26 | static const unsigned long HASH_PRIME        = 14695981039346656037U;
      |                                                ^~~~~~~~~~~~~~~~~~~~~
/home/builder/.termux-build/sundials/src/src/sundials/sundials_hashmap.h:27:48: warning: conversion from 'long long unsigned int' to 'long unsigned int' changes value from '1099511628211' to '435' [-Woverflow]
   27 | static const unsigned long HASH_OFFSET_BASIS = 1099511628211U;
      |                                                ^~~~~~~~~~~~~~
```
It seems that `sundials` cannot work on 32-bit platforms. I'll disable it on `i686` and `arm`.

licy183:
Re-enable in 960c801a6d3cf173ae1bcd1ac7150fe2bbd54f2f. But I'm not sure this library could run successfully. Seems that its developers think that `long int` is `int64_t` but on 32-bit Android `long int` is actually `int32_t`. Maybe it should be reported upstream.

PR #182, opened 2022-12-10
Electron

licy183:


kawanakaiku:
Running the process exits with this error.```[10536:1213/141806.469205:ERROR:platform_shared_memory_region_posix.cc(220)] Unable to access(W_OK|X_OK) /dev/shm: No such file or directory (2)```
It may need one more argument ```--disable-dev-shm-usage```.

licy183:
Forget to patch this. Will fix later. Thanks for your report.

PR #254, opened 2023-02-05
biber: move from main repo

Grimler91:
Since the package contains just an installation script rather than pre-compiled binaries I think it fits better in the user repository.

Biber depends on perl-net-ssleay, so by moving it to tur we can depend on the already packaged perl-net-ssleay package instead of building it from source in installation script.


Grimler91:
Seems dependency on packages in tur-on-device is not allowed (similar to how packages in main repo can't depend on packages from x11 repo for example), so I'll move the package to tur-on-device

licy183:
Emmm... Seems that it is time to merge `tur-on-device` in this repository. I'll work on it in the following days.

PR #287, opened 2023-02-17
mesa-zink: migrate to llvm-15

licy183:


licy183:
Tested on Adreno (TM) 660.

PR #307, opened 2023-03-01
Use rust nightly from release 

Ludea:
Use rust official release instead of src date based.

It can lead to issue between rustc and rust src if they haven't pulled from GH.

### DIFF
--- a/tur/rustc-nightly/build.sh
+++ b/tur/rustc-nightly/build.sh
@@ -2,6 +2,7 @@ TERMUX_PKG_HOMEPAGE=https://www.rust-lang.org
 TERMUX_PKG_DESCRIPTION="Rust compiler and utilities (nightly version)"
 TERMUX_PKG_DEPENDS="libc++, clang, openssl, lld, zlib, libllvm"
 TERMUX_PKG_LICENSE="MIT"
+TERMUX_PKG_MAINTAINER="@termux-user-repository"
 TERMUX_PKG_REVISION=2
 _DATE="2023-02-27"
 TERMUX_PKG_VERSION="1.67.1-${_DATE//-/.}-nightly"
--- a/tur/rustc-nightly/build.sh
+++ b/tur/rustc-nightly/build.sh
@@ -3,10 +3,10 @@ TERMUX_PKG_DESCRIPTION="Rust compiler and utilities (nightly version)"
 TERMUX_PKG_DEPENDS="libc++, clang, openssl, lld, zlib, libllvm"
 TERMUX_PKG_LICENSE="MIT"
 TERMUX_PKG_REVISION=2
-TERMUX_PKG_MAINTAINER="@termux-user-repository"
-TERMUX_PKG_VERSION="1.67.1-nightly"
-TERMUX_PKG_SRCURL=https://static.rust-lang.org/dist/rustc-nightly-src.tar.xz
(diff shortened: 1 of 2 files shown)


licy183:
Emmm... I don't think this is a good idea. In AUR, [`rust-nightly`](https://aur.archlinux.org/cgit/aur.git/tree/PKGBUILD?h=rust-nightly-bin) is built on your local machine and you can always keep it up to date. But TUR is not like AUR, this package is pre-built using GitHub Action, so I don't think the date should be removed.

licy183:
I don't know what you really want to do. Actually, `https://static.rust-lang.org/dist/rustc-nightly-src.tar.xz` is the same file of `https://static.rust-lang.org/dist/$(date +%Y-%m-%d)/rustc-nightly-src.tar.xz`. If you want TUR to keep up to date with the `rustc-nightly`, you need to customize `termux_pkg_auto_update`.

Ludea:
Maybe [this](https://discord.com/channels/273534239310479360/1080209248526532708)  can explain

Ludea:
> Emmm... I don't think this is a good idea. In AUR, [`rust-nightly`](https://aur.archlinux.org/cgit/aur.git/tree/PKGBUILD?h=rust-nightly-bin) is built on your local machine and you can always keep it up to date. But TUR is not like AUR, this package is pre-built using GitHub Action, so I don't think the date should be removed.

With date, rustc is newer than rust src tarball (not pulled from GH, but download from rust website with latest tag).
If someone add/remove a feature into rustc, you release rust nightly package with this feature, building rust from source can fail.

I see 2 solutions
- Build rustc nightly from tag (what this PR bring) 
- provide rust src (maybe from another package) with same date as rustc

licy183:
The source of stdlib has been provided as `rustc-nightly-src`. As I've said above, `https://static.rust-lang.org/dist/rustc-nightly-src.tar.xz` (which URL your PR is using) is the same file of `https://static.rust-lang.org/dist/$(date +%Y-%m-%d)/rustc-nightly-src.tar.xz` (the source tarball generated in the morning). If this package is not up to date, you could submit a PR or custom `termux_pkg_auto_update`.

Ludea:
> The source of stdlib has been provided as `rustc-nightly-src`. As I've said above, `https://static.rust-lang.org/dist/rustc-nightly-src.tar.xz` (which URL your PR is using) is not from tag too, it is the same file of `https://static.rust-lang.org/dist/$(date +%Y-%m-%d)/rustc-nightly-src.tar.xz` (the source tarball generated in the morning). If this package is not up to date, you could submit a PR or custom `termux_pkg_auto_update`.

So the only way is to provide src with toolchain

Ludea:
I include src as subpackage

licy183:
So, why not download the source tarball from `https://static.rust-lang.org/dist/$_DATE/rustc-nightly-src.tar.xz` rather than packaging it? The date of the original source can be queried through apt (in Termux).

Ludea:
It could be a solution, but but we need to do this on each package update

licy183:
> It could be a solution, but but we need to do this on each package update

I still don't think it is a good choice to distribute source tarball with binary packages release... Getting the source tarball on Termux is easy, and there is no other binary distribution packing the source tarball too...

Ludea:
> I still don't think it is a good choice to distribute source tarball with binary packages release... Getting the source tarball on Termux is easy, and there is no other binary distribution packing the source tarball too...

Ok will do by hand

PR #320, opened 2023-03-14
add(tur): blender deps

T-Dynamos:
As Blender does not support cross-compiling, I decided to use Tur-on-device. In order to build Blender, I need these dependencies, as they would not be accepted in termux-packages.


T-Dynamos:
Do I need to split commits?

licy183:
> Do I need to split commits?

I think it is better.

T-Dynamos:
@licy183 It's ready to be merged!

licy183:
Thanks!

PR #335, opened 2023-03-29
wine-stable: make wine-tools only

kawanakaiku:


T-Dynamos:
Please bump revision

kawanakaiku:
I don't think it's necessary because the package itself hasn't changed.

licy183:
Thanks!

PR #361, opened 2023-04-16
libicu: Revbump revdeps to rebuild

2096779623:
https://github.com/termux/termux-packages/issues/16222


licy183:
There is no need to revbump `chromium` as it links libicu statically.

PR #364, opened 2023-04-19
[tur] new package hare

Sertonix:
This adds the Hare language packages.
This includes `harec` the bootstrap compiler and `hare` an overall toolset.

~~The build of `hare` on my android device varied a lot depending on the used compiler/linker/assembler.~~
~~I hope they run in the CI.~~

~~I will merge all commits to 2 commits later!~~

Edit: Done

### DIFF
--- a/tur/hare/Makefile.patch
+++ b/tur/hare/Makefile.patch
@@ -1,0 +1,36 @@
+diff --git a/Makefile b/Makefile
+index 33b7887b..3414735c 100644
+--- a/Makefile
++++ b/Makefile
+@@ -10,7 +10,7 @@ all:
+ .SUFFIXES: .ha .ssa .s .o .scd .1
+ .ssa.s:
+ 	@printf 'QBE\t%s\n' "$@"
+-	@$(QBE) -o $@ $<
++	@$(QBE) $(QBEFLAGS) -o $@ $<
+ 
+ .s.o:
+ 	@printf 'AS\t%s\n' "$@"
+@@ -72,8 +72,8 @@ $(BINOUT)/hare-tests: $(TESTCACHE)/hare.o
+ $(BINOUT)/harec2: $(BINOUT)/hare $(harec_srcs)
+ 	@mkdir -p $(BINOUT)
+ 	@printf 'HARE\t%s\n' "$@"
+-	@env HAREPATH=. HAREC=$(HAREC) QBE=$(QBE) $(BINOUT)/hare build \
+-		$(HARE_DEFINES) -o $(BINOUT)/harec2 cmd/harec
++	@env HAREPATH=. HAREC=$(HAREC) QBE=$(QBE) $(HOST_HARE) build \
++		$(HARE_DEFINES) $(HAREBUILDFLAGS) -o $(BINOUT)/harec2 cmd/harec
+ 
+ # Prevent $(BINOUT)/hare from running builds in parallel, workaround for build
+ # driver bugs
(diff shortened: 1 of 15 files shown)


Sertonix:
Why does it say that it has already build qbe but qbe is not found?

SunPodder:
Termux doesn’t have the qbe package yet. You may add it to the repo first then add hare.

SunPodder:
Sorry my package list was outdated

Sertonix:
It seems that the qbe binary is in `$TERMUX_PREFIX/bin` but that directory is not in the `PATH`.
Should be binary be somewhere?

licy183:
The packages in `./tur` folders should be able to perform cross-compiling. I'm not familiar with hare compilers so I don't know how to cross-compile it. If this package is finally confirmed that it is unable to cross-compile, you can submit this package in `tur-on-device` and see if it is able to compile using `termux-docker`.

Sertonix:
I can specify a target architecture on all components so cross-compiling should work. I only have problems to get it to work.
I am pretty sure that I need the hostbuild functions but unfortunatly they are not documented.
I currently test on my own repo [Sertonix/tur-tests](https://github.com/Sertonix/tur-tests).
I have still some ideas why it is failing and hope to get that fixed soon.

Sertonix:
Can somebody explain me where I am supposed to get the hostbuild binaries?
For example with [scdoc](https://github.com/termux/termux-packages/blob/master/packages/scdoc/build.sh) it would be `$TERMUX_TOPDIR/scdoc/host-build/scdoc` but I haven't seen that used anywhere.

Sertonix:
I have finally got the build working!
I currently have used a small hack by accessing the host build binaries from ther packages.
eg. `$TERMUX_TOPDIR/qbe/host-build/qbe`

I could make scripts like `termux_setup_qbe` but there is no good place to access the from multiple files and they would need to be updated in addition to the main packages.
It would be nice if a maintainer could comment on that because they should be able to maintain it.

Sertonix:
I just have seen [this](https://github.com/termux-user-repository/tur/commit/439f323b48733c38a50d316b27899e985aba4a94) commit that uses a subpackage package for the cross build.
I think that is a good solution and will try to do that as soon as I have time.

SunPodder:
Or how about creating a `termux_setup_qbe` script and put it in the qba directory but in a separate file? So, whenever a package needs to setup qbe it can source that file and run the function.

Sertonix:
Now it is finally working!
The cross packages now include a setup script. I used a similar path to the cross binaries.
eg: `$TERMUX_PREFIX/opt/qbe/cross/setup.sh`

Also on device build can work. It works on my device at least but I used some hacky tricks to get it to work.

I have the feeling that the default assembler is not correctly set up (maybe not commonly used standalone by other packages). I had to set the assembler to `$TERMUX_PREFIX/opt/binutils/cross/$TERMUX_HOST_PLATFORM/bin/as` multiple times to get it to work. Could this maybe be the default when building crossplatform?

licy183:
> I have the feeling that the default assembler is not correctly set up (maybe not commonly used standalone by other packages). I had to set the assembler to `$TERMUX_PREFIX/opt/binutils/cross/$TERMUX_HOST_PLATFORM/bin/as` multiple times to get it to work. Could this maybe be the default when building crossplatform?

I'm not sure, but AS environment variable is set by default for C toolchain in Termux package builder, so I think it may be the right method to cross compile hare.

licy183 (approved):
Thanks for your contribution! I'll merge this this afternoon if there are no more reviews.

Issue #375, opened 2023-04-26
[Package]: python-selenium/python-seledroid


nacho00112:
### Name

Selenium/SeleDroid

### Home page URL

https://github.com/luanon404/seledroid

### Source code URL

https://github.com/luanon404/seledroid

### Packaging policy acknowledgement

- [X] I understand [Tur Package Policy](https://github.com/termux-user-repository/tur/wiki/TUR-package-policy.).

### Additional information

Yeah, it is possible to have [selenium](https://github.com/SeleniumHQ/selenium/wiki/Python-Bindings) working on termux, thanks to this project: [seledroid](https://github.com/luanon404/seledroid) of [luanon404](https://github.com/luanon404), it is third-app dependent of an apk provided of 2MB approximately in [android-webdrivers](https://github.com/luanon404/android-webdrivers) of [luanon404](https://github.com/luanon404) too, but we can solve this with `termux-open` to install it automatically, some names of things that seledroid have are slightly different from the selenium things, for example, in selenium there are a `selenium.webdriver.common.keys.Keys.RETURN` constant but in seledroid it is `selenium.webdriver.common.keys.Keys.ENTER`, so to make it same as selenium we just need to rename those things in the source code

nacho00112:
I created a pull request with the package here [pull#377](https://github.com/termux-user-repository/tur/pull/377)

SunPodder:
Well, selenium itself works in termux. What is different in it from selenium (probably any drawbacks of selenium that it solves)?

> seledroid-app uses termux-open to install the application that allows using the phone's chromium via python-seledroid
Do you mean it will use my phone's pre-installed chrome.apk? It'll surely a great feature then.

nacho00112:
> Well, selenium itself works in termux. What is different in it from selenium (probably any drawbacks of selenium that it solves)?
> 
> > seledroid-app uses termux-open to install the application that allows using the phone's chromium via python-seledroid
> 
> Do you mean it will use my phone's pre-installed chrome.apk? It'll surely a great feature then.

Selenium did not working because there are no way to install a webdriver, but now i noticed tur has chromium, i will try it

PR #377, opened 2023-04-28
python-selenium

nacho00112:
Finally a termux package for Selenium, i taked the idea from the issue i opened at [issue#375](https://github.com/termux-user-repository/tur/issues/375), this pull-request add the following packages: python-selenium and, python-seledroid and seledroid-app because they are dependencies, seledroid-app uses termux-open to install the application that allows using the phone's chromium via python-seledroid, seledroid-app temporarily enables allow-external-apps in ~/.termux/termux.properties and then install the application, once it is installed, the allow-external-apps comes back to his original configuration, python-selenium imports the things from python-seledroid, this is the first pull-request of repository that i have made, and the first termux package i have created too, so if i did something wrong please let me know so i can solve it.

### DIFF
--- a/tur/libllvm-classic-flang/build.sh
+++ b/tur/libllvm-classic-flang/build.sh
@@ -1,21 +1,20 @@
 TERMUX_PKG_HOMEPAGE=https://github.com/flang-compiler/classic-flang-llvm-project/
-TERMUX_PKG_DESCRIPTION="Modular compiler and toolchain technologies library (Version 10, Classic Flang Fork)"
+TERMUX_PKG_DESCRIPTION="Modular compiler and toolchain technologies library (Version 15, Classic Flang Fork)"
 TERMUX_PKG_LICENSE="Apache-2.0"
 TERMUX_PKG_LICENSE_FILE="llvm/LICENSE.TXT"
 TERMUX_PKG_MAINTAINER="@termux-user-repository"
-_DATE=2022.05.23
+_DATE=2023.04.19
 TERMUX_PKG_VERSION=${_DATE//./}
-TERMUX_PKG_REVISION=4
-_LLVM_COMMIT=4a566ddfa4157ed1a24920ab7673ccc01c46fd99
+_LLVM_COMMIT=cd736e11b188a8f6ff14041abd818ad86f36b9bb
 TERMUX_PKG_SRCURL=https://github.com/flang-compiler/classic-flang-llvm-project/archive/${_LLVM_COMMIT}.zip
-TERMUX_PKG_SHA256=d390229b7ffa7b1db7b8bf2fe9352b597da801913df8d2dc2e65d27377433d2d
+TERMUX_PKG_SHA256=6a5caa2ccfabf9492443c31762900fc7c945201d43b3a705f31d56256091b109
 TERMUX_PKG_HOSTBUILD=true
 TERMUX_PKG_DEPENDS="binutils-is-llvm, libc++, ncurses, ndk-sysroot, libffi, zlib"
-TERMUX_PKG_SUGGESTS="libandroid-mathlib, classic-flang"
+TERMUX_PKG_SUGGESTS="libandroid-complex-math, classic-flang"
 # XXX: We may add this package later I suppose.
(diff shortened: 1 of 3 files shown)


SunPodder:
If you're naming one of your packages `python-selenium` then it must be installing only the core selenium package independently from seledroid. Please look into that.

nacho00112:
Is a dummy selenium, so we can import via `selenium` instead of `seledroid`

SunPodder:
Then this name might be perfect `python-selenium-is-seledroid`. `python-selednium` will create misconception among the users as most of them will assume its the selenium package, not seledroid

licy183:
It seems that seledroid is unmaintained, and I think the original selenium will work on Termux because TUR has webdriver in chromium package. If you really need this package, please just name it `python-seledroid`.

nacho00112:
> It seems that seledroid is unmaintained, and I think the original selenium will work on Termux because TUR has webdriver in chromium package. If you really need this package, please just name it `python-seledroid`.

We can see it as a final version

> It seems that seledroid is unmaintained, and I think the original selenium will work on Termux because TUR has webdriver in chromium package. If you really need this package, please just name it `python-seledroid`.

Since luanon used all his knowledge we can see it as a final version, anyway for now i think is no need to edit the code, until we encounter some bug, but it is quite stable, i didn't find any bug

nacho00112:
Seems the i686 CI check depends on a non-existing package "sse2neon", the other checks builds correctly

tur/seledroid-app/build.sh
```diff
@@ -0,0 +1,52 @@
...
+termux_step_make_install() {
+    SHARED=$TERMUX_PREFIX/share/seledroid-app
+    mkdir -p $SHARED
+    cp Apk-Android-WebDrivers/Chromium/Chromium-$TERMUX_PKG_VERSION.apk $SHARED
+    echo $TERMUX_PKG_VERSION > VERSION
+    cp VERSION $SHARED
+    # Highly compressed install-seledroid-app script, ";" are replaced by "\n" later.
+    # compressed because so saves space in this build.sh script.
+    # because the install-seledroid-app script is short, it not have to be hosted, it can be here locally.
+    echo '#!/data/data/com.termux/files/usr/bin/bash
```
licy183: I think these code should be store in a seperate file and copy this file to the target folder in `build.sh`. It is not readable for maintainers.

tur/python-selenium-is-seledroid/build.sh
```diff
@@ -0,0 +1,23 @@
...
+TERMUX_PKG_DESCRIPTION="A browser automation framework and ecosystem. This dummy selenium uses seledroid as his backend."
+TERMUX_PKG_SKIP_SRC_EXTRACT=true
+TERMUX_PKG_LICENSE="Unlicense" # seledroid is unlicensed, the original selenium is Apache-2.0
+TERMUX_PKG_MAINTAINER="@termux-user-repository"
+TERMUX_PKG_VERSION=4.8.2
+TERMUX_PKG_AUTO_UPDATE=true
+TERMUX_PKG_DEPENDS="python, python-seledroid"
+
+termux_step_make_install() {
+    echo 'from setuptools import setup;setup(name="selenium",version="'"$TERMUX_PKG_VERSION"'",description="'"$TERMUX_PKG_DESCRIPTION"'",
```
licy183: The same here.

licy183 (changes requested):

tur/seledroid-app/build.sh
nacho00112: The thing is that i don't know how to do that, i believed that it is not possible, but now i learned how, i already pushed the commit with the changes

tur/python-selenium-is-seledroid/build.sh
nacho00112: Here i changed too

nacho00112:
Everything should work stable now

T-Dynamos:
@nacho00112 You should consider squashing commits.

nacho00112:
I tried and googled a lot but i can't merge the commits into one

PR #383, opened 2023-04-29
bump(tur/openimageio): 2.4.10.0

T-Dynamos:


T-Dynamos:
@licy183 I have a doubt, If I am building openimageio, why it is saying:
```
ERROR: Package blender depends on non-existing package "sse2neon"
```

licy183:
I have no idea about this error...

kawanakaiku:
I think this commit https://github.com/termux/termux-packages/commit/40c7b533db14beff2beaa28ac9724dfe3d4165dd causes the error.

T-Dynamos:
I saw issue: https://github.com/termux/termux-packages/issues/16427
It says:

>Lets say package A depends on package B and package B can't be build on aarch64.
When building package A with target aarch64 it runs the build even thought package B should never be build with target aarch64.

But according to this situation, I am building package B and why it is checking package A.

Issue #421, opened 2023-05-24
[Bug]: python-selenium-is-seledroid is installed in wrong directory


nacho00112:
### Problem description

`python-seledroid-is-selenium` should be installed at `$PREFIX/lib/python3.11/site-packages`, but it's installed in `$PREFIX/local/lib/python3.10/dist-packages`

### What steps will reproduce the bug?

```bash
pkg install tur-repo
pkg install python-selenium-is-seledroid
```

### What is the expected behavior?

`python-selenium-is-seledroid` to be installed at `$PREFIX/lib/python3.11/site-packages`

### System information

termux-info:

```
.
```
The problem seems to be with the installation rather than my device.

PR #432, opened 2023-05-27
Fixes issue #421 but for python-seledroid

nacho00112:
Add the `TERMUX_PKG_SETUP_PYTHON=true` declaration that i missed in the `python-selenium-is-seledroid` but for `python-seledroid`


licy183:
Thanks!

PR #453, opened 2023-06-11
perl-net-ssleay: bump after perl update

Grimler91:
So that perl-net-ssleay installs to `$PREFIX/lib/perl5/5.36.1` where the other perl stuff currently resides.


licy183:
Thanks!

PR #454, opened 2023-06-13
new package: biber-installer

Grimler91:
Replaces biber in the main repo, will remove that package after this is merged. Have kept git history and hence the PR has commits that should normally be squashed.

Fixes termux/termux-packages#16530.


licy183:
Thanks!

PR #467, opened 2023-06-25
new package: vkmark

licy183:


Biswa96:
vkmark was added in main termux-packages repository in https://github.com/termux/termux-packages/commit/e7c7b3c5ee7624d4076bd16e49e376bfb4b5dc7c commit. I think vkmark can be removed from TUR.

PR #481, opened 2023-07-03
python-selenium-is-seledroid: fix bug of python selenium.webdriver.su…

nacho00112:
For some reason `import selenium.webdriver.support.expected_conditions as EC` does not worked but
`import selenium.webdriver.supported.expected_conditions` worked,
So i changed the way the dummy works and the error is solved


tur/python-selenium-is-seledroid/build.sh
```diff
@@ -4,7 +4,7 @@ TERMUX_PKG_SKIP_SRC_EXTRACT=true
 # seledroid is unlicensed, the original selenium is Apache-2.0
 TERMUX_PKG_LICENSE="Unlicense"
 TERMUX_PKG_MAINTAINER="@nacho00112"
-TERMUX_PKG_VERSION=4.8.2
+TERMUX_PKG_VERSION=4.10.0
 TERMUX_PKG_REVISION=1
```
2096779623: Please remove this line.
```diff
- TERMUX_PKG_REVISION=1
```

2096779623 (changes requested):

tur/python-selenium-is-seledroid/build.sh
nacho00112: Done

licy183:
Thanks!

PR #526, opened 2023-08-08
New Package: clspkg

turbomaster95:
I Have added a package called 'clspkg'
it adds the command 'cls' to clear the screen

i am using gh cli to type this


turbomaster95:
pls merge

T-Dynamos:
I have one question: why?
Why can't you do `alias cls=clear` or `echo "clear" > $PREFIX/bin/cls && chmod +x $PREFIX/bin/cls`
What special does `clspkg` do?

licy183:
I don't think this package will be added, see our package policy.

```
TUR may not accepts these packages:
Few lines of scripting code.
```

turbomaster95:
i just had this same conversation with two prople on the official termux-packages repo

PR #772, opened 2024-01-10
new package: boxedwine

amuramatsu:
Boxedwine (https://github.com/danoon2/Boxedwine) is an linux-x86 emulator specialized for running wine.
This software can run 16/32 bit windows binaries on termux with arm and intel architecture.


licy183:
Thanks!

PR #805, opened 2024-02-05
new package: stern

idj0:
feel free to let me know or change anything if there are any mistakes

### DIFF
--- a/tur/stern/build.sh
+++ b/tur/stern/build.sh
@@ -13,28 +13,32 @@ TERMUX_PKG_AUTO_UPDATE=true
 TERMUX_PKG_UPDATE_TAG_TYPE="newest-tag"
 
 termux_step_make() {
-	termux_setup_golang
+    termux_setup_golang
 
-	mkdir bin
-	go build -o ./bin ./...
-
-	mkdir -p completions/zsh
-	./bin/stern --completion zsh >completions/zsh/_stern
-
-	mkdir -p completions/bash
-	./bin/stern --completion bash >completions/bash/stern
-
-	mkdir -p completions/fish
-	./bin/stern --completion fish >completions/fish/stern.fish
+    mkdir bin
+    go build -o ./bin ./...
 }
 
 termux_step_make_install() {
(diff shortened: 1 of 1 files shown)


tur/stern/build.sh
```diff
@@ -0,0 +1,40 @@
...
+TERMUX_PKG_UPDATE_TAG_TYPE="newest-tag"
+
+termux_step_make() {
+	termux_setup_golang
+
+	mkdir bin
+	go build -o ./bin ./...
+
+	mkdir -p completions/zsh
+	./bin/stern --completion zsh >completions/zsh/_stern
```
licy183: The target of `./bin/stern` is `$TERMUX_ARCH-linux`. It may be able to execute because of qemu translation layer, but it is not a proper way.

Please ref https://github.com/termux/termux-packages/blob/master/packages/starship/build.sh#L49 to see how to properly generate and install shell completions.

idjoo: ok got it i'll use the postinst instead, thankyou for pointing out

idj0:
Hi @licy183, Just made the changes you asked for. Could you take another look?

tur/stern/build.sh
```diff
@@ -0,0 +1,44 @@
+TERMUX_PKG_HOMEPAGE=https://github.com/stern/stern
+TERMUX_PKG_DESCRIPTION="⎈ Multi pod and container log tailing for Kubernetes -- Friendly fork of https://github.com/wercker/stern"
```
licy183: Ahh.. It seems that this description is too long

idjoo: ahhh ic 🤣, ok i'll fix it, maybe it should be just "Multi pod and container log tailing for Kubernetes"?

idj0:
Done truncating the description to 93 char

idj0:
Sorry for the inconvenience it seems the license also invalid i already look at the other package and change it accordingly to "Apache-2.0"

licy183:
Thanks!

PR #832, opened 2024-02-17
Dev Container and revised README.md 

knyipab:
The devcontainer should faciliate new developers in making new packages with VS Code (local) or Codespaces (online), especially when they hesiate to install dependencies in their native environment. It is much like the official docker image, but it comes with better integration with VS Code. 

The README.md "Contribution" section is also revised to provide clearer instructions for developers who want to contribute. 

Hope this can be accepted.


licy183:
Thanks!

PR #834, opened 2024-02-18
new package: darktable

knyipab:
Thanks @licy183 for your help. `darktable` GUI app is tested on my aarch64 device. 

My first time making a pacakge. Indeed, some workarounds are needed but I don't know why it happens, such as:
* `ln -s $TERMUX_PREFIX/lib/darktable/libdarktable.so $TERMUX_PREFIX/lib/libdarktable.so` (binary file `darktable` cannot find libdarktable.so if this link is not done)
* `FIND_LIBRARY(MATH_LIBRARY m)` does not work in CMakeLists.txt

Few more things remain puzzling for me: 
* The `proot-distro` ubuntu version of darktable 3.x seems to render and its UI responds smoother and faster than this termux native build
* On my adreno device, OpenCL is not stable (some DNG files can crash the program) and can only activate with `LD_LIBRARY_PATH=/vendor/lib64 darktable` (indeed installing `ocl-icd` and `opencl-vendor-driver` does not work) and preference -> processing are properly configured. However, this OpenCL does not help much and it is still slower than the proot version. 

Happy to discuss.


knyipab:
I am not familiar with the matter: should I make this package auto-updating? And how can I do so? Thanks.

licy183:
Thanks for your contribution!

> `ln -s $TERMUX_PREFIX/lib/darktable/libdarktable.so $TERMUX_PREFIX/lib/libdarktable.so` (binary file `darktable` cannot find libdarktable.so if this link is not done)

By default, `cmake` will set RPATH rather than RUNPATH of a binary file, but RPATH doesn't support by Android, so the building system of Termux packages automatically set `CMAKE_SKIP_INSTALL_RPATH` to `ON` to get rid of this, and manually set RUNPATH to `$PREFIX/lib` in `LDFLAGS`. `darktable` uses customize rpath in cmake, but it will not be properly set due to `CMAKE_SKIP_INSTALL_RPATH=ON`.

> `FIND_LIBRARY(MATH_LIBRARY m)` does not work in CMakeLists.txt

`FIND_LIBRARY` in `cmake` will find libraries in `CMAKE_INSTALL_PREFIX` (`$PREIFX/lib` in Termux), but `libm.so` doesn't exist there. It exists in `$SYSROOT/usr/lib/24/libm.so`.

knyipab:
I see. Really appreciate your answer! And I learn from the changes you made.

PR #837, opened 2024-02-20
new package: libiptcdata

knyipab:
One of dependencies of rawtherapee


knyipab:
sorry, I force-pushed to correct the license of libiptcdata

PR #849, opened 2024-02-26
Call entrypoint on docker-exec

Kapral67:
## BREAKING CHANGE

- Breaks [common-files/build-termux-docker.sh](https://github.com/termux-user-repository/tur/blob/57b28c4921381567294b456370e2b4818e5d2512/common-files/build-termux-docker.sh) unless termux-user-repository/termux-docker#1 is merged

PR #863, opened 2024-02-29
new package(tur-on-device): awscli

Kapral67:
- workflow will probably fail unless #849 gets merged

### DIFF
--- a/tur-on-device/aws-cli-v2/build.sh
+++ b/tur-on-device/aws-cli-v2/build.sh
@@ -1,0 +1,152 @@
+# shellcheck shell=bash
+# shellcheck disable=SC2034
+TERMUX_PKG_HOMEPAGE="https://aws.amazon.com/cli"
+TERMUX_PKG_DESCRIPTION="A Unified Tool to Manage Your AWS Services"
+TERMUX_PKG_LICENSE="Apache-2.0"
+TERMUX_PKG_MAINTAINER="@termux-user-repository"
+TERMUX_PKG_VERSION="2.15.23"
+TERMUX_PKG_SRCURL="https://awscli.amazonaws.com/awscli-${TERMUX_PKG_VERSION}.tar.gz"
+TERMUX_PKG_SHA256="SKIP_CHECKSUM"
+TERMUX_PKG_SKIP_SRC_EXTRACT=true
+TERMUX_PKG_AUTO_UPDATE=true
+TERMUX_PKG_BUILD_IN_SRC=true
+TERMUX_WITHOUT_DEPVERSION_BINDING=true
+TERMUX_PKG_SETUP_PYTHON=true
+TERMUX_PKG_DEPENDS="man"
+TERMUX_PKG_BUILD_DEPENDS="ldd, python-pip"
+
+_import_awscli_pgp_key() {
+	# This key expired 2023-09-17 but it is still in use
+	# https://docs.aws.amazon.com/cli/latest/userguide/getting-started-source-install.html#source-getting-started-install-reqs
+	cat <<-EOF | gpg --import -
+		-----BEGIN PGP PUBLIC KEY BLOCK-----
+
+		mQINBF2Cr7UBEADJZHcgusOJl7ENSyumXh85z0TRV0xJorM2B/JL0kHOyigQluUG
+		ZMLhENaG0bYatdrKP+3H91lvK050pXwnO/R7fB/FSTouki4ciIx5OuLlnJZIxSzx
+		PqGl0mkxImLNbGWoi6Lto0LYxqHN2iQtzlwTVmq9733zd3XfcXrZ3+LblHAgEt5G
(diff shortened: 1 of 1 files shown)


Kapral67:
workflow is still pulling old termux-user-repository/termux-docker image, this requires latest image that has proper dns resolution

tur-on-device/awscli/build.sh
```diff
@@ -0,0 +1,152 @@
...
+		termux_error_exit "Error: failed to extract latest version"
+	fi
+
+	if [ "${TERMUX_PKG_VERSION}" != "${version}" ]; then
+		termux_pkg_upgrade_version "${version}" --skip-version-check
+	fi
+}
+
+termux_step_get_source() {
+	local tarball
```
licy183: Why not use GitHub to fetch source? Using GitHub will make it easy to enable auto-update.

tur-on-device/awscli/build.sh
```diff
@@ -0,0 +1,152 @@
+# shellcheck shell=bash
+# shellcheck disable=SC2034
+TERMUX_PKG_HOMEPAGE="https://aws.amazon.com/cli"
+TERMUX_PKG_DESCRIPTION="A Unified Tool to Manage Your AWS Services"
+TERMUX_PKG_LICENSE="Apache-2.0"
+TERMUX_PKG_MAINTAINER="@termux-user-repository"
+TERMUX_PKG_VERSION="2.15.23"
```
licy183: I think this package should be renamed to `aws-cli-v2` because there is a legacy version for `aws-cli` and it is still under development.

tur-on-device/awscli/build.sh
```diff
@@ -0,0 +1,152 @@
...
+
+	termux_setup_rust
+	pip3 install "cryptography${cryptography_version}"
+
+	termux_setup_cmake
+	pip3 install "awscrt${awscrt_version}"
+}
+
+termux_step_configure() {
+	./configure --prefix="${TERMUX_PREFIX}" --with-install-type=portable-exe
```
licy183: It would be more suitable to install aws-cli-v2 as a python package, rather than a standalone binary. Some extra licenses should be added to this if we release a standalone binary as it will package so many third-party software.

tur-on-device/awscli/build.sh
```diff
@@ -0,0 +1,152 @@
...
+termux_step_get_source() {
+	local tarball
+	tarball="$(_get_awscli_src_tarball)"
+	mkdir -p "${TERMUX_PKG_SRCDIR}"
+	tar --strip-components=1 -xf "${tarball}" -C "${TERMUX_PKG_SRCDIR}"
+	rm -f "${tarball}"
+}
+
+termux_step_pre_configure() {
+	if ! ${TERMUX_ON_DEVICE_BUILD}; then
```
licy183: If this package is installed as a python package, is it support cross-compiling?

tur-on-device/awscli/build.sh
Kapral67: I was thinking about that but I wasn't sure how the auto updates would work given that they simultaneously track 1.x and 2.x updates in the same repo with tags. If we can somehow force the auto updater to only look at 2.x tags that might be better.

tur-on-device/awscli/build.sh
Kapral67: awscli 1.x is fully installable via pypi so it is unlikely a package named aws-cli would be accepted either here or in termux-packages that is not 2.x version. Also, 1.x and 2.x are not recommended to be installed simulatenously as both executables are the same name `aws`.

Still, I can see changing the name to aws-cli-v2 for clarity

tur-on-device/awscli/build.sh
Kapral67: aws-cli-v2 is not available on pypi aws/aws-cli#4947

If we do not package as a standalone binary, then this package needs to have a dependency on python. In that case an update to python3 in termux-packages could break this package when python is updated. AWS does not always like to support the latest software, so building a portable binary would ensure other package updates do not break this package. aws/aws-cli#8342 for example

From a licensing perspective, would we need to include licenses from all of aws-cli-v2 dependencies along with this package? I would think as long as we package with the license aws distributes with their source code we should be good (I am not a lawyer though). If we package as a portable binary, we are not distributing any of the dependencies, but only the `aws` binary so I only see the need for the Apache-2.0 License here

tur-on-device/awscli/build.sh
Kapral67: when compiled, aws uses [pyinstaller](https://pyinstaller.org) which does not support cross compilation.

Kapral67:
actions still not pulling latest built image from termux-user-repoistory/termux-docker so dns resolution is failing

tur-on-device/awscli/build.sh
licy183: Actually this is not a problem. If the tag does not meet the increment conditions, the auto-update script will not update checksum for it.

Kapral67:
I'm don't think this is going to be able to compile in github actions

PR #908, opened 2024-04-03
addpkg(main/yc): Yao scripting language and Rig build system

YzenaLLC:
More information about Yao and Rig is at https://rigbuild.dev/ .

Yao and Rig are part of a monorepo, so this package includes both.

Since Rig is a build system, it is its own build system, so this package requires a host build for bootstrap.

Both are currently in alpha, but a Termux user requested that I make this package.

That said, I understand if this package is not wanted, especially since this package is under the SSPL and another non-Open Source license, which may cause problems with parsing the license identifier.

If accepted, I intend to maintain this package.

### DIFF
--- a/tur/yc/build.sh
+++ b/tur/yc/build.sh
@@ -1,0 +1,44 @@
+TERMUX_PKG_HOMEPAGE=https://git.yzena.com/Yzena/Yc
+TERMUX_PKG_DESCRIPTION="Monopackage with Rig, Yao, and Yvm"
+TERMUX_PKG_LICENSE="custom" # Yc is under two licenses; both must be followed.
+TERMUX_PKG_LICENSE_FILE="LICENSE.YNL.md, LICENSE.SSPL.txt"
+TERMUX_PKG_MAINTAINER="Gavin D. Howard <gavin@yzena.com>"
+TERMUX_PKG_VERSION=24.04.04
+TERMUX_PKG_SRCURL=https://git.yzena.com/Yzena/Yc/releases/download/${TERMUX_PKG_VERSION}/${TERMUX_PKG_NAME}-${TERMUX_PKG_VERSION}.tar.gz
+TERMUX_PKG_SHA256=5e0ca7e087048da37192c51ea9e6873b69c68dbd2668ac7682250e8498315fed
+TERMUX_PKG_HOSTBUILD=true
+
+# Yc has its own build system, so it needs to bootstrap on the host first.
+termux_step_host_build() {
+
+	cd "$TERMUX_PKG_BUILDDIR"
+
+	# The source needs to be in the build directory for bootstrap.
+	cp -r "$TERMUX_PKG_SRCDIR" ./bootstrap
+	cd ./bootstrap
+
+	# Yc has a C program that does the bootstrap.
+	gcc -o bootstrap/bootstrap bootstrap/bootstrap.c
+	./bootstrap/bootstrap gcc
+}
+
+termux_step_configure() {
+	cd "$TERMUX_PKG_BUILDDIR"
+	# It's just easier to have the source in the build directory for the build.
+	cp -r "$TERMUX_PKG_SRCDIR" ./yc
+}
+
+termux_step_make() {
+	cd "$TERMUX_PKG_BUILDDIR/yc"
+	# Yc's build likes to stomp on $PATH, so we need to stop that. It also needs
+	# to know that it is being built for Android, as well as what compiler to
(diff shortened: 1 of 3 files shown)


tur/yc/build.sh
```diff
@@ -0,0 +1,59 @@
+TERMUX_PKG_HOMEPAGE=https://git.yzena.com/Yzena/Yc
+TERMUX_PKG_DESCRIPTION="Monopackage with Rig, Yao, and Yvm"
+TERMUX_PKG_LICENSE="LicenseRef-LICENSE.YNL.md AND SSPL-1.0"
```
TomJo2000: This should probably be set to `custom` and the license files be listed using [`TERMUX_PKG_LICENSE_FILE`](https://github.com/termux/termux-packages/wiki/Creating-new-package#table-of-available-package-control-fields), which is a comma or newline separated string of filepaths (relative to the project root)
See: https://github.com/termux/termux-packages/blob/08b1a141cbf804c3450a2649d7a7fc6449e519a9/packages/swig/build.sh#L3-L4
for and example.

In this case that might look something like:
```suggestion
TERMUX_PKG_LICENSE="custom" # Yc is under two licenses, and both must be followed.
TERMUX_PKG_LICENSE_FILE="LICENSE.YNL.md, LICENSE.SSPL.txt"
```
Which should additionally obviate the need to overwrite `termux_step_install_license()` below

tur/yc/build.sh
```diff
@@ -0,0 +1,59 @@
...
+	# The executable is a multicall binary, so we link all of the possible
+	# programs to it.
+	ln -sf ./yc $TERMUX_PREFIX/bin/rig
+	ln -sf ./yc $TERMUX_PREFIX/bin/rigr
+	ln -sf ./yc $TERMUX_PREFIX/bin/rigc
+	ln -sf ./yc $TERMUX_PREFIX/bin/yao
+	chmod 700 $TERMUX_PREFIX/bin/rig{r,c,} $TERMUX_PREFIX/bin/yao
+}
+
+termux_step_install_license() {
```
TomJo2000: Overriding this function shouldn't be necessary when using `TERMUX_PKG_LICENSE_FILE` to manually specify the licenses

tur/yc/build.sh
```diff
@@ -0,0 +1,59 @@
+TERMUX_PKG_HOMEPAGE=https://git.yzena.com/Yzena/Yc
+TERMUX_PKG_DESCRIPTION="Monopackage with Rig, Yao, and Yvm"
```
TomJo2000, line 2: It might be worth splitting out the individual programs of the `yc` multi-call binary,
into [subpackages](https://github.com/termux/termux-packages/wiki/Creating-new-package#writing-a-subpackage-script) which then create their respective links in `$PREFIX/bin`, or create a simple wrapper shell script to call the main `yc` binary in the correct manner.

This may be more flexible than a simple `ln -sf` in `termux_make_install()`.
For a *similar-ish* example see: https://github.com/termux/termux-packages/blob/37e77ab30f722ade25b9a24c8ffed077b083c356/packages/libllvm/libcompiler-rt.subpackage.sh#L18-L40

tur/yc/build.sh
```diff
@@ -0,0 +1,59 @@
...
+	cp -r "$TERMUX_PKG_SRCDIR" ./yc
+}
+
+termux_step_make() {
+	cd "$TERMUX_PKG_BUILDDIR/yc"
+	# Yc's build likes to stomp on $PATH, so we need to stop that. It also needs
+	# to know that it is being built for Android, as well as what compiler to
+	# use. Finally, the build automatically defines _BSD_SOURCE, which doesn't
+	# work on Android.
+	../bootstrap/release/yc rigr --preserve-path --os=Android --compiler=gcc \
```
TomJo2000: Termux uses `clang` as its C compiler by default.
Be aware that `gcc` may be a symlink to Clang.

TomJo2000 (commented):
Since I didn't get around to reviewing the build script in the original PR (termux/termux-packages#19657), I am doing so now.

I have no maintainer privileges in this repo, and these are mainly "best practice" recommendations.

tur/yc/build.sh
YzenaLLC: Making a subpackage was confusing, and I don't know if I did it correctly, but it doesn't give any warnings.

YzenaLLC:
Withdrawing the PR because I can't get the download to work. I apologize for the noise.

PR #909, opened 2024-04-03
pynvim: init at 0.5.0

Freed-Wu:
Continue #856. Or should it be pushed to termux-packages due to https://github.com/termux/termux-packages/issues/19368 ?


licy183:
You can try to push it to termux-packages. It seems better to go there.

PR #1010, opened 2024-05-10
new package: dotload

okineadev:
I added my `dotload` package to your repository, here is the package page - https://github.com/okineadev/dotload

I have one question - is it possible to auto-update the package?

I would like to automatically update my package in this repository when I create a new release


okineadev:
Thanks!

PR #1017, opened 2024-05-13
new packages: rife-ncnn-vulkan{,-nihui,-tntwise}

knyipab:
managed to build [rife-ncnn-vulkan](https://github.com/nihui/rife-ncnn-vulkan) and tested with: 
```bash
rife-ncnn-vulkan -0 0.png -1 1.png -o test.png -m rife-v4.6 -g -1
```

It works **BUT** it cannot work using gpu (i.e. without `-g -1`), with error below: 
```
SPIR-V WARNING:
    In file ../src/compiler/spirv/spirv_to_nir.c:4862
    Unsupported SPIR-V capability: SpvCapabilityStorageBuffer8BitAccess (4448)
    36 bytes into the SPIR-V binary
validation fail: (type_size(instr->cat6.type) <= 16) == !!((instr->srcs[2])->flags & IR3_REG_HALF)
  -> for instruction: MESA: info: 0000:000: 	stib.base0 ssa_43, ssa_75, hssa_76
```
I have no ideas, but I speculate that even after solving this error, vulkan will likely generates black image (an issue I raised at https://github.com/nihui/rife-ncnn-vulkan/issues/41). 

Anyways, even with this caveat, I still want it to be listed on TUR. Thanks.

### DIFF
--- a/tur/rife-ncnn-vulkan-nihui/build.sh
+++ b/tur/rife-ncnn-vulkan-nihui/build.sh
@@ -1,0 +1,43 @@
+TERMUX_PKG_HOMEPAGE=https://github.com/nihui/rife-ncnn-vulkan
+TERMUX_PKG_DESCRIPTION="RIFE, Real-Time Intermediate Flow Estimation for Video Frame Interpolation implemented with ncnn library"
+TERMUX_PKG_LICENSE="MIT"
+TERMUX_PKG_MAINTAINER="@termux-user-repository"
+TERMUX_PKG_VERSION=20221029
+TERMUX_PKG_SRCURL=git+https://github.com/nihui/rife-ncnn-vulkan
+TERMUX_PKG_DEPENDS="libwebp, vulkan-loader"
+TERMUX_PKG_ANTI_BUILD_DEPENDS="vulkan-loader"
+TERMUX_PKG_BUILD_DEPENDS="vulkan-loader-generic"
+TERMUX_PKG_GIT_BRANCH=20221029
+TERMUX_PKG_BUILD_IN_SRC=true
+TERMUX_PKG_AUTO_UPDATE=true
+TERMUX_PKG_UPDATE_TAG_TYPE="newest-tag"
+TERMUX_PKG_EXTRA_CONFIGURE_ARGS="-DUSE_SYSTEM_WEBP=ON"
+TERMUX_PKG_CONFLICTS="rife-ncnn-vulkan-tntwise"
+
+termux_step_pre_configure () {
+	git submodule update --init --recursive "$TERMUX_PKG_SRCDIR"
+	mv "$TERMUX_PKG_SRCDIR/src/"* "$TERMUX_PKG_SRCDIR/"
+	LDFLAGS+=" -llog -landroid"
+}
+
+termux_step_make_install () {
+	mkdir -p "$TERMUX_PREFIX/opt/rife-ncnn-vulkan"
+	install -Dm700 rife-ncnn-vulkan "$TERMUX_PREFIX/opt/rife-ncnn-vulkan"
+	ln -s "$TERMUX_PREFIX/opt/rife-ncnn-vulkan/rife-ncnn-vulkan" "$TERMUX_PREFIX/bin/rife-ncnn-vulkan"
+	cp -r models/rife-v2.3 "$TERMUX_PREFIX/opt/rife-ncnn-vulkan/rife-v2.3" # default value of model arg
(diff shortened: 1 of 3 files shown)


knyipab:
By the way, little more info, rife-vulkan-ncnn uses older version of glslang so it cannot be built to dynamically link to the `libncnn` package (i.e. `-DUSE_SYSTEM_NCNN=ON` does not compile).

tur/rife-ncnn-vulkan-nihui/build.sh
```diff
@@ -0,0 +1,41 @@
+TERMUX_PKG_HOMEPAGE=https://github.com/nihui/rife-ncnn-vulkan
+TERMUX_PKG_DESCRIPTION="RIFE, Real-Time Intermediate Flow Estimation for Video Frame Interpolation implemented with ncnn library"
+TERMUX_PKG_LICENSE="MIT"
+TERMUX_PKG_MAINTAINER="@termux-user-repository"
+TERMUX_PKG_VERSION=20221029
+TERMUX_PKG_SRCURL=git+https://github.com/nihui/rife-ncnn-vulkan
+TERMUX_PKG_DEPENDS="libwebp, vulkan-loader"
+TERMUX_PKG_ANTI_BUILD_DEPENDS="vulkan-loader"
+TERMUX_PKG_BUILD_DEPENDS="vulkan-loader-generic"
+TERMUX_PKG_GIT_BRANCH="master"
```
licy183: This should point to a tag rather than `master`.

tur/rife-ncnn-vulkan-nihui/build.sh
```diff
@@ -0,0 +1,41 @@
...
+
+termux_step_pre_configure () {
+	git submodule update --init --recursive "$TERMUX_PKG_SRCDIR"
+	mv "$TERMUX_PKG_SRCDIR/src/"* "$TERMUX_PKG_SRCDIR/"
+	LDFLAGS+=" -llog -landroid"
+}
+
+termux_step_make_install () {
+	install -Dm700 rife-ncnn-vulkan $TERMUX_PREFIX/bin
+	cp -r models/rife-v2.3 "$TERMUX_PREFIX/bin/rife-v2.3" # default value of model arg
```
licy183: Installing the models to `$PREFIX/bin` is strange. You can install them to `$PREFIX/opt/$PKG_NAME` and symlink the binary to `$PREFIX/bin`

knyipab:
@licy183 Thanks. Fixed and tested with the command.

knyipab:
I investigated and found few things:
* the `SPIR-V` error should be an issue of `vulkan-loader-generic`. If `vulkan-loader-android` is installed instead. While I doubt if it is fixable or anyone will try to fix `vulkan-lodaer-generc`, should this package be built against `/system/lib(64)/libvulkan.so`?
* it can indeed be built against `libncnn` using `-DUSE_SYSTEM_NCNN=ON` by replacing `include("${GLSLANG_TARGET_DIR}/*.cmake` with `include("${GLSLANG_TARGET_DIR}/glslang-targets.cmake")` in CMakeLists.txt
* in that case, it probably won't need git clone the repo, but use release tar.gz instead. Also, the command line `git submodule update ...` in `build.sh` should be unnecessary. 

Thank you for your advice.

knyipab:
@licy183 two things: 
* No more editing from me and ready for review. Sorry for any confusion and inconvenience caused by my frequent editing. 
* Your PR build action failed. But in fact, there is nothing wrong in my commits. My Actions work ([this](https://github.com/knyipab/tur/actions/runs/9099228085) and [this](https://github.com/knyipab/tur/actions/runs/9099226500)).

knyipab:
Thanks a lot for help!

TNTwise updated the repo to `20240515` after I reflected the cpu mode segfault bug yesterday in https://github.com/TNTwise/rife-ncnn-vulkan/issues/4. 

After that, we may test all commands below and they should work for both `-nihui` and `-tntwise`
```
git clone https://github.com/nihui/rife-ncnn-vulkan
cd ./rife-ncnn-vulkan/images
# GPU mode
rife-ncnn-vulkan -0 0.png -1 1.png -o test.png
rife-ncnn-vulkan -0 0.png -1 1.png -o test.png -m rife-v4.6
rife-ncnn-vulkan -0 0.png -1 1.png -o test.png -m rife-v4-latest
# CPU mode
rife-ncnn-vulkan -0 0.png -1 1.png -o test.png -g -1
rife-ncnn-vulkan -0 0.png -1 1.png -o test.png -m rife-v4.6 -g -1
rife-ncnn-vulkan -0 0.png -1 1.png -o test.png -m rife-v4-latest -g -1
```

PR #1023, opened 2024-05-16
ettercap: move from termux-packages/root-packages

Grimler91:
Ettercap could be considered a dual-use package and could perhaps be put in tur-hacking instead. Plese let me know if that is preferable.


Grimler91:
Aha, right, have to remove from main repo first

licy183:
I think it should be put in `tur-hacking`.

PR #1066, opened 2024-06-06
chore(tur/darktable): rebuild against `libicu` 75.1

TomJo2000:
- termux/termux-packages#20438


TomJo2000:
I'm still unsure what is actually pulling in `libicu` as a transitive dependency, but I can figure that out in the morning after I get some sleep.

TomJo2000:
Thanks for the assist @licy183, I hadn't had time to work on this yet.

PR #1121, opened 2024-07-17
fix(tur-hacking/ettercap): permission denied

1q23lyc45:
Fixed the Permission Denied issue caused by UID downgrading to an illegal user during Ettercap startup. 

(Users who have already flashed the KernelSU GKI kernel do not have this issue. This issue is only reproduced in non GKI system kernels.)


licy183:
Thanks!

PR #1138, opened 2024-08-10
addpkg(tur-hacking/bettercap): v2.33.0

### LINKED ISSUE #1122: [Package]: bettercap
### Name

bettercap

### Home page URL

https://www.bettercap.org/

### Source code URL

https://github.com/bettercap/bettercap

- [X] I understand [Tur Package Policy](https://github.com/termux-user-repository/tur/wiki/TUR-package-policy.).

### Additional information

Without patching, binary files can be compiled directly.

```
git clone https://github.com/bettercap/bettercap
cd bettercap
apt install golang make pkg-config nmap root-repo libusb
export GOPROXY=https://goproxy.io,direct
apt install libnetfilter-queue
make -j8
```

Executable file: ./bettercap

Root is required to run. However, it can run directly without modifying the system. 

### It is important to note that do not execute "mount -o remount,rw /" and "mkdir -p /home/builder/.termux-build/_cache/18-arm-21-v2/bin/" before use, otherwise the phone will become bricked and can only be fixed by clearing data and flashing the device!

1q23lyc45:
Close #1122

https://github.com/1q23lyc45/tur/actions/runs/10330330061 This is the compiled output of CI.

### DIFF
--- a/tur-hacking/bettercap/build.sh
+++ b/tur-hacking/bettercap/build.sh
@@ -1,0 +1,20 @@
+TERMUX_PKG_HOMEPAGE=https://www.bettercap.org/
+TERMUX_PKG_DESCRIPTION="The Swiss Army knife for 802.11, BLE, IPv4 and IPv6 networks reconnaissance and MITM attacks. "
+TERMUX_PKG_LICENSE="GPL-3.0"
+TERMUX_PKG_VERSION="2.33.0"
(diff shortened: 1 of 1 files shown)


1q23lyc45:
I would like to ask why I can use CI compilation here to generate usable deb files, but the runtime result is a red cross, while you also use CI compilation, but the result is a green mark?

PR #1146, opened 2024-08-22
dosemu: new package help wanted

john-peterson:
you might have to specify the dos path

```
export DOSEMU2_COMCOM_DIR=$(HOME)/freedos
dosemu
```


IntinteDAO:
@john-peterson I am a noob, however as I understand by this you just show a place where is COMCOM binary (some kind of software of FreeDOS)

https://github.com/dosemu2/dosemu2/blob/5e9600edae85b388e0faab58c410c47609b51eb0/src/base/init/config.c#L511
I think you can modify this file to set a localization and push your build.sh with patch

PR #1148, opened 2024-08-22
addpkg(tur-hacking/burpsuite{,-pro}): 2023.12.1.5

knyipab:
Version higher than 2023.12.1.5 may require openjdk >= 21. 

LICENSE is from 
* https://portswigger.net/burp/eula/community
* https://portswigger.net/burp/eula/pro

### DIFF
--- a/tur-hacking/burpsuite-pro/LICENSE
+++ b/tur-hacking/burpsuite-pro/LICENSE
@@ -1,0 +1,319 @@
+Burp Suite Professional Terms & Conditions of Supply
+
+Burp Suite Professional Terms & Conditions of Supply
+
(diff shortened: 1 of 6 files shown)


licy183:
I'm not sure whether burpsuite allows third party to redistribute their binaries or not.

knyipab:
I am by no means expert on this. Just note that Kali (official) package burpsuite and AUR packages both versions.  
https://wiki.archlinux.org/title/Burp_Suite
https://www.kali.org/tools/burpsuite/

licy183:
Kali packages burpsuite community so I think it is acceptable to package it too.

AUR actually performs on device build and it is different from TUR, so I think it is better to package an installer script for `burpsuite-pro`.

knyipab:
By installer script, do you mean shell script like this?
```bash
termux_step_create_debscripts() {
	cat <<- EOF > ./postinst
		# ...
	EOF

	cat <<- EOF > ./prerm
		# ...
	EOF
}
```

licy183:
Yes. `burpsuite-pro` can be downloaded in `postinst`.

licy183 (approved):
LGTM. Thanks!

PR #1157, opened 2024-08-30
save data with minimal clone

john-peterson:
this is just a plain download no tags are necessary

if the signal is poor a full clone take ages and drain all battery


licy183:
I don't want to merge this as these commits may be useful. It would be better to patch this script locally.

PR #1158, opened 2024-08-30
add makerom for unpacking installation files to runnable format

john-peterson:


licy183:
If this package only supports `on-device` building, it should be put in `tur-on-device` folder.

And, please ensure that the ident of your `build.sh` is the same as others.

Issue #1174, opened 2024-09-03
Help on bringing libreoffice deps to TUR


knyipab:
As you may know and I mentioned in another PR that I am bringing libreoffice deps to TUR, there are few issues encountered and really grateful for your help
* [Many of the deps packages](https://github.com/knyipab/tur/commit/c68491b095522ad64614c1671ab952009de52feb) are from dev-www.libreoffice.org or sourceforge and cannot be updated automatically. Could TUR by any chance offer auto update mechanism relying on repology? Just like `termux-packages` offers. Perhaps by matching package name. 
* `libcmis` (build script [here](https://github.com/knyipab/tur/commit/5f4cf7f670407c13b410972a43f705afa36d6c5f)) cannot configure to compile a test that uses boost datetime ([this action](https://github.com/knyipab/tur/actions/runs/10654968616/job/29531921960)). I tried various build scripts as well as [on device build](https://github.com/knyipab/tur/actions/runs/10654628581/job/29531186231) but still not working. The error seems to correspond to [this line](https://github.com/tdf/libcmis/blob/3659d32999ff7593662dcf5136bcb7ac15c13f61/m4/boost.m4#L419) in the source code. 
* `clucene` (build script [here](https://github.com/knyipab/tur/commit/e1599c955818095836e4e46027ea92743290bd4c)) goes further with on device build. But the re-declaration error of basic std::string in [x86_64 build](https://github.com/knyipab/tur/actions/runs/10654969336/job/29531923745) and other weird error in [i686 build](https://github.com/knyipab/tur/actions/runs/10654969336/job/29531923608) are really puzzling. 
* `libfreehand` (build script [here](https://github.com/knyipab/tur/commit/eeefe75f27bf3f7dfb2aa4a2ddfb959c70ab4396)) sees `[-Werror,-Wdeprecated-copy-with-user-provided-copy]` in [the builds](https://github.com/knyipab/tur/actions/runs/10655671401/job/29533472912). No ideas what it means and how to solve. 

After all above issues being solved, we should have libreoffice deps on par with archlinux. Your help is much appreciated!

licy183:
`auto-update` is actually difficult currently. I don't know how to solve the rate limit.

knyipab:
How about requesting repology devs to add `TUR`? Just like `termux-pacakges` was added thru this issue: https://github.com/repology/repology-updater/issues/628. That way I suppose repology could generate a single page to status of all packages in TUR and won't trigger rate limit?

[...]

PR #1180, opened 2024-09-08
libreoffice deps

knyipab:
Related #1174

### DIFF
--- a/tur/box2d/0002-unsupported-cpu.patch
+++ b/tur/box2d/0002-unsupported-cpu.patch
@@ -1,0 +1,23 @@
+--- a/src/core.h
++++ b/src/core.h
+@@ -40,9 +40,9 @@
+ #endif
+ 
+ // Define CPU
+-#if defined( __x86_64__ ) || defined( _M_X64 )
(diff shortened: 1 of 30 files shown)


licy183:
Emmm... Do they all have to be explicitly linked to `compiler-rt` statically?

knyipab:
Do you mean the `$($CC -print-libgcc-file-name)` patch? Not all of them but many of them prompt the relevant errors in GitHub actions builds, and I added that only when needed. It seems that all `librevenge`-dependent packages would need that patch.

tur/box2d/0002-unsupported-cpu.patch
```diff
@@ -0,0 +1,13 @@
+diff --git a/src/core.h b/src/core.h
+index fe495c6..baaa5b3 100644
+--- a/src/core.h
++++ b/src/core.h
+@@ -47,7 +47,7 @@
+ #elif defined( __EMSCRIPTEN__ )
+ 	#define B2_CPU_WASM
+ #else
+-	#error Unsupported CPU
```
licy183: This is a little strange... I'll have a test this weekend.

licy183:
Thanks for working on this!

PR #1186, opened 2024-09-10
addpkg(tur-on-device/juice-shop): 17.1.0

knyipab:


licy183:
I'll merge this. If it doesn't work properly I'll move it back to tur-on-device.

PR #1191, opened 2024-09-11
new package stirling-pdf

knyipab:
* follow the practice of AUR to keep working files at `/var/lib/stirling-pdf`
* It is probably better to have a package `ocrmypdf` instead of installing through `pip`, many in `TERMUX_PKG_DEPENDS` are actually for `ocrmypdf`.


licy183:
Yes, I think TUR should package `ocrmypdf` rather than installing it in `postinst`.

knyipab:
I'm actually not so sure the right way to `ocrmypdf` provided that it has quite a number of python python packages (that may need postinst to run pip install I suppose). Otherwise, will need to package them to TUR. 

Actually what is the best practice of packaging python package? Would the termux pkg version mess up with pip?

PR #1217, opened 2024-09-24
[WIP] new package: libreoffice-still

knyipab:
Would be grateful for turning this to be cross compilation. I tried but couldn't even after looking for clues at https://github.com/LibreOffice/core/blob/master/README.cross

Running `make debugrun` on my phone encounters a lot of runtime path issues. Running `make install` will contaminate my environment. While there is no easy way to proot a Termux userspace, perhaps the only clean way to use a deb package.

### DIFF
--- a/tur-on-device/libreoffice-still/0001-configure-error-and-disable-isystem.patch
+++ b/tur-on-device/libreoffice-still/0001-configure-error-and-disable-isystem.patch
@@ -1,0 +1,72 @@
+From 8f05899b0b9ee8817ceec610589341a8b9cb3a4d Mon Sep 17 00:00:00 2001
+From: Ronald Y <knyipab@gmail.com>
+Date: Sun, 22 Sep 2024 09:21:18 +0800
+Subject: [PATCH 1/8] configure-error-and-disable-isystem
+
+Change-Id: I70125356c9306f442fd5c2ec8343b55019cd118b
+---
+ configure.ac | 27 ++++-----------------------
+ 1 file changed, 4 insertions(+), 23 deletions(-)
+
+diff --git a/configure.ac b/configure.ac
+index d31963fce..5bda2889d 100644
+--- a/configure.ac
++++ b/configure.ac
+@@ -4058,26 +4058,7 @@ AC_SUBST(ENABLE_WASM_STRIP)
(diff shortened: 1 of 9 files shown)


knyipab:
@licy183 TUR's tur-on-device facility seems to have unstable connection issue. 

Currently, it is a pain to debug runtime errors and then recompile the code base. After building on my phone, `make debugrun` does not seem to launch the app in right way, requiring some workarounds like 
```
export gb_DBGARGS="-env:BRAND_BASE_DIR=file://$(realpath .)/instdir -env:UserInstallation=file://$(realpath .)/instdir -env:CONFIGURATION_LAYERS=xcsxcu:file://$(realpath .)/instdir/share/registry res:file://$(realpath .)/instdir/share/registry"
```
and just barely makes progress launching the libreoffice app. 

But using `make install` will contaminate my `$TERMUX` directory. Meanwhile, isolated environments like proot Termux userspace or docker-like tool e.g. udocker, or Termux dual app should help a lot but none of them are readily available. Any suggestions?

knyipab:
Libreoffice does not seem to provide `make uninstall`. 
Ref: https://github.com/LibreOffice/core/blob/master/Makefile.in

IntinteDAO:
> Libreoffice does not seem to provide `make uninstall`. Ref: https://github.com/LibreOffice/core/blob/master/Makefile.in

Sad :( In theory you can use DESTDIR variable to install a LibreOffice on other directory, get file list and using rm - delete it if you want.

Or somehow using Virtual Machine

ReeaoX:
> Sad :(.

It's not sad yet; you're see. ``make clean`` is exist as the provide parody of ``make uninstall``. Because it's better to uninstall ``make``-based binary without ``make uninstall``.

With this. You rule the child process of these line:
```
$(1).clean $(1).showdeliverables:
	cd $(SRCDIR)/$(2) && $$(MAKE) $(PARALLELISM_OPTION) $(GMAKE_OPTIONS) $$(patsubst $(1).%,%,$$@)
```

But except. ``make clean`` recursively executed under the child process. The option to run in competition between foreground, and apparently background:
```
echo -ne "<your root password>\n" | /bin/bash --login -c "make clean"
```

robertkirkman:
Now it is in the termux-packages repository

PR #1257, opened 2024-10-13
bump(tur/pacman-contrib): 1.10.6

TomJo2000:
This came up in the Discord, specifically the systemd services not being removed.
So I decided to take a look at a build script and to be diplomatic this was just really damn sloppy.
I overhauled the build script and updated the package to the latest version.

<h1><em></em></h1> 

I don't want to keep banging on about this, but I just want to quickly point out how many problems the old build script had.
- Systemd services were not being removed
- `paclist` was getting shipped despite supposedly being disabled
because accessing an array without an index only returns the first element.
https://github.com/termux-user-repository/tur/blob/d83bcc0b428088a99b4be6d0cc6b5c9c7a46f2cd/tur/pacman-contrib/build.sh#L26-L28
- That entire removal shenanigans should have been put into `termux_step_post_massage()`
Or better yet just use `TERMUX_PKG_RM_AFTER_INSTALL` which is the control variable specifically for removing files from a package.
- Auto updates were only half disable, I've turned them back on as a test.
- No `sv` services were being provided as equivalents to the Systemd services.
- `PKG_CONFIG_PATH` was being exported to its existing default value for no apparent reason

### DIFF
--- a/tur/pacman-contrib/build.sh
+++ b/tur/pacman-contrib/build.sh
@@ -1,33 +1,32 @@
 TERMUX_PKG_HOMEPAGE=https://gitlab.archlinux.org/pacman/pacman-contrib
 TERMUX_PKG_DESCRIPTION="Additional libalpm utilities contributed by Arch community"
 TERMUX_PKG_LICENSE="GPL-3.0"
-TERMUX_PKG_MAINTAINER="@flosnvjx"
-TERMUX_PKG_VERSION=1.7.1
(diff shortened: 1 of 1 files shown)


TomJo2000:
Actually have no idea why those two are being removed.
Both should work just fine.
`checkupdates` might need a small patch, will look into that later.

licy183 (approved):

PR #1277, opened 2024-10-24
Bump: Hangover Wine to 9.20

IntinteDAO:
I built Hangover locally with this version and it works fine. So I assume that it is safe to upgrade Hangover Wine to version 9.20


licy183:
Thanks!

PR #1288, opened 2024-11-02
Add PyPI extra index URL to pip.conf

Efreak:
Add a note with instructions to add the extra index URL to pip.conf.

Not really relevant to this PR, but I personally also have the pip cache set as well, since I update infrequently (*months* in between) and have many packages and frequently forget to flush the cache; this allows Android to manage that.
```
[global]
cache-dir = /data/data/com.termux/cache/pip
```


licy183:
Thanks!

PR #1290, opened 2024-11-03
hangover-wine: move to tur

### LINKED ISSUE #749: [Package]: hangover
### Name

Hangover

### Source code URL

https://github.com/AndreRH/hangover

- [X] I understand [Tur Package Policy](https://github.com/termux-user-repository/tur/wiki/TUR-package-policy.).

### Additional information

We have `wine-stable` package so it should be pretty easy to port `hangover` too. It has support of builtin box64, qemu and fex (dlls) so it can run x86_64 and x86 binaries even in the case if there is no box86/box64 support in termux itself.

licy183:
Closes #749


tur/hangover-wine/build.sh
```diff
@@ -126,3 +132,24 @@ termux_step_make() {
 termux_step_make_install() {
 	make -j $TERMUX_PKG_MAKE_PROCESSES install
 }
+
+termux_step_post_make_install() {
+	# Install FEX-based dlls
+	local _type
+	for _type in wow64fex arm64ecfex; do
+		mkdir -p wow64fex
+		cd wow64fex
```
Trass3r: Was this intended to be _type?

licy183: Emmm... Forget to change this...

PR #1313, opened 2024-11-13
fixing some packages and improving yml scripts

Maxython:
- made minor changes to `electron` and `rustc-nightly` packages so they pass the check by `scripts/lint-packages.sh` (fixing https://github.com/termux-pacman/tur/issues/2 and https://github.com/termux-pacman/tur/issues/5)
- changed ubuntu version (from 22.04 to 24.04) in yml scripts
- added `big-pkgs.list` file for correct operation of mirror repository

I'll note that it's not just `electron` and `rustc-nightly` that have compilation issues. There are also issues with `supertuxkart`, `libandroid-mathlib`, `electron-headers-for-code-oss` and `dd_rescue` (though with the latter, it's probably a problem with the Termux-Pacman service algorithms). You can find all the information here - https://github.com/termux-pacman/tur/issues.

### DIFF
--- a/.github/workflows/packages.yml
+++ b/.github/workflows/packages.yml
@@ -39,7 +39,7 @@ on:
 
 jobs:
   build:
-    runs-on: ubuntu-22.04
+    runs-on: ubuntu-24.04
     env:
       ANDROID_HOME: "/opt/termux/android-sdk"
       NDK: "/opt/termux/android-ndk"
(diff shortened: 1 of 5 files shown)


.github/workflows/packages.yml
```diff
@@ -170,7 +170,7 @@ jobs:
         done
     - name: Free additional disk space (if necessary)
       run: |
-        if grep -Eq '^(chromium|code-server|code-oss|electron-headers-.*|rustc-nightly|python-polars)$' ./built_tur_packages.txt; then
+        if grep -Eq "^($(paste -s -d '|' ./big-pkgs.list))$" ./built_tur_packages.txt; then
           sudo apt purge -yq $(dpkg -l | grep '^ii' | awk '{ print $2 }' | grep -P '(cabal-|dotnet-|ghc-|libmono|php|aspnetcore)') \
```
licy183: Should update these packages if the host VM bumps to Ubuntu 24.04.

licy183 (changes requested):

.github/workflows/packages.yml
Maxython: If you are afraid that some package will not be found, this is not true. In termux-packages when the ubuntu version was updated, the packages that are removed to free up space were not updated. You can see it [here](https://github.com/termux/termux-packages/blob/master/.github/workflows/packages.yml#L203).

Or do you want to add other packages to free up more space?

licy183: IIRC it will not work, see https://github.com/termux/termux-packages/commit/c392e1bf75eda1b01075b5e1ac3e1de3340c3a99. Better to copy them and keep up to date with the main repo.

Maxython: > IIRC it will not work, see https://github.com/termux/termux-packages/commit/c392e1bf75eda1b01075b5e1ac3e1de3340c3a99

There (in the termux-packages repository) was a problem with removing the `google-cloud-cli` package and the `powershell` package. As far as I know, the yml scripts that build packages in tur do not remove these packages. The exception is `package_on_device.yml` and `package_update.yml`, since the first one uses your own VM, and the second one does not have updated ubuntu (i.e. it uses version 22.04).

> Better to copy them and keep up to date with the main repo.

That is, you want to add other packages that will be removed to free up space?

licy183: Yes. It is better to sync with the main repository.

licy183 (approved):

licy183:
Thanks!

PR #1423, opened 2024-12-27
Dosemu tur

stsp:
Hi guys!

This is my attempt to build dosemu2
under termux. I haven't tried dosemu1
for that task, as it doesn't support aarch64,
whereas dosemu2 does.
Unfortunately, dosemu2 is a monster,
(while dosemu1 is very small), so I spent
almost a month figuring out its deps and
packing it up into termux... I don't even
know if people will appreciate that amount
of new packages, all of which are just a
build-deps. :( I actually only added the
most essential deps - it had much more,
I removed some.

Status:
- Builds on CI: from termux, from tur and the hacked-up my own
- Tested on waydroid/x86_64 only, so please test on aarch64!
- All acceleration (jit, kvm, native execution on x86 host) - disabled
- pthread_cancel() replaced with pthread_kill(), so on exit you'll see messages like "terminated on signal" etc

I'd say its a pure miracle that it even works.
But I believe it suits termux better than
dosbox does, because it can flawlessly work
on text terminal without gfx mode.
Which termux is all about. :)


stsp:
Here's the demo of dosemu2 on terminal:
https://www.youtube.com/watch?v=_seqxpqvuXs
As can be seen, its not just for simple text -
most of complex TUI things actually work.

stsp:
Oh, can someone please tell me what
is this "dry-run" thing? I only made sure
the packages can be built.

stsp:
I removed the `.git` suffixes from URLs.
I suspect this is what makes dry-run to
fail... But only the trial can tell.

tur/comcom64/build.sh
```diff
@@ -0,0 +1,28 @@
+TERMUX_PKG_HOMEPAGE=https://github.com/dosemu2/comcom64
+TERMUX_PKG_DESCRIPTION="64bit command.com"
+TERMUX_PKG_LICENSE="GPL-3.0"
+TERMUX_PKG_MAINTAINER="@stsp"
+TERMUX_PKG_VERSION="0.3"
+TERMUX_PKG_SRCURL=git+https://github.com/dosemu2/comcom64
+TERMUX_PKG_GIT_BRANCH=master
```
licy183: `master` it not valid. If it uses thunk version, you should checkout specific commit in `termux_step_post_get_source`. See the example below.

https://github.com/termux-user-repository/tur/blob/7e0f324ef3a38551182b9e68c134753586064a31/tur/vkmark/build.sh#L15-L36.

tur/fdpp/build.sh
```diff
@@ -0,0 +1,26 @@
+TERMUX_PKG_HOMEPAGE=https://github.com/dosemu2/fdpp
+TERMUX_PKG_DESCRIPTION="64-bit DOS core"
+TERMUX_PKG_LICENSE="GPL-3.0"
+TERMUX_PKG_MAINTAINER="@stsp"
+TERMUX_PKG_VERSION=1.8
+TERMUX_PKG_SRCURL=git+https://github.com/dosemu2/fdpp
+TERMUX_PKG_GIT_BRANCH=master
```
licy183: Ditto

tur/instfd/build.sh
```diff
@@ -0,0 +1,14 @@
+TERMUX_PKG_HOMEPAGE=https://github.com/dosemu2/install-freedos
+TERMUX_PKG_DESCRIPTION="FreeDOS installed for dosemu2"
+TERMUX_PKG_LICENSE="GPL-3.0"
+TERMUX_PKG_MAINTAINER="@stsp"
+TERMUX_PKG_VERSION=0.1
+TERMUX_PKG_SRCURL=git+https://github.com/dosemu2/install-freedos
+TERMUX_PKG_GIT_BRANCH=master
```
licy183: Ditto

tur/nasm-segelf/build.sh
```diff
@@ -0,0 +1,18 @@
+TERMUX_PKG_HOMEPAGE=https://github.com/stsp/nasm-segelf
+TERMUX_PKG_DESCRIPTION="nasm fork with segelf patches"
+TERMUX_PKG_LICENSE="BSD 2-Clause"
+TERMUX_PKG_MAINTAINER="@stsp"
+TERMUX_PKG_VERSION="2.16.01"
+TERMUX_PKG_SRCURL=git+https://github.com/stsp/nasm-segelf
+TERMUX_PKG_GIT_BRANCH=master
```
licy183: Ditto

tur/dj64dev/build.sh
```diff
@@ -0,0 +1,25 @@
+TERMUX_PKG_HOMEPAGE=https://github.com/stsp/dj64dev
+TERMUX_PKG_DESCRIPTION="development suite that allows to cross-build 64-bit programs for DOS"
+TERMUX_PKG_LICENSE="GPL-3.0"
+TERMUX_PKG_MAINTAINER="@stsp"
+TERMUX_PKG_VERSION=0.2
+TERMUX_PKG_SRCURL=git+https://github.com/stsp/dj64dev
+TERMUX_PKG_GIT_BRANCH=master
```
licy183: Ditto

tur/djstub/build.sh
```diff
@@ -0,0 +1,31 @@
+TERMUX_PKG_HOMEPAGE=https://github.com/stsp/djstub
+TERMUX_PKG_DESCRIPTION="go32-compatible stub that supports COFF, PE and ELF payloads"
+TERMUX_PKG_LICENSE="GPL-3.0"
+TERMUX_PKG_MAINTAINER="@stsp"
+TERMUX_PKG_VERSION=0.2
+TERMUX_PKG_SRCURL=git+https://github.com/stsp/djstub
+TERMUX_PKG_GIT_BRANCH=main
```
licy183: Ditto

tur/dosemu2/build.sh
```diff
@@ -0,0 +1,17 @@
+TERMUX_PKG_HOMEPAGE=https://github.com/dosemu2/dosemu2
+TERMUX_PKG_DESCRIPTION="Run DOS programs under linux."
+TERMUX_PKG_LICENSE="GPL-2.0"
+TERMUX_PKG_MAINTAINER="@stsp"
+TERMUX_PKG_VERSION="2.0pre9-git"
+TERMUX_PKG_SRCURL=git+https://github.com/dosemu2/dosemu2
+TERMUX_PKG_GIT_BRANCH=devel
```
licy183, line 7: Ditto

tur/thunk-gen/build.sh
```diff
@@ -0,0 +1,49 @@
+TERMUX_PKG_HOMEPAGE=https://github.com/stsp/thunk_gen
+TERMUX_PKG_DESCRIPTION="thunk generator for C and assembler"
+TERMUX_PKG_LICENSE="GPL-3.0"
+TERMUX_PKG_MAINTAINER="@stsp"
+TERMUX_PKG_VERSION=1.2
+TERMUX_PKG_SRCURL=git+https://github.com/stsp/thunk_gen
+TERMUX_PKG_GIT_BRANCH=main
```
licy183: Ditto

tur/ctags-cross/build.sh
```diff
@@ -0,0 +1,41 @@
+TERMUX_PKG_HOMEPAGE=https://ctags.io/
+TERMUX_PKG_DESCRIPTION="Universal ctags: Source code index builder"
+TERMUX_PKG_LICENSE="GPL-2.0"
+TERMUX_PKG_MAINTAINER="@termux-user-repository"
+TERMUX_PKG_VERSION="3-6.1.0"
```
licy183:
```suggestion
TERMUX_PKG_VERSION="6.1.0"
```

licy183:
As you're the maintainer of these packages, I'll merge this PR after these things are get resolved. Thanks for your contribution!

tur/comcom64/build.sh
stsp: Do I really need to do this, or maybe
just putting commit sha as branch is
enough?

stsp: Also I see packages specify
`TERMUX_PKG_SHA256` in case of
the particular commit. What kind
of sha is this? I only know how to
calc sha of a release zip file.

stsp: Done.

tur/ctags-cross/build.sh
stsp: Since ctags is also present in termux,
I set it to "2:6.1.0" now to match termux.

stsp:
> As you're the maintainer of these packages, I'll merge this PR after these things are get resolved.

I addressed your comments.
As for maintaining - I only tested
these packages under waydroid/x86_64.
If setting the maintenance to
`@termux-user-repository` helps
better testing, I can do that.

Status update:
- enabled jit on x86 platforms
- enabled LTO

So mostly this is a fully-functional build,
just with GUI stripped off (termux doesn't
need GUI, it needs TUI). It seems, for arm
platforms there is no JIT even in upstream
version, so for now this is probably the
best what I could package.

licy183:
LGTM. Thanks!

PR #1426, opened 2025-01-01
Tur fix

stsp:
After my previous dosemu2 patchset
was merged into tur, my own CI started
to fail. This is because it is derived from
tur CI, and now, instead of building all
packages, it was downloading most of
them in a pre-built form.
This quickly revealed the wrong deps
and other oddities.
Here is the first round of fixes. My CI
for now passes again, but no guarantee
it won't find something new, once this
is applied. :)


stsp:
Done.

licy183:
Thanks!

Issue #1445, opened 2025-01-17
[Bug]: ERROR: Duplicated package: vkmark


stsp:
### Problem description

I am using my own tur-based CI.
Today it started to fail:
https://github.com/stsp/tur/actions/runs/12829391617/job/35775356140?pr=1
```
ERROR: Duplicated package: vkmark
ERROR: Obtaining buildorder failed
```
I think the problem is caused by https://github.com/termux/termux-packages/commit/ada8270bc9cb22a26a6ee730766323fc8b08b287

Likely the vkmark package should
be rev-bumped or removed from tur?

### What steps will reproduce the bug?

Not sure, the problem is reproducible
on my own CI so far.

### System information

termux-info:

```

```

truboxl:
I think better merge TUR vkmark into main x11 repo before removing in TUR

stsp:
> I think better merge TUR vkmark into main x11 repo before removing in TUR

What does this mean, are those
vkmarks different? There is one
already in termux/x11, which caused
the problem.
Anyway, seems to late. :)

PR #1446, opened 2025-01-17
remove vkmark

stsp:
This fixes

ERROR: Duplicated package: vkmark

... on CI.

The problem is because vkmark package already exist in termux repo. So removing it from tur seems plausible.
Fixes https://github.com/termux-user-repository/tur/issues/1445


licy183:
Thanks!

PR #1472, opened 2025-01-29
hangover-wine: bump to 10.0

IntinteDAO:
Works on Termux without any problems


licy183:
Thanks! Will update `wine-stable/devel/staging` later

PR #1548, opened 2025-02-28
addpkg(tur/payload-dumper-go): 1.3.0

1q23lyc45:
This package only supports ARM and ARM64 and does not support other architectures.


licy183:
The build failure is caused by the prebuilt libzstd.a from `gozstd`. I'll fix it if I have more free time.

PR #1585, opened 2025-03-13
Newpkg: Supertux

IntinteDAO:


IntinteDAO:
@robertkirkman

robertkirkman:
do you know how to enable touchscreen buttons in this version? 

also, Do you know how to make it buildable in `TERMUX_ON_DEVICE_BUILD=true`? 

also, I think that probably this variable should not be outside of a function in a `build.sh`, because there is no other package that has this variable outside of a function in a `build.sh`.

IntinteDAO:
> do you know how to enable touchscreen buttons in this version?
> 
> also, Do you know how to make it buildable in `TERMUX_ON_DEVICE_BUILD=true` mode?
> 
> also, I think that probably this variable `TERMUX_ON_DEVICE_BUILD` should not be outside of a function in a `build.sh`, because there is no other package that has this variable outside of a function in a `build.sh`.

1) The question is whether we want it to be touchscreen by default. Generally Supertux is also on Android directly. This version can be “like the PC version”. But I don't want to decide it myself.

2)  The problem is that there is no OpenSLES in Android for on-device building. If you comment out LDFLAGS, the compilation will execute successfully.

3) Forgive me, I don't know much about building packages. What should I change from TERMUX_ON_DEVICE_BUILD?

robertkirkman:
1. sometimes, there is something called an "ubuntu tablet"

image

generally, I prefer to partially pretend that Termux:X11 is probably not a desktop PC, but instead directly compare it to the "ubuntu tablet" **for the purposes of touchscreen buttons support**.

The thought goes like this: If an X11 application (which is internally using almost exactly the same touchscreen + windowing APIs, driven by the same touchscreen application-side backend code, on an "ubuntu tablet", as it would on Termux:X11) does have the ability to have touchscreen automatically enabled somehow, then it should do so, because the "ubuntu tablet" user would definitely prefer that to be the case.

The keyboard functionality, or gamepad functionality, should also automatically activate when an external keyboard, or gamepad input, is used, respectively.

2. could you tell me whether this command is successful for you or not on your device?

`./build-package.sh -I -f luanti`

I have been testing this command with my devices, and it has been working, but if on your device, the OpenSLES linking step has an error, then I would like to know about that so I can find out what is inconsistent about it.

3. You can simply remove that line, and then make sure that the package builds on-device with the same compatibility as it does in the cross-compiler, and then everything will be fine.

IntinteDAO:
> generally, I prefer to partially pretend that Termux:X11 is probably not a desktop PC, but instead directly compare it to the "ubuntu tablet" for the purposes of touchscreen buttons support.

The thought goes like this: If an X11 application (which is internally using almost exactly the same touchscreen + windowing APIs, driven by the same application-side touchscreen backend code, on an "ubuntu tablet", as it would on Termux:X11) does have the ability to have touchscreen automatically enabled somehow, then it should do so, because the "ubuntu tablet" user would definitely prefer that to be the case.

The keyboard, mouse, or gamepad functionality, should also automatically activate when an external keyboard, mouse, or gamepad input, is used, respectively.

___

I understand your point of view, although I think Termux rather “emulates” desktop solutions. At least for now. It doesn't support touch screens on a screen-by-screen basis (multi-touch, etc.), but on a mouse-by-mouse basis.

___

2) Yes, luanti compilation works from your script, but so does Supertux. On-device compilation, on the other hand, does not work for either Supertux or Luanti.

robertkirkman:
> I understand your point of view, although I think Termux rather “emulates” desktop solutions. At least for now. It doesn't support touch screens on a screen-by-screen basis (multi-touch, etc.), but on a mouse-by-mouse basis.

this is not correct. in Termux:X11 settings, touch the "direct touch" setting. This is multitouch support.

image

> Yes, luanti compilation works from your script, but so does Supertux. On-device compilation, on the other hand, does not work for either Supertux or Luanti.

could you send the result error of what happens from this command for you?

```bash
./build-package.sh -I -f luanti
```
on-device, like I mentioned, for your device,
since on my Samsung Galaxy A70 SM-A705FN, and other devices, that works.

IntinteDAO:
> this is not correct. in Termux:X11 settings, touch the "direct touch" setting. This is multitouch support.

Ok, didn't know that.

> could you send the result error of what happens from this command for you?

Not an error - just a loop
```
CPAN build and cache directory? [/pzstd                          packages/zstd] /pzstd                          packages/zstd
CPAN build and cache directory? [/pzstd                          packages/zstd] /pzstd                          packages/zstd
CPAN build and cache directory? [/pzstd                          packages/zstd] /pzstd                          packages/zstd
CPAN build and cache directory? [/pzstd                          packages/zstd] /pzstd                          packages/zstd
CPAN build and cache directory? [/pzstd                          packages/zstd] /pzstd                          packages/zstd
CPAN build and cache directory? [/pzstd                          packages/zstd] /pzstd                          packages/zstd
CPAN build and cache directory? [/pzstd                          packages/zstd] /pzstd                          packages/zstd

```

robertkirkman:
I can reproduce that, but I do know that it is not exactly directly related to luanti, well, it kind of is, but it's debatable. It happens because of code that is outside of the `luanti` folder and cannot be really fixed without reorganizing code somewhere else in the termux-packages repository. I will try to fix it in the future but not right now.

The text `CPAN build and cache directory?` only happens if you try to use `build-package.sh` on any package that depends on the package `xdg-utils` on-device while not already ever having had the package `xdg-utils` installed before trying to build. Unfortunately `luanti` is the first package in termux-packages that has a legitimate runtime dependency on `xdg-utils` (unless there are other packages I haven't noticed that have runtime failures because of not being patched to depend on `xdg-utils`), so it is the only package for which the edge case of `build-package.sh` trying to install `xdg-utils` during dependency fetching is reachable.

The code that is entering an infinite loop for you is somewhere inside this command that runs during the `postinst` script of `xdg-utils`.

```bash
cpan -fi File::MimeInfo
```

https://github.com/termux/termux-packages/blob/4c8fe0c5784e2b43fcdef579ff2d54484a2e37ad/x11-packages/xdg-utils/build.sh#L36

I know that because the text `CPAN build and cache directory?` only appears inside the code of perl, not anywhere else,

https://github.com/Perl/perl5/blob/27e7aaa9d0cc16ae5413e08b6dd5ab61cc839652/cpan/CPAN/lib/CPAN/FirstTime.pm#L1706

and the command `cpan` is provided by the package `perl`.

The way to reproduce this infinite loop is using these commands:

```bash
pkg remove -y xdg-utils
rm -rf ~/.cpan
cd tur/
./setup-environment.sh
scripts/setup-termux.sh
./build-package.sh -I -f luanti
```

Could you use this commands instead to make sure that you do not reach that infinite loop case?

```
pkg upgrade -y
pkg install xdg-utils
cd tur/
git pull origin master
./setup-environment.sh
scripts/setup-termux.sh
./build-package.sh -I -f luanti
```

tur/supertux/build.sh
```diff
@@ -0,0 +1,24 @@
+TERMUX_PKG_HOMEPAGE=https://www.supertux.org
+TERMUX_PKG_DESCRIPTION="SuperTux is a jump'n'run game with strong inspiration from the Super Mario Bros. games for the various Nintendo platforms."
+TERMUX_PKG_LICENSE="GPL-3.0"
+TERMUX_PKG_MAINTAINER="@termux-user-repository"
+TERMUX_PKG_VERSION="0.6.3"
+TERMUX_PKG_SRCURL="https://github.com/SuperTux/supertux/releases/download/v${TERMUX_PKG_VERSION}/SuperTux-v${TERMUX_PKG_VERSION}-Source.tar.gz"
+TERMUX_PKG_SHA256=f7940e6009c40226eb34ebab8ffb0e3a894892d891a07b35d0e5762dd41c79f6
+TERMUX_PKG_BUILD_IN_SRC=true
+TERMUX_PKG_BUILD_DEPENDS="boost-headers"
+TERMUX_PKG_DEPENDS="boost, glm, sdl2, sdl2-image, sdl2-ttf, glew, openal-soft, libphysfs, supertux-data, freetype, libandroid-execinfo"
```
robertkirkman: While on the subject of it, this `supertux` package also depends on `xdg-utils`, because it makes calls to `system("xdg-open")`, these need to be patched to use `xdg-utils-xdg-open` from the `xdg-utils` package instead of `xdg-open` from the `termux-tools` package, because `xdg-open` from `termux-tools` does not work on folders, simply nothing happens, while `xdg-utils-xdg-open` from the `xdg-utils` package works on folders, this is why `luanti` depends on that package. This would cause `supertux` to also cause `build-package.sh` to enter the infinite loop if used in `$TERMUX_ON_DEVICE_BUILD=true` mode on a device where `xdg-utils` has never been installed before, but if you don't do this, then this button will not work in-game.

image

https://github.com/SuperTux/supertux/blob/c1ddb4f28c54de77f07aa9965231a91c0ed06708/src/util/file_system.cpp#L237

```diff
--- a/src/util/file_system.cpp
+++ b/src/util/file_system.cpp
@@ -234,7 +234,7 @@ void open_path(const std::string& path)
   #if defined(__APPLE__)
   std::string cmd = "open \"" + path + "\"";
   #else
-  std::string cmd = "xdg-open \"" + path + "\"";
+  std::string cmd = "xdg-utils-xdg-open \"" + path + "\"";
   #endif
 
   int ret = system(cmd.c_str());
```

robertkirkman (changes requested):

tur/supertux/build.sh
IntinteDAO: Fixed :)

IntinteDAO:
@robertkirkman I have one question. Can you jump (using space) in game? I don't know that I have a problem with the game or Termux X11 or something else.

IntinteDAO:
@hansm629 Try now

IntinteDAO:
> @IntinteDAO It works very well! Screenshot_2025-04-17_00-23-25 Screenshot_2025-04-17_00-25-20

Does space button works to you or you need to remap a key?

hansm629:
@IntinteDAO 
Sorry for the late reply. Both the jump key and the attack key work fine!
Screenshot_2025-04-19_10-50-43

IntinteDAO:
@robertkirkman I think the bug with keyboard is fixed, please upgrade Termux X11 and confirm. It also works on Luanti

robertkirkman:
> @robertkirkman I think the bug with keyboard is fixed, please upgrade Termux X11 and confirm. It also works on Luanti

In the build of SuperTux I downloaded from the artifacts here just now, on Samsung Galaxy A70 with Termux:X11 APK version 20250410 and `termux-x11-nightly` version 1.03.01-4, by default the Space bar on an external USB keyboard is still not working with Jump by default, however, I have just noticed that Space does begin working in SuperTux for jumping IF "prefer scancodes when possible" is set to enabled in Termux:X11 preferences. I don't think I had tested that setting in SuperTux previously, and I had only used it for Luanti.

Regarding Luanti, what version of Luanti did you test, version 1:5.11.0-1, or version 1:5.11.0-2?

Version  1:5.11.0-2 is a version that I have patched specifically to force the Space bar to work for jump by default with no custom keybind required, fixing the problem that now seems to only exist in Luanti.

- https://github.com/termux/termux-packages/pull/24231

IntinteDAO:
5.11.0-2 This one. So probably a patch is merged. But still something fix a issue, maybe this one:
https://github.com/termux/termux-x11/commit/295381c5df9a89860d0920f853c2e12843e7b15f

But still, we need to downgrade luanti, check and then maybe contact with twaik to fix a issue on USB Keyboard if needed

robertkirkman:
> 5.11.0-2 This one. So probably a patch is merged. But still something fix a issue, maybe this one: [termux/termux-x11@295381c](https://github.com/termux/termux-x11/commit/295381c5df9a89860d0920f853c2e12843e7b15f)
> 
> But still, we need to downgrade luanti, check and then maybe contact with twaik to fix a issue on USB Keyboard if needed

I tested that PR before it was merged and described what happened for me here, you can read about what I saw in Luanti there:

- https://github.com/termux/termux-x11/pull/768#issuecomment-2692202601

if you downgrade luanti, but keep the current Termux:X11 version, you should be able to reproduce the exact behavior i explained there (because at the time I wrote that, was using the older Luanti version, but using the then-future Termux:X11 version)

robertkirkman:
Oh, but I only explained about the Backspace there. Ok, specifically regarding Space and 9-Key: if you downgrade Luanti and use any known Termux:X11 version (older or newer), then Space and 9-Key will have swapped behavior, meaning that 9-Key will jump and Space will type "9" symbol. but in the current Luanti from `pkg install` I forcibly changed that by applying a patch to the package and merging it into termux-packages repository.

robertkirkman:
Originally I thought that would also affect SuperTux, but it seems like that specific problem only affects Luanti as far as is currently known.

IntinteDAO:
@robertkirkman @hansm629 
Does "space" button still works for you?

robertkirkman:
It does, if "prefer scancodes when possible" is set to enabled.

SuperTux 7 which will support touchscreen buttons in X11 (as seen in the version I packaged as a test) might take a long time to officially release, so this could be added initially in TUR and then after a little while if SuperTux 7 does not officially release, I can open a PR here which bumps to the development commit and cherry picked upstream PR, and fixes any other problems.

For example another problem that this package is supposed to depend on `xdg-utils`, but if it were marked for it, clean on-device building would get an infinite loop because of the bug in the `postinst` script of the `xdg-utils` package. really the correct solution should be to rewrite the `postinst` script to of `xdg-utils` to make it not buggy

robertkirkman (approved):
IMO this game is pretty much playable like this IF you have have a keyboard, the problems are really small details that can be improved later.
- I have a different version of this package with some differences, but they aren't that important to have in a single PR.
- If this game ends up being approved, then soon afterward I would play it, and make a different PR later with any changes I want to apply (like unofficial touchscreen buttons from the PR to the upstream repository, which I tested works as the solution to touchscreen buttons for this game) https://github.com/SuperTux/supertux/pull/3021,
- and those other changes could be discussed at that time.

robertkirkman:
Ok the infinite loop should now be actually fixed, so you can add `xdg-utils` to dependencies here and it should work without that problem anymore

IntinteDAO:
@robertkirkman 
Supertux 0.7.0-beta1

Known issues:
* Sound not working. I believe (I have this in other game also, but I'm not sure about it) that old LDFlags Patch (-Wl,--no-as-needed,-lOpenSLES,--as-needed) not work now.
* Probably some patches can be better
* Supertux can into termux-packages?

tur/supertux/build.sh
```diff
@@ -0,0 +1,34 @@
+TERMUX_PKG_HOMEPAGE=https://www.supertux.org
+TERMUX_PKG_DESCRIPTION="SuperTux is a jump'n'run game with strong inspiration from the Super Mario Bros. games for the various Nintendo platforms."
+TERMUX_PKG_LICENSE="GPL-3.0"
+TERMUX_PKG_MAINTAINER="@IntinteDAO"
+TERMUX_PKG_VERSION="0.7.0-beta.1"
```
robertkirkman: Write it like this

```
TERMUX_PKG_VERSION="0.7.0~beta.1"
```

and `TERMUX_PKG_SRCURL` like this

```
TERMUX_PKG_SRCURL="https://github.com/SuperTux/supertux/archive/refs/tags/v${TERMUX_PKG_VERSION//~/-}.tar.gz"
```

this helps auto update work properly, after that you can write `TERMUX_PKG_AUTO_UPDATE=true` and it will work

tur/supertux/build.sh
```diff
@@ -0,0 +1,34 @@
...
+"
+
+termux_step_pre_configure() {
+	export LDFLAGS+=" -Wl,--no-as-needed,-lOpenSLES,--as-needed -landroid-spawn -llog"
+	export CMAKE_PREFIX_PATH=$TERMUX_PREFIX
+
+	git clone --recurse-submodules https://github.com/SuperTux/tinygettext external/tinygettext
+	git clone --recurse-submodules https://github.com/SuperTux/simplesquirrel external/simplesquirrel
+	git clone --recurse-submodules https://github.com/SuperTux/sexp-cpp external/sexp-cpp
+	git clone --recurse-submodules https://github.com/SuperTux/SDL_ttf external/SDL_ttf
```
robertkirkman: later I can try to help fix audio and this (better for `git clone` commands to be in `termux_step_get_source()` and organized differently, I showed how here):

https://github.com/robertkirkman/tur/commit/40edef6d8713e17257ae503345155e95b6fd56dd

robertkirkman:
> Sound not working. I believe (I have this in other game also, but I'm not sure about it) that old LDFlags Patch (-Wl,--no-as-needed,-lOpenSLES,--as-needed) not work now.

That is caused by this:

- https://github.com/termux/termux-packages/pull/27775

I will try to bisect the problem.

tur/supertux/0002-force-linux-build.patch
```diff
@@ -0,0 +1,137 @@
...
++#if 0
+ #include <android/log.h>
+ #endif
+ 
+@@ -27,7 +27,7 @@
+ #include "supertux/gameconfig.hpp"
+ #include "supertux/globals.hpp"
+ 
+-#ifdef __ANDROID__
++#if 0
```
robertkirkman: If you looked at my version I have been posting here earlier, https://github.com/robertkirkman/tur/commit/40edef6d8713e17257ae503345155e95b6fd56dd, you would have known that I resolved many of these using a way with much less code by using `find` `xargs` `sed`,

did you try that here and find that it was no longer suitable, or do you just prefer to do it this way instead?

tur/supertux/build.sh
```diff
@@ -0,0 +1,39 @@
...
+TERMUX_PKG_BUILD_DEPENDS="glm"
+TERMUX_PKG_DEPENDS="fmt, freetype, glew, libandroid-execinfo, libandroid-spawn, libandroid-stub, libcurl, libogg, libphysfs, libpng, libvorbis, openal-soft, sdl2, sdl2-image, supertux-data, xdg-utils, zlib"
+TERMUX_PKG_FORCE_CMAKE=true
+TERMUX_PKG_GROUPS="games"
+
+TERMUX_PKG_EXTRA_CONFIGURE_ARGS="
+-DCMAKE_SYSTEM_NAME=Linux
+-DIS_SUPERTUX_RELEASE=true
+-DINSTALL_SUBDIR_BIN=bin
+-DINSTALL_SUBDIR_SHARE=share/games/supertux2
```
robertkirkman: why does this need to be changed from the default? where does it go by default?

IntinteDAO: Well—I think the path was different in the beta, or something was different. I deleted it.

IntinteDAO:
Closing (forever I think) - try to push to termux-packages

PR #1593, opened 2025-03-14
migrate(tur/python-yt-dlp): Promote to main repository

TomJo2000:
Moves `python-yt-dlp` to the main repository.
It is a popular package and warrants the extra accessibility from being in the main repository.

- Corresponding termux-packages PR: https://github.com/termux/termux-packages/pull/23789


TomJo2000:
- Follwing termux/termux-packages#23789 I am removing the migrated packages from the TUR.

PR #1600, opened 2025-03-16
Add pkg: itstool

lunsokhasovan:


robertkirkman:
This can come from `pip install itstool`, but this is TUR where there are many such packages, so I guess it is good for it to be here

lunsokhasovan:
> This can come from `pip install itstool`, but this is TUR where there are many such packages, so I guess it is good for it to be here

Oh, I didn't notice.

PR #1601, opened 2025-03-16
addpkg(tur/stash): 0.27.2

aicynide:


licy183:
It fails to build with `go1.24`

PR #1625, opened 2025-03-27
New pkg: libical-glib

lunsokhasovan:
~~I copied from [termux/termux-packages](https://github.com/termux/termux-packages/blob/master/packages/libical/build.sh) and modify to build libical-glib.~~ I'm not sure, it will build another libical, but with libical-glib.

Edited: ~~Nevermind~~
2nd Edited: This pkg for `evolution-data-server`.
3rd & 4th Edited: Nevermind, recode soon.


lunsokhasovan:
It's bad idea! Maybe move to termux/termux-pack with create subpackage?

PR #1665, opened 2025-04-10
New package: 3dstool

NoNameWasDefined:
Fork with patches for Termux from @john-peterson

GitHub actions' build works on device and cross-compilation


tur/3dstool/build.sh
```diff
@@ -0,0 +1,13 @@
+TERMUX_PKG_HOMEPAGE=https://github.com/dnasdw/3dstool
+TERMUX_PKG_DESCRIPTION="An all-in-one tool for extracting/creating 3ds roms"
+TERMUX_PKG_LICENSE="MIT"
+TERMUX_PKG_MAINTAINER="@termux-user-repository"
+TERMUX_PKG_VERSION="1.2.6"
+TERMUX_PKG_SRCURL=git+https://github.com/john-peterson/3dstool/
```
licy183: For binary distribution repositories, it is not appropriate to use a third-party fork. The general practice is to generate patches from commits and add them.

NoNameWasDefined: I've tried patching since the last release but it's not enough, it's beyond my level.

Issue #34, opened 2022-07-16
[Package request] code server


T-Dynamos:
- [x] I understand [TUR package policy](https://github.com/termux-user-repository/tur/wiki/TUR-package-policy)

**Describe package:**
Name:code-server
Homepage:https://github.com/coder/code-server
Source link:https://github.com/coder/code-server/archive/refs/tags/v4.5.0.tar.gz

**Additional Info:**
I don't know how to build nodejs packages

2096779623:
```
~/code-server-4.5.0 $ yarn install
yarn install v1.22.19
[1/5] Validating package.json...
[2/5] Resolving packages...
[3/5] Fetching packages...
[4/5] Linking dependencies...
warning " > @coder/logger@1.1.16" has unmet peer dependency "@google-cloud/logging@^4.5.2".
[5/5] Building fresh packages...
[1/3] ⠈ argon2
[2/3] ⠈ yarn
error /data/data/com.termux/files/home/code-server-4.5.0/node_modules/shellcheck: Command failed.
Exit code: 2
Command: ./install.sh
Arguments: 
Directory: /data/data/com.termux/files/home/code-server-4.5.0/node_modules/shellcheck
Output:
xz: (stdin): File format not recognized
tar: Child died with signal 13
tar: Error is not recoverable: exiting now
node:events:505
      throw er; // Unhandled 'error' event
      ^

Error: write EPIPE
    at afterWriteDispatched (node:internal/stream_base_commons:160:15)
    at writeGeneric (node:internal/stream_base_commons:151:3)
    at Socket._writeGeneric (node:net:817:11)
    at Socket._write (node:net:829:8)
    at writeOrBuffer (node:internal/streams/writable:389:12)
    at _write (node:internal/streams/writable:330:10)
    at Socket.Writable.write (node:internal/streams/writable:334:10)
    at IncomingMessage.ondata (node:internal/streams/readable:754:22)
    at IncomingMessage.emit (node:events:527:28)
    at addChunk (node:internal/streams/readable:315:12)
Emitted 'error' event on Socket instance at:
    at Socket.onerror (node:internal/streams/readable:773:14)
    at Socket.emit (node:events:527:28)
    at emitErrorNT (node:internal/streams/destroy:157:8)
    at emitErrorCloseNT (node:internal/streams/destroy:122:3)
    at processTicksAndRejections (node:internal/process/task_queues:83:21) {
  errno: -32,
  code: 'EPIPE',
[...]

Issue #29, opened 2022-07-11
Issues of duplicate packages in current build system. 


kcubeterm:
**Issues 1**: Two packages with same name can't be exist. Suppose we want to add zlib in TUR. Since zlib exist in termux packages already, Buildorder.py will report duplicate package and exit. 

**Workaround 1**: Name package after version. Like zlib1 or zlib1.3 

**Workaround 2**: add ability in setup_environment that could remove those packages from Termux packages directory which has same name in tur dir.

**Issue 2**: Suppose we added a package Holla_worldo and after sometimes Termux also added same package then again buildorder.py reports issues, Here workaround 2 could save us.

licy183:
I think we could rename or remove these packages from TUR when termux main has already have those packages. Having duplicated names between apt repos doen't seem a good solution.

kcubeterm:
Renaming package name seems nice idea. Suppose the packages added in termux repo is older version or newer. But tur needs that specific version of package(others package depends on it)

Issue #28, opened 2022-07-11
[Discussion]: Add a GPG key


licy183:
Currently, our repo is using @kcubeterm 's GPG key to sign the `Release` file. Rather than using our own GPG keys, I'd prefer adding a new GPG key that is used to sign the TUR repo only. We are going the sign the repo using Github Action, but I don't think it is safe for us to store our own GPG passphrase to the repo's Action Secret.

/CC @kcubeterm @2096779623 @T-Dynamos

2096779623:
Add a dedicated GPG public key like in [termux-packages](https://github.com/termux/termux-packages/blob/master/packages/termux-keyring/termux-autobuilds.gpg).

kcubeterm:
We are going to sign repo in github action. And it is safe if github secret is safe enough. I just have to create a workflow and all done


Is it nice idea to add our key in termux-keyring for an unofficial apt repo ?

kcubeterm:
> No, We can create a package(e.g tur-keyring). 

Its seems fine. Let's be lazy and add it on termux-keyring. Aren't we a trusted guy :)

2096779623:
> > No, We can create a package(e.g tur-keyring).
> 
> Its seems fine. Let's be lazy and add it on termux-keyring. Aren't we a trusted guy :)

Well its-pointless was added long ago :)

kcubeterm:
No, idea, why he didn't do. But let's try. What's other member think about it

licy183:
> We are going to sign repo in github action. And it is safe if github secret is safe enough.

Actually, I don't think Github Action Secret is safe enough to store our own GPG key's passphrase. So I think we need to generate another GPG key for the usage of TUR Github Action only. And adding the generated GPG key of TUR Github Action to `termux-keyring` is not a good idea. `termux-keyring` should only contain the GPG keys of termux official repositories.

> No, We can create a package(e.g `tur-keyring`).

I'd prefer this solution. We could create packages named `tur-keyring` and `tur-repo` in `tur` repository, and provide a script to help users download these packages and install them.

kcubeterm:
Nice!!

kcubeterm:
Just create one repo, tur-repo 
That will setup all the requirements when someone install it via apt.

licy183:
TUR uses the same building scripts of current termux packages, so the packages in TUR will not support Android 5/6.

kcubeterm:
added a gpg key at organisations level in github secrets.

[...]

PR #1668, opened 2025-04-12
[Draft] Newpkg: Flightgear

IntinteDAO:
Well, it works using ./scripts/run-docker.sh ./build-package -I flightgear, but no here :(
Maybe more intelligent person can fix it
@robertkirkman


robertkirkman:
If you have to, you can try the tur-on-device folder instead of tur folder, right?

Also, could you test Space bar for jumping in my luanti PR?

- https://github.com/termux/termux-packages/pull/24231

if that also works for you, then maybe we can try to make a similar patch for SuperTux to fix the Space bar (and possibly other keys) in SuperTux when using hardware keyboard (and when using patches to fix the touchscreen buttons in SuperTux, jump works with the touchscreen button without hardware keyboard being necessary)

IntinteDAO:
> If you have to, you can try the tur-on-device folder instead of tur folder, right?
> 

Specifically, what I did was to rewrite CMake a bit so that I didn't have to use TUR-On-Device, but TUR specifically, which on my computer drastically reduced performance.

The problem is in Plib, unfortunately it's a library that should be plowed, but a couple of games use it :(

> Also, could you test Space bar for jumping in my luanti PR?

Your fix works, ONLY I found another interesting case. Well, in the beginning almost no button worked almost. In the sense, WASD didn't work either and I couldn't move the character. I was able to turn on Chat (Y) and all the keys are responsive, they just don't work in the game.

However, if you enable “Prefer scancodes when possible” in the Termux X11 settings under the Keyboard tab, WASD activates. Maybe it's a matter of bad scancode processing or something?

In addition, what I noticed - the keyboard itself looks like it is being captured by the game (Luanti and Supertux), because in Luanti when you run Chat all the keys go through without a problem, and in Supertux when you bind the keys the reaction is correct. Maybe it's a matter of which library?

robertkirkman:
> Specifically, what I did was to rewrite CMake a bit so that I didn't have to use TUR-On-Device, but TUR specifically, which on my computer drastically reduced performance.

I do not completely understand what you mean here, what kind of performance was reduced? The performance of the software at runtime, or the time it takes to build, or something else?

Thank you for testing Luanti,

> However, if you enable “Prefer scancodes when possible” in the Termux X11 settings under the Keyboard tab, WASD activates. Maybe it's a matter of bad scancode processing or something?

this is correct, actually to play Luanti using hardware keyboard at all, **"Prefer scancodes when possible" On is required, there is no other way.** In all my published documentation of the `luanti` package in its PRs and issues, I recommend this setting.

> in Supertux when you bind the keys the reaction is correct.

If you are referring to Spacebar, this is good because, it sounds like very, very similar behavior to the problem of scrambled Spacebar keycodes in Luanti.

If the exact same thing is really happening, then it should be possible to write a patch that works around it in SuperTux by **swapping the keycode definitions of the Space key and the 9-Key** within the source code of SuperTux.


If we are going to try to time the release of the SuperTux package roughly for the release of SuperTux 0.7.0 upstream in order to get the official, merged and final version of their improved touchscreen buttons and CMake files (similarly to how I timed the initial publishing of my Luanti package roughly for the release of Luanti 5.10 upstream in order to get a similar set of upstream improvements), then we have some time before then still to do this, because their tracker has 0.7.0 marked at 80% completed.

https://github.com/SuperTux/supertux/milestone/10

IntinteDAO:
> I do not completely understand what you mean here, what kind of performance was reduced? The performance of the software at runtime, or the time it takes to build, or something else?

It's all about the speed of building the application locally. When creating a build script on x86 you prefer if the application can cross compile, as it doesn't require wasting a lot of time on QEMU emulation using TUR-On-Device.

I'll be thinking about how to get around this in that case so that Plib builds and everything works.

> this is correct, actually to play Luanti using hardware keyboard at all, "Prefer scancodes when possible" On is required, there is no other way. In all my published documentation of the luanti package in its PRs and issues, I recommend this setting.

It makes me wonder if it could be that there is an unimplemented feature in Termux X11 and then couldn't twaik check and correct it based on that? I don't know how realistically it works, but if such a bug is present in hundreds of programs, then you will have to patch each one.

It is strange to me that the keys are detected but do not work. Maybe it's an issue with some function - I have no idea.

> If we are going to try to time the release of the SuperTux package roughly for the release of SuperTux 0.7.0 upstream in order to get the official, merged and final version of their improved touchscreen buttons and CMake files (similarly to how I timed the initial publishing of my Luanti package roughly for the release of Luanti 5.10 upstream in order to get a similar set of upstream improvements), then we have some time before then still to do this, because their tracker has 0.7.0 marked at 80% completed.

We can wait. It also depends how much, but I think probably a couple of months maximum.

robertkirkman:
> It's all about the speed of building the application locally. When creating a build script on x86 you prefer if the application can cross compile, as it doesn't require wasting a lot of time on QEMU emulation using TUR-On-Device.

tur-on-device no longer uses QEMU for CI. Currently, it is now fully using **GitHub Actions official ARM runner**:

https://github.com/termux-user-repository/tur/blob/d7792b3d84647bc14d8c6ce051c09900271d7757/.github/workflows/package_on_device.yml#L228

however, I do understand what you mean because, for example, a package can more easily be migrated from TUR to x11-repo in github.com/termux/termux-packages if it is fully compatible with cross-compilation. I am going to try to fix your error anyway.

> It makes me wonder if it could be that there is an unimplemented feature in Termux X11 and then couldn't twaik check and correct it based on that? I don't know how realistically it works, but if such a bug is present in hundreds of programs, then you will have to patch each one.

so far, it seems to only be present (the scrambled keycode of the Space key) in two known programs, Luanti and SuperTux, but I do not know what the true root cause is. It is possible that other games might have a similar problem too, but I don't know of any yet. It is reasonable to suspect that there is a library shared between both games which contains the root cause of the scrambled keycodes bug, but I haven't identified what library that could be yet.

robertkirkman:
@IntinteDAO

> `configure: error: could not find working GL library`

For your error, could you try this `build.sh` and check whether this is working as desired? Everything here seems to build successfully for me when I do this.

```bash
TERMUX_PKG_HOMEPAGE=https://plib.sourceforge.net
TERMUX_PKG_DESCRIPTION="Provides a Joystick interface, a simple GUI built on top of OpenGL, some standard geometry functions, a sound library and a simple scene graph API built on top of OpenGL."
TERMUX_PKG_LICENSE="BSD"
TERMUX_PKG_VERSION=1.8.5
TERMUX_PKG_SRCURL=http://deb.debian.org/debian/pool/main/p/plib/plib_${TERMUX_PKG_VERSION}.orig.tar.gz
TERMUX_PKG_SHA256=485b22bf6fdc0da067e34ead5e26f002b76326f6371e2ae006415dea6a380a32
TERMUX_PKG_DEPENDS="opengl"
TERMUX_PKG_BUILD_IN_SRC=true
TERMUX_PKG_EXTRA_CONFIGURE_ARGS="
ac_cv_lib_MesaGL_glNewList=yes
"

termux_step_pre_configure() {
	export LDFLAGS+=" -Wl,--no-as-needed,-lOpenSLES,--as-needed"
	./autogen.sh
	touch ltmain.sh
}
```

robertkirkman:
And the reason why I suggest changing the dependency from `mesa` to `opengl`, is because, there 49 total existing packages that depend on `opengl`, but there are only 15 total existing packages that depend directly on `mesa`, so it is more common for OpenGL software to be marked as depending on `opengl`.

robertkirkman:
> `/home/builder/.termux-build/simgear/src/simgear/structure/SGAtomic.cxx:47:5: error: ISO C++17 does not allow 'register' storage class specifier [-Wregister]`

the code that is producing this error seems to me to be very, very old legacy code. It comes from this commit:

https://github.com/FlightGear/simgear/commit/83772c87ac3f4dc8d18a8a4e9a28948ba8b2037c

What seems like a possible best solution to me is simply to apply this patch,

```diff
--- /dev/null
+++ b/tur/simgear/simgear-structure-SGAtomic.hxx-i686-sgatomic-use-gcc4-builtins.patch
@@ -0,0 +1,12 @@
+--- a/simgear/structure/SGAtomic.hxx
++++ b/simgear/structure/SGAtomic.hxx
+@@ -21,8 +21,7 @@
+ #ifndef SGAtomic_HXX
+ #define SGAtomic_HXX
+ 
+-#if defined(__GNUC__) && ((4 < __GNUC__)||(4 == __GNUC__ && 1 <= __GNUC_MINOR__)) && \
+-    defined(__x86_64__)
++#if defined(__GNUC__) && ((4 < __GNUC__)||(4 == __GNUC__ && 1 <= __GNUC_MINOR__))
+ // No need to include something. Is a Compiler API ...
+ # define SGATOMIC_USE_GCC4_BUILTINS
+ #elif defined(__GNUC__) && defined(__i386__)
```

this makes the compilation error go away for me.

Do you happen to have any 32-bit Android-x86 device to try testing the build that results from this at runtime on? If not, then I could try to test it for you.

IntinteDAO:
> Do you happen to have any 32-bit Android-x86 device to try testing the build that results from this at runtime on? If not, then I could try to test it for you.

If you can :) I just have only my phone and laptop that will be obsolete, because I want to use phone as a computer only

robertkirkman:
It might take a while because, I thought it is a good idea to check on a bare metal device for real for a patch like this just in case, to make sure it is actually going to work, and I was going to use a [Pentium 4 Northwood](https://en.wikipedia.org/wiki/Pentium_4#Northwood), but I found out that it actually can't because it has a lot of illegal instruction errors and kernel panic when booting any Android-x86 ROM, and it's because Pentium 4 Northwood really can't run Android 7 because 32-bit x86 Android 7 requires SSE3 instructions at the absolute minimum and Pentium 4 Northwood does not have SSE3. I actually did not know about that before, I know about some other CPU requirements of Android but I did not know Pentium 4 Northwood cannot run Android 7.
So instead, one option I have is to try to use my Intel Atom Z510. Intel Atom Z510 does support SSE3. However, Intel Atom Z510 is **extremely slow**, it might sound surprising but, it is actually much slower than a Pentium 4 Northwood, so if I use it then it will take much longer to set up and do anything.

IntinteDAO:
> It might take a while because, I thought it is a good idea to check on a bare metal device for real for a patch like this just in case, to make sure it is actually going to work, and I was going to use a [Pentium 4 Northwood](https://en.wikipedia.org/wiki/Pentium_4#Northwood), but I found out that it actually can't because it has a lot of illegal instruction errors and kernel panic when booting any Android-x86 ROM, and it's because Pentium 4 Northwood really can't run Android 7 because 32-bit x86 Android 7 requires SSE3 instructions at the absolute minimum and Pentium 4 Northwood does not have SSE3. I actually did not know about that before, I know about some other CPU requirements of Android but I did not know Pentium 4 Northwood cannot run Android 7.
> 
> So instead, one option I have is to try to use my Intel Atom Z510. Intel Atom Z510 does support SSE3. However, Intel Atom Z510 is **extremely slow**, it might sound surprising but, it is actually much slower than a Pentium 4 Northwood, so if I use it then it will take much longer to set up and do anything.

Sure :) In general I also just don't believe Flightgear works well on this hardware (32 bit). It's a pretty resource-intensive simulator and on a Snapdragon 865 I'm getting 7-8 frames on a Cessna 172p on minimum settings, but that could be (and probably is) due to Zink and Turnip. It is possible that it would run a little better on KGSL.

robertkirkman:
> BTW. can you set up automatic package updates? Because I don't know how exactly it works

I do not fully understand what you mean, do you mean automatic package updates from `pkg upgrade`? or, do you mean something that is equivalent to [UnattendedUpgrades feature of Debian-based distributions](https://wiki.debian.org/UnattendedUpgrades)?

IntinteDAO:
> > BTW. can you set up automatic package updates? Because I don't know how exactly it works
> 
> I do not fully understand what you mean, do you mean automatic package updates from `pkg upgrade`? or, do you mean something that is equivalent to [UnattendedUpgrades feature of Debian-based distributions](https://wiki.debian.org/UnattendedUpgrades)?

What I mean is that the TERMUX_PKG_AUTO_UPDATE function should work if it doesn't. I don't know how to set it

robertkirkman:
ohh ok, well, actually I don't know how to make that either, but maybe you can set `TERMUX_PKG_AUTO_UPDATE=true` and then maybe the `termux_gitlab_auto_update()` function will automatically work by itself if you do that. but if it doesn't then you can wait until it makes an error, and then try to fix it by implementing a custom auto update function.

robertkirkman:
Ok, I have started testing on all architectures, and it turns out, currently when I run `fgfs` on **64-bit Android-x86 or 32-bit Android-x86 both**, the initial window appears normally at first, but then clicking the "Download" button freezes the program and in terminal this appears endlessly:

```
   67.94 [ALRT]:general    Floating point interrupt (SIGFPE)
   67.94 [ALRT]:general    Floating point interrupt (SIGFPE)
   67.94 [ALRT]:general    Floating point interrupt (SIGFPE)
   67.94 [ALRT]:general    Floating point interrupt (SIGFPE)
   67.94 [ALRT]:general    Floating point interrupt (SIGFPE)
   67.94 [ALRT]:general    Floating point interrupt (SIGFPE)
   67.94 [ALRT]:general    Floating point interrupt (SIGFPE)
   67.94 [ALRT]:general    Floating point interrupt (SIGFPE)
   67.94 [ALRT]:general    Floating point interrupt (SIGFPE)
   67.94 [ALRT]:general    Floating point interrupt (SIGFPE)
   67.94 [ALRT]:general    Floating point interrupt (SIGFPE)
   67.94 [ALRT]:general    Floating point interrupt (SIGFPE)
   67.94 [ALRT]:general    Floating point interrupt (SIGFPE)
```

however, transferring a copy of `~/.fgfs` folder from an ARM device that successfully downloaded the assets allows `fgfs` to successfully launch on 32-bit Android-x86 and 64-bit Android-x86.

I do not know what causes that error, but I will try to find it.

If you want to try to debug it also, I can explain how to install termux-docker and `tigervnc` into your 64-bit x86 laptop so you can reproduce this.

robertkirkman:
Here is a backtrace of what happens

```
Thread 1 "fgfs" received signal SIGFPE, Arithmetic exception.
0x0191e7f9 in ?? ()
(gdb) bt
#0  0x0191e7f9 in ?? ()
#1  0xb3d768f9 in ?? () from /data/data/com.termux/files/usr/lib/libQt6Core.so.6.9.0
#2  0xb3d758c9 in QMetaObject::activate(QObject*, QMetaObject const*, int, void**) ()
   from /data/data/com.termux/files/usr/lib/libQt6Core.so.6.9.0
#3  0x0182124d in QMetaObject::activate<void, unsigned long long, unsigned long long> (
    sender=0xaf58e280, mo=0x0, local_signal_index=2, ret=0x0, args=<optimized out>, 
    args=<optimized out>) at /data/data/com.termux/files/usr/include/qt6/QtCore/qobjectdefs.h:306
#4  InstallFGDataThread::downloadProgress (this=0xaf58e280, _t1=<optimized out>, 
    _t2=<optimized out>) at src/GUI/fglauncher_autogen/include/SetupRootDialog.moc:160
#5  InstallFGDataThread::onDownloadProgress (this=0xaf58e280, got=0, total=0)
    at /home/builder/.termux-build/flightgear/src/src/GUI/SetupRootDialog.cxx:236
#6  0x0182199f in QtPrivate::FunctorCall<std::__ndk1::integer_sequence<unsigned int, 0u, 1u>, QtPrivate::List<long long, long long>, void, void (InstallFGDataThread::*)(unsigned long long, unsigned long long)>::call(void (InstallFGDataThread::*)(unsigned long long, unsigned long long), InstallFGDataThread*, void**)::{lambda()#1}::operator()() const (this=<optimized out>)
    at /data/data/com.termux/files/usr/include/qt6/QtCore/qobjectdefs_impl.h:127
#7  QtPrivate::FunctorCallBase::call_internal<void, QtPrivate::FunctorCall<std::__ndk1::integer_sequence<unsigned int, 0u, 1u>, QtPrivate::List<long long, long long>, void, void (InstallFGDataThread::*)(unsigned long long, unsigned long long)>::call(void (InstallFGDataThread::*)(unsigned long long, unsigned long long), InstallFGDataThread*, void**)::{lambda()#1}>(void**, QtPrivate::FunctorCall<std::__ndk1::integer_sequence<unsigned int, 0u, 1u>, QtPrivate::List<long long, long long>, void, void (InstallFGDataThread::*)(unsigned long long, unsigned long long)>::call(void (InstallFGDataThread::*)(unsigned long long, unsigned long long), InstallFGDataThread*, void**)::{lambda()#1}&&) (
    args=0xbf82e810, fn=...)
    at /data/data/com.termux/files/usr/include/qt6/QtCore/qobjectdefs_impl.h:65
#8  QtPrivate::FunctorCall<std::__ndk1::integer_sequence<unsigned int, 0u, 1u>, QtPrivate::List<long long, long long>, void, void (InstallFGDataThread::*)(unsigned long long, unsigned long long)>::call (f=<optimized out>, o=<optimized out>, arg=0xbf82e810)
    at /data/data/com.termux/files/usr/include/qt6/QtCore/qobjectdefs_impl.h:126
#9  QtPrivate::FunctionPointer<void (InstallFGDataThread::*)(unsigned long long, unsigned long long)>::call<QtPrivate::List<long long, long long>, void> (f=<optimized out>, o=<optimized out>, 
    arg=0xbf82e810) at /data/data/com.termux/files/usr/include/qt6/QtCore/qobjectdefs_impl.h:174
#10 QtPrivate::QCallableObject<void (InstallFGDataThread::*)(unsigned long long, unsigned long long), QtPrivate::List<long long, long long>, void>::impl (which=1, this_=0xb7877e20, r=0xaf58e280, 
    a=0xbf82e810, ret=0x0)
--Type <RET> for more, q to quit, c to continue without paging--
    at /data/data/com.termux/files/usr/include/qt6/QtCore/qobjectdefs_impl.h:545
#11 0xb3d768f9 in ?? () from /data/data/com.termux/files/usr/lib/libQt6Core.so.6.9.0
#12 0xb3d758c9 in QMetaObject::activate(QObject*, QMetaObject const*, int, void**) ()
   from /data/data/com.termux/files/usr/lib/libQt6Core.so.6.9.0
#13 0xb488a8b7 in ?? () from /data/data/com.termux/files/usr/lib/libQt6Network.so.6.9.0
#14 0xb48973d4 in ?? () from /data/data/com.termux/files/usr/lib/libQt6Network.so.6.9.0
#15 0xb3d69b60 in QMetaCallEvent::placeMetaCall(QObject*) ()
   from /data/data/com.termux/files/usr/lib/libQt6Core.so.6.9.0
#16 0xb3d6c015 in QObject::event(QEvent*) ()
   from /data/data/com.termux/files/usr/lib/libQt6Core.so.6.9.0
#17 0xb1bd7101 in QApplicationPrivate::notify_helper(QObject*, QEvent*) ()
   from /data/data/com.termux/files/usr/lib/libQt6Widgets.so.6.9.0
#18 0xb1bd830a in QApplication::notify(QObject*, QEvent*) ()
   from /data/data/com.termux/files/usr/lib/libQt6Widgets.so.6.9.0
#19 0xb3d0aaa0 in QCoreApplication::sendEvent(QObject*, QEvent*) ()
   from /data/data/com.termux/files/usr/lib/libQt6Core.so.6.9.0
#20 0xb3d0b289 in QCoreApplicationPrivate::sendPostedEvents(QObject*, int, QThreadData*) ()
   from /data/data/com.termux/files/usr/lib/libQt6Core.so.6.9.0
#21 0xb403b75b in ?? () from /data/data/com.termux/files/usr/lib/libQt6Core.so.6.9.0
#22 0xb73c110d in g_main_dispatch (context=<optimized out>) at ../src/glib/gmain.c:3398
#23 g_main_context_dispatch_unlocked (context=<optimized out>, context@entry=0xaf58e080)
    at ../src/glib/gmain.c:4249
#24 0xb73c15f5 in g_main_context_iterate_unlocked (context=0xaf58e080, block=<optimized out>, 
    dispatch=1, self=<optimized out>) at ../src/glib/gmain.c:4314
#25 0xb73c16b9 in g_main_context_iteration (context=0xaf58e080, may_block=1)
    at ../src/glib/gmain.c:4379
#26 0xb403ad61 in QEventDispatcherGlib::processEvents(QFlags<QEventLoop::ProcessEventsFlag>) ()
   from /data/data/com.termux/files/usr/lib/libQt6Core.so.6.9.0
#27 0xafd3dfcd in ?? () from /data/data/com.termux/files/usr/lib/libQt6XcbQpa.so.6.9.0
#28 0xb3d17086 in QEventLoop::exec(QFlags<QEventLoop::ProcessEventsFlag>) ()
   from /data/data/com.termux/files/usr/lib/libQt6Core.so.6.9.0
#29 0xb1ebb7ac in QDialog::exec() ()
   from /data/data/com.termux/files/usr/lib/libQt6Widgets.so.6.9.0
#30 0x0181adb0 in SetupRootDialog::runDialog (prompt=SetupRootDialog::DefaultPathCheckFailed)
    at /home/builder/.termux-build/flightgear/src/src/GUI/SetupRootDialog.cxx:337
--Type <RET> for more, q to quit, c to continue without paging--
#31 SetupRootDialog::runDialog (usingDefaultRoot=<optimized out>)
    at /home/builder/.termux-build/flightgear/src/src/GUI/SetupRootDialog.cxx:327
#32 0x017e6892 in flightgear::showSetupRootDialog (usingDefaultRoot=<optimized out>)
    at /home/builder/.termux-build/flightgear/src/src/GUI/QtLauncher.cxx:797
#33 0x011d1274 in pc_normal () from /data/data/com.termux/files/usr/bin/fgfs
#34 0x011cdab3 in pc_normal () from /data/data/com.termux/files/usr/bin/fgfs
#35 0x0118ed01 in pc_normal () from /data/data/com.termux/files/usr/bin/fgfs
#36 0x011c964c in pc_normal () from /data/data/com.termux/files/usr/bin/fgfs
#37 0x0120a179 in pc_normal () from /data/data/com.termux/files/usr/bin/fgfs
```

And, the reason I know that this is consistently **reproducible only on Android-x86 generally, both 64-bit and 32-bit**, is because I have isolated it to Android-x86 specifically through process of elimination using the following set of test devices. I didn't end up actually being able to set up a device with a 32-bit x86 CPU yet that can run Android 7 or newer bare metal, but I have arrived at this conclusion using the following tests:

- 64-bit x86 GNU/Linux device running termux-docker x86_64 directly in docker: **can reproduce this**
- the same 64-bit x86 GNU/Linux device running a 32-bit x86 GNU/Linux virtual machine (to get a 32-bit guest kernel) running termux-docker i686 in docker inside the virtual machine: **can reproduce this**
- 64-bit x86 device running 64-bit Android-x86 13 bare metal (this means an entire [Zygote](https://source.android.com/docs/core/runtime/zygote) and [SurfaceFlinger](https://source.android.com/docs/core/graphics/surfaceflinger-windowmanager) compiled for 64-bit x86 and booted by GRUB running the entire x86_64 Termux and Termux:X11 APKs directly outside of any virtual machine and outside of Docker): **can reproduce this**
- 64-bit ARM GNU/Linux device running termux-docker aarch64 directly in its 64-bit ARM docker: **cannot reproduce this**
- 64-bit ARM bare metal Android 13 phone device: **cannot reproduce this**

IntinteDAO:
> Here is a backtrace of what happens
> 
> ```
> Thread 1 "fgfs" received signal SIGFPE, Arithmetic exception.
> 0x0191e7f9 in ?? ()
> (gdb) bt
> #0  0x0191e7f9 in ?? ()
> #1  0xb3d768f9 in ?? () from /data/data/com.termux/files/usr/lib/libQt6Core.so.6.9.0
> #2  0xb3d758c9 in QMetaObject::activate(QObject*, QMetaObject const*, int, void**) ()
>    from /data/data/com.termux/files/usr/lib/libQt6Core.so.6.9.0
> #3  0x0182124d in QMetaObject::activate<void, unsigned long long, unsigned long long> (
>     sender=0xaf58e280, mo=0x0, local_signal_index=2, ret=0x0, args=<optimized out>, 
>     args=<optimized out>) at /data/data/com.termux/files/usr/include/qt6/QtCore/qobjectdefs.h:306
> #4  InstallFGDataThread::downloadProgress (this=0xaf58e280, _t1=<optimized out>, 
>     _t2=<optimized out>) at src/GUI/fglauncher_autogen/include/SetupRootDialog.moc:160
> #5  InstallFGDataThread::onDownloadProgress (this=0xaf58e280, got=0, total=0)
>     at /home/builder/.termux-build/flightgear/src/src/GUI/SetupRootDialog.cxx:236
> #6  0x0182199f in QtPrivate::FunctorCall<std::__ndk1::integer_sequence<unsigned int, 0u, 1u>, QtPrivate::List<long long, long long>, void, void (InstallFGDataThread::*)(unsigned long long, unsigned long long)>::call(void (InstallFGDataThread::*)(unsigned long long, unsigned long long), InstallFGDataThread*, void**)::{lambda()#1}::operator()() const (this=<optimized out>)
>     at /data/data/com.termux/files/usr/include/qt6/QtCore/qobjectdefs_impl.h:127
> #7  QtPrivate::FunctorCallBase::call_internal<void, QtPrivate::FunctorCall<std::__ndk1::integer_sequence<unsigned int, 0u, 1u>, QtPrivate::List<long long, long long>, void, void (InstallFGDataThread::*)(unsigned long long, unsigned long long)>::call(void (InstallFGDataThread::*)(unsigned long long, unsigned long long), InstallFGDataThread*, void**)::{lambda()#1}>(void**, QtPrivate::FunctorCall<std::__ndk1::integer_sequence<unsigned int, 0u, 1u>, QtPrivate::List<long long, long long>, void, void (InstallFGDataThread::*)(unsigned long long, unsigned long long)>::call(void (InstallFGDataThread::*)(unsigned long long, unsigned long long), InstallFGDataThread*, void**)::{lambda()#1}&&) (
>     args=0xbf82e810, fn=...)
>     at /data/data/com.termux/files/usr/include/qt6/QtCore/qobjectdefs_impl.h:65
> #8  QtPrivate::FunctorCall<std::__ndk1::integer_sequence<unsigned int, 0u, 1u>, QtPrivate::List<long long, long long>, void, void (InstallFGDataThread::*)(unsigned long long, unsigned long long)>::call (f=<optimized out>, o=<optimized out>, arg=0xbf82e810)
>     at /data/data/com.termux/files/usr/include/qt6/QtCore/qobjectdefs_impl.h:126
> #9  QtPrivate::FunctionPointer<void (InstallFGDataThread::*)(unsigned long long, unsigned long long)>::call<QtPrivate::List<long long, long long>, void> (f=<optimized out>, o=<optimized out>, 
>     arg=0xbf82e810) at /data/data/com.termux/files/usr/include/qt6/QtCore/qobjectdefs_impl.h:174
> #10 QtPrivate::QCallableObject<void (InstallFGDataThread::*)(unsigned long long, unsigned long long), QtPrivate::List<long long, long long>, void>::impl (which=1, this_=0xb7877e20, r=0xaf58e280, 
>     a=0xbf82e810, ret=0x0)
> --Type <RET> for more, q to quit, c to continue without paging--
>     at /data/data/com.termux/files/usr/include/qt6/QtCore/qobjectdefs_impl.h:545
> #11 0xb3d768f9 in ?? () from /data/data/com.termux/files/usr/lib/libQt6Core.so.6.9.0
> #12 0xb3d758c9 in QMetaObject::activate(QObject*, QMetaObject const*, int, void**) ()
>    from /data/data/com.termux/files/usr/lib/libQt6Core.so.6.9.0
> #13 0xb488a8b7 in ?? () from /data/data/com.termux/files/usr/lib/libQt6Network.so.6.9.0
> #14 0xb48973d4 in ?? () from /data/data/com.termux/files/usr/lib/libQt6Network.so.6.9.0
> #15 0xb3d69b60 in QMetaCallEvent::placeMetaCall(QObject*) ()
>    from /data/data/com.termux/files/usr/lib/libQt6Core.so.6.9.0
> #16 0xb3d6c015 in QObject::event(QEvent*) ()
>    from /data/data/com.termux/files/usr/lib/libQt6Core.so.6.9.0
> #17 0xb1bd7101 in QApplicationPrivate::notify_helper(QObject*, QEvent*) ()
>    from /data/data/com.termux/files/usr/lib/libQt6Widgets.so.6.9.0
> #18 0xb1bd830a in QApplication::notify(QObject*, QEvent*) ()
>    from /data/data/com.termux/files/usr/lib/libQt6Widgets.so.6.9.0
> #19 0xb3d0aaa0 in QCoreApplication::sendEvent(QObject*, QEvent*) ()
>    from /data/data/com.termux/files/usr/lib/libQt6Core.so.6.9.0
> #20 0xb3d0b289 in QCoreApplicationPrivate::sendPostedEvents(QObject*, int, QThreadData*) ()
>    from /data/data/com.termux/files/usr/lib/libQt6Core.so.6.9.0
> #21 0xb403b75b in ?? () from /data/data/com.termux/files/usr/lib/libQt6Core.so.6.9.0
> #22 0xb73c110d in g_main_dispatch (context=<optimized out>) at ../src/glib/gmain.c:3398
> #23 g_main_context_dispatch_unlocked (context=<optimized out>, context@entry=0xaf58e080)
>     at ../src/glib/gmain.c:4249
> #24 0xb73c15f5 in g_main_context_iterate_unlocked (context=0xaf58e080, block=<optimized out>, 
>     dispatch=1, self=<optimized out>) at ../src/glib/gmain.c:4314
> #25 0xb73c16b9 in g_main_context_iteration (context=0xaf58e080, may_block=1)
>     at ../src/glib/gmain.c:4379
> #26 0xb403ad61 in QEventDispatcherGlib::processEvents(QFlags<QEventLoop::ProcessEventsFlag>) ()
>    from /data/data/com.termux/files/usr/lib/libQt6Core.so.6.9.0
> #27 0xafd3dfcd in ?? () from /data/data/com.termux/files/usr/lib/libQt6XcbQpa.so.6.9.0
> #28 0xb3d17086 in QEventLoop::exec(QFlags<QEventLoop::ProcessEventsFlag>) ()
>    from /data/data/com.termux/files/usr/lib/libQt6Core.so.6.9.0
> #29 0xb1ebb7ac in QDialog::exec() ()
>    from /data/data/com.termux/files/usr/lib/libQt6Widgets.so.6.9.0
> #30 0x0181adb0 in SetupRootDialog::runDialog (prompt=SetupRootDialog::DefaultPathCheckFailed)
>     at /home/builder/.termux-build/flightgear/src/src/GUI/SetupRootDialog.cxx:337
> --Type <RET> for more, q to quit, c to continue without paging--
> #31 SetupRootDialog::runDialog (usingDefaultRoot=<optimized out>)
>     at /home/builder/.termux-build/flightgear/src/src/GUI/SetupRootDialog.cxx:327
> #32 0x017e6892 in flightgear::showSetupRootDialog (usingDefaultRoot=<optimized out>)
>     at /home/builder/.termux-build/flightgear/src/src/GUI/QtLauncher.cxx:797
> #33 0x011d1274 in pc_normal () from /data/data/com.termux/files/usr/bin/fgfs
> #34 0x011cdab3 in pc_normal () from /data/data/com.termux/files/usr/bin/fgfs
> #35 0x0118ed01 in pc_normal () from /data/data/com.termux/files/usr/bin/fgfs
> #36 0x011c964c in pc_normal () from /data/data/com.termux/files/usr/bin/fgfs
> #37 0x0120a179 in pc_normal () from /data/data/com.termux/files/usr/bin/fgfs
> ```
> 
> And, the reason I know that this is consistently **reproducible only on Android-x86 generally, both 64-bit and 32-bit**, is because I have isolated it to Android-x86 specifically through process of elimination using the following set of test devices. I didn't end up actually being able to set up a device with a 32-bit x86 CPU yet that can run Android 7 or newer bare metal, but I have arrived at this conclusion using the following tests:
> 
>     * 64-bit x86 GNU/Linux device running termux-docker x86_64 directly in docker: **can reproduce this**
> 
>     * the same 64-bit x86 GNU/Linux device running a 32-bit x86 GNU/Linux virtual machine (to get a 32-bit guest kernel) running termux-docker i686 in docker inside the virtual machine: **can reproduce this** (but also, just like the 64-bit Android-x86 devices, can run the flight simulator successfully if the `~/.fgfs` folder is copied to it from an ARM device)
> 
>     * 64-bit x86 device running 64-bit Android-x86 13 bare metal (this means an entire [Zygote](https://source.android.com/docs/core/runtime/zygote) and [SurfaceFlinger](https://source.android.com/docs/core/graphics/surfaceflinger-windowmanager) compiled for 64-bit x86 and booted by GRUB running the entire x86_64 Termux and Termux:X11 APKs directly outside of any virtual machine and outside of Docker): **can reproduce this**
> 
>     * 64-bit ARM GNU/Linux device running termux-docker aarch64 directly in its 64-bit ARM docker: **cannot reproduce this**
> 
>     * 64-bit ARM bare metal Android 13 phone device: **cannot reproduce this**

I don't know, unfortunately. It looks like a problem with Qt6, however I don't know if it's problems with Qt6 itself or the Launcher code.

You could always try to build the Qt5 Launcher or not add the launcher at all, but in my opinion it should be there (and cool if on a newer version rather than an older one).

If you have any ideas it's cool if you can implement them and fix the problem, if not I would try other people from the Termux world to ask them to do so

robertkirkman:
This is TUR so I think (as far as I know) it is ok and fine to have the package built for x86 here even if it has a problem.

If there is no solution yet, then maybe it is a good idea to add a postinst script with a condition into only the i686 and x86_64 builds of the package that prints "there is a problem with the Download button of the launcher, you will need to manually install the assets before launching" when users install it.

You can see an example of how to make a postinst script with a condition here,

https://github.com/termux/termux-packages/blob/9286dcba793835c6c0ee01bd065b31d1b68a3cd5/packages/glibc-repo/build.sh#L21

basically you can put that, but instead of `[ "$TERMUX_PACKAGE_FORMAT" = "pacman" ] && return 0` make it `[ "$TERMUX_ARCH" = "arm" ] && return 0` and `[ "$TERMUX_ARCH" = "aarch64" ] && return 0`, then put `echo "#!$TERMUX_PREFIX/bin/sh" > postinst` and `echo "echo ..." >> postinst` with the text you want

robertkirkman:
Also, I think that there's a way to have the assets pre-available through `pkg install` as a `.deb` package, like some other games, and make it a dependency of this package.

Look at what Debian's package does, isn't that the needed assets? `flightgear-data`

https://salsa.debian.org/debian/flightgear-data/-/blob/b583c936e6c5c27c255c091c08fc83cd3c6e8017/debian/watch 


https://packages.debian.org/bookworm/flightgear-data-base


It's 1 gigabyte though. I don't know if that is an acceptable size in TUR repo. the biggest one in TUR currently is probably `supertuxkart-data` which I think is 500 MB.

IntinteDAO:
> Also, I think that there's a way to have the assets pre-available through `pkg install` as a `.deb` package, like some other games, and make it a dependency of this package.
> 
> Look at what Debian's package does, isn't that the needed assets? `flightgear-data`
> 
> https://salsa.debian.org/debian/flightgear-data/-/blob/b583c936e6c5c27c255c091c08fc83cd3c6e8017/debian/watch
> 
> https://packages.debian.org/bookworm/flightgear-data-base
> 
> It's 1 gigabyte though. I don't know if that is an acceptable size in TUR repo. the biggest one in TUR currently is probably `supertuxkart-data` which I think is 500 MB.

Flightgear-data is the same collection that downloads Launcher. The difference is that Flightgear launcher can only download and install simulator data from the latest version (which is not yet available in Debian). Previously, it was up to the user to provide files to the simulator.

We can supply them ourselves as a package, but it's a question of whether we want to, if the launcher can potentially do it.

IntinteDAO:
> @IntinteDAO Sorry for the late reply.
> 
> S24+(Exynos 2400) test results It is freezing and not working on that screen. Screenshot_2025-04-15_21-28-37

No problem, try

> fgfs --offset-distance=4000 --offset-azimuth=90 --altitude=500 --heading=0 --model-hz=60 --bpp=16 --disable-random-objects --prop:/sim/rendering/quality-level=0 --prop:/sim/rendering/shaders/quality-level=0 --disable-ai-traffic --prop:/sim/ai/enabled=0 --prop:/sim/rendering/random-vegetation=0 --prop:/sim/rendering/random-buildings=0 --disable-specular-highlight --disable-ai-models --disable-clouds --disable-clouds3d --fog-fastest --visibility=5000 --disable-distance-attenuation --disable-real-weather-fetch

I want it as a benchmark to see if cool this Exynos.

hansm629:
@IntinteDAO 
Oh! It works!
But I don't know how to control it.
Is there a way to measure the frame?
Screenshot_2025-04-16_00-30-49

IntinteDAO:
> @IntinteDAO Oh! It works! But I don't know how to control it. Is there a way to measure the frame? Screenshot_2025-04-16_00-30-49

 Through the in-sim menubar: View > Display Options, tick the "Show frame rate" box 
 
 How to start:
- Cessna C172P -> Autostart, wait until the engine wlll works
Page up to increase throttle

If you get 60 knots, you can type 3-4 times "down" button to climb. 

___

However you are in water, so you can't start. You need to set in launcher Download scenery and set DNS for Google in settings and FG will download a scenery

hansm629:
@IntinteDAO 
Autostart doesn't seem to be working properly. I'm getting about 16-17fps.
Screenshot_2025-04-19_11-11-11
Screenshot_2025-04-19_11-11-49

IntinteDAO:
> @IntinteDAO Autostart doesn't seem to be working properly. I'm getting about 16-17fps.

Ok, but it's not a Termux bug, just it's hard to start a airplane :P But 17 FPS is playable IMO

IntinteDAO:
@robertkirkman 
Bump to new version with some bugs fixed. Maybe it will work for some devices

But I have a glitches on Adreno 735 (but it's a driver issue). Maybe you can fix them somehow :P

tur/flightgear/build.sh
```diff
@@ -0,0 +1,21 @@
+TERMUX_PKG_HOMEPAGE=https://www.flightgear.org
+TERMUX_PKG_DESCRIPTION="Free open-source flight simulator"
+TERMUX_PKG_LICENSE="GPL-2.0"
+TERMUX_PKG_MAINTAINER="@termux-user-repository"
+TERMUX_PKG_VERSION=2024.1.3
+TERMUX_PKG_SRCURL=https://gitlab.com/flightgear/flightgear/-/archive/${TERMUX_PKG_VERSION}/flightgear-${TERMUX_PKG_VERSION}.tar.bz2?ref_type=tags
```
robertkirkman: I think this does not need `?ref_type=tags` on the end?

tur/plib/00_fix_kfreebsd_FTBFS.patch
```diff
@@ -0,0 +1,49 @@
...
+ #include <string.h>
+ #include <errno.h>
+ #include <sys/ioctl.h>
+-#if defined(__FreeBSD__)
++#if defined(__FreeBSD__) || defined(__FreeBSD_kernel__)
+ # include <sys/joystick.h>
+ #else
+ # include <machine/joystick.h>		// For analog joysticks
+Index: plib/src/sl/slDSP.cxx
+===================================================================
```
robertkirkman: I don't think this needs to really have GNU/kFreeBSD or GNU/Hurd patches in it, I understand they are from Debian, but would it be ok for you to trim it down to only the patches which do something in Termux?

tur/plib/build.sh
```diff
@@ -0,0 +1,17 @@
+TERMUX_PKG_HOMEPAGE=https://plib.sourceforge.net
+TERMUX_PKG_DESCRIPTION="Provides a Joystick interface, a simple GUI built on top of OpenGL, some standard geometry functions, a sound library and a simple scene graph API built on top of OpenGL."
+TERMUX_PKG_LICENSE="BSD"
+TERMUX_PKG_MAINTAINER="@termux-user-repository"
+TERMUX_PKG_VERSION=1.8.5
+TERMUX_PKG_SRCURL=http://deb.debian.org/debian/pool/main/p/plib/plib_${TERMUX_PKG_VERSION}.orig.tar.gz
```
robertkirkman: Can this source URL work instead?

```bash
http://plib.sourceforge.net/dist/plib-${TERMUX_PKG_VERSION}.tar.gz
```

tur/simgear/build.sh
```diff
@@ -0,0 +1,18 @@
+TERMUX_PKG_HOMEPAGE=https://www.flightgear.org/
+TERMUX_PKG_DESCRIPTION="SimGear - simulation libraries for FlightGear"
+TERMUX_PKG_LICENSE="GPL-2.0"
+TERMUX_PKG_MAINTAINER="@termux"
+TERMUX_PKG_VERSION="2024.1.3"
+TERMUX_PKG_SRCURL=http://gitlab.com/flightgear/simgear/-/archive/${TERMUX_PKG_VERSION}/simgear-${TERMUX_PKG_VERSION}.tar.bz2?ref_type=tags
+TERMUX_PKG_SHA256=d395dcca2aa3875c5468f285f23d053e41e5e8d92576e16fd9801d685441115d
+TERMUX_PKG_DEPENDS="boost, mesa, openal-soft, openscenegraph, libcurl, c-ares"
+TERMUX_PKG_BUILD_DEPENDS="libglvnd-dev, mesa-dev"
+TERMUX_PKG_EXTRA_CONFIGURE_ARGS=''
```
robertkirkman: This line can be removed because `TERMUX_PKG_EXTRA_CONFIGURE_ARGS` is empty by default

tur/simgear/build.sh
```diff
@@ -0,0 +1,18 @@
...
+TERMUX_PKG_CPPFLAGS="-DXML_POOR_ENTROPY -D_LIBCPP_HAS_NO_COND_CLOCKWAIT"
```
robertkirkman: Could you make these variables applied inside the `termux_step_pre_configure()` function instead of global variable? it would be more conventional and less mixed up with other `TERMUX_PKG` variables. Like this:

```bash
export CPPFLAGS+=" -DXML_POOR_ENTROPY -D_LIBCPP_HAS_NO_COND_CLOCKWAIT"
export CFLAGS+=" $CPPFLAGS"
export CXXFLAGS+=" $CPPFLAGS"
```

tur/flightgear/build.sh
robertkirkman: It still has `?ref_type=tags` on my screen, did you try it without?

IntinteDAO:
> It still has `?ref_type=tags` on my screen, did you try it without?

I was sure I had deleted it correctly. Too many windows, copies of repositories, etc. :D

robertkirkman:
Except for the problem with the auto downloader of the assets on Android-x86, well I don't know what causes that, but anyway it works on ARM devices which is most termux devices and if someone wants it on x86 they can, I think, find this thread and know that it can work if they manually transfer the assets from somewhere else

IntinteDAO:
> Except for the problem with the auto downloader of the assets on Android-x86, well I don't know what causes that, but anyway it works on ARM devices which is most termux devices and if someone wants it on x86 they can, I think, find this thread and know that it can work if they manually transfer the assets from somewhere else

Maybe it's DNS releated issue? 

Overall, I see a small bug to fix + 2 improvements for another game.

I'm wondering whether to move it to Termux Packages. One patch (fake-fgrcc) would need to be rewritten.

When building Flightgear, it executes the fgrcc command, which creates a file that serves no purpose. So I worked around the problem, but you can just make it so that the binary actually runs somehow and creates that file.

tur/flightgear/fake_fgrcc.patch
```diff
@@ -0,0 +1,130 @@
...
++      OUTPUT
++        ${CMAKE_CURRENT_BINARY_DIR}/FlightGear-resources.cxx
++        ${CMAKE_CURRENT_BINARY_DIR}/FlightGear-resources.hxx
++      MAIN_DEPENDENCY
++        fgrcc ${CMAKE_CURRENT_SOURCE_DIR}/FlightGear-resources.xml
++      COMMENT
++        "Re-generating embedded resources"
++  )
++else()
++  # Dummy command dla cross-kompilacji: nic nie rób, ale zwróć sukces
```
robertkirkman, line 56: Yes hopefully the comments can be changed to English

IntinteDAO: I hope I will remove this patch and use step_host_build :)

IntinteDAO: I have another question about fgrcc.

In the case of xmoto, it was quite logical to use the Linux version to build the game files.

However, what about fgrcc, which is built but never makes it into the “final” package?

If you know how to do it, write it down in general terms, and I'll look for a solution myself if it's reasonably simple.

robertkirkman: > I have another question about fgrcc.
> 
> In the case of xmoto, it was quite logical to use the Linux version to build the game files.
> 
> However, what about fgrcc, which is built but never makes it into the “final” package?
> 
> If you know how to do it, write it down in general terms, and I'll look for a solution myself if it's reasonably simple.

Yes, that sounds like a good idea, but I do not really know about what fgrcc is yet or now to use it, so I will need to research it and do some tests to learn about it.

I will let you know in the future if I'm able to create a better solution than this or not.

PR #1716, opened 2025-05-06
addpkg(tur/erlang26-manpages): 26.2.5.11

robertkirkman:
- Fixes the rest of https://github.com/termux/termux-packages/issues/24624 by implementing the `erl -man io` command
  - Depends on https://github.com/termux/termux-packages/pull/24627 so that `erl -man` is patched to search the desired paths for man pages

- In all Erlang versions after https://github.com/erlang/otp/pull/8026, the `man3`, `man4`, `man6` and `man7` manpages no longer exist in upstream releases (Erlang 27+), so these manpages can no longer be up to date with the current Erlang releases.

- Can potentially conflict with some other packages - conflicts with `perl` and `zlib` found and resolved here

### DIFF
--- a/tur/erlang26-manpages/build.sh
+++ b/tur/erlang26-manpages/build.sh
@@ -1,0 +1,21 @@
+TERMUX_PKG_HOMEPAGE=https://www.erlang.org/
+TERMUX_PKG_DESCRIPTION="Prebuilt legacy manpages for Erlang"
+TERMUX_PKG_LICENSE="Apache-2.0"
+TERMUX_PKG_MAINTAINER="@termux-user-repository"
+TERMUX_PKG_VERSION="26.2.5.11"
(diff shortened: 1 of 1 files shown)


robertkirkman:
This package has similar contents and purpose to the `erlang-manpages` package of Debian, and in Debian the phenomenon of the manpages being dropped by upstream can be observed in that the package existed in Debian 10, 11 and 12, but will not exist in Debian 13 because Debian 13 has Erlang 27. https://packages.debian.org/bookworm/erlang-manpages

But there is a user who needs to use the `erl -man io` command so this might be able to serve as a workaround for them.

robertkirkman:
Since OP has not responded about their problem,

and this is essentially a replica of a soon-to-be-obsolete Debian package that is not going to be maintained either by Erlang upstream or by other distros anymore,

I am going to close this, and,

@lost22git it can be opened again some time in the future if you end up still needing this, let me know and we can discuss it.

PR #1723, opened 2025-05-10
chore(tur/apkeditor): runnable latest jar

Veha0001:
# Summary
- Select APKEditor-*.jar, which can be downloaded from the latest build, with the highest version in `/libexec/apkeditor`. Which can be linked anywhere btw.

- Remove an unnecessary Java argument.


licy183:
Thanks!

PR #1123, opened 2024-07-18
addpkg(tur/leanify): 0.4.3+20231217

TomJo2000:
Requested in: 
- termux/termux-packages#20885

> [!NOTE]
> Checklist:
> - [x] Builds locally
> - [ ] Builds on CI
> - [x] Tested on device (Android 13, AArch64, Fairphone 5)

<h1><em></em></h1> 

@TPS, if you wouldn't mind I'd appreciate if you could check the package for correct functioning.
I just did a basic "does it print its help and version information" test.
Instructions below.

<sup>(This is a pre-written, saved reply.)</sup>
If you want to test this PR please download the appropriate DEB package(s)
from the build artifacts of the [associated PR's latest CI run](https://github.com/termux-user-repository/tur/actions/runs/9987167934).
Screenshot_20240619_232413

After downloading the build artifact, make sure to `unzip` and un-`tar` it.
<details><summary>Detailed instructions, if needed.</summary>
<p>

```bash
# finding out what architecture you need
# architecture is just below the TERMUX_VERSION
termux-info

# e.g.
# [...]
# TERMUX_MAIN_PACKAGE_FORMAT=debian
# TERMUX_VERSION=0.118.0
# TERMUX__USER_ID=0
# Packages CPU architecture:
# aarch64
# [...]

# =======================

# make sure `unzip` and `tar` are installed using
pkg install unzip tar

# unzip the artifact (if you have a different architecture this might be arm, i686 or x86_64 instead)
unzip debs-aarch64-*.zip

# untar the artifact
tar xf debs-aarch64-*.tar

# You should now have a debs/ directory in your current working directory
# Install the packages from the local source using
pkg install -- ./debs/*.deb

# to clean up, you can remove the debs/ directory, .tar file and .zip file
rm -rfi debs debs-aarch64-*.zip debs-aarch64-*.tar
```

</p>
</details>

PR #1786, opened 2025-06-01
bump(tur/leanify): 0.4.3+20250520

TomJo2000:
- addresses https://github.com/termux-user-repository/tur/pull/1123#issuecomment-2927007577

Also cleaned up the build script and added proper SHA256 sum verification,
based on the approach used in [`go-findimagedupes`](https://github.com/termux/termux-packages/blob/master/packages/go-findimagedupes/build.sh).


TPS:
@TomJo2000 Thanks very much! 🙇🏾‍♂️

Issue #1788, opened 2025-06-02
[Bug]: `libllvm-16` conflicts with `llvm-tools-16`


johndebia2022:
### Problem description

`ALSULTAN_SUBPKG_INCLUDE` must use `_INSTALL_PREFIX_R` instead of `_INSTALL_PREFIX`

### What steps will reproduce the bug?

Build

### System information

termux-info:

```

```

samigamal2025:
this shoud fix it
https://github.com/termux-user-repository/tur/pull/1800

PR #1800, opened 2025-06-05
fix conflict between libllvm-16 and llvm-tools-16

samigamal2025:
Fixes: https://github.com/termux-user-repository/tur/issues/1788


licy183:
Please bump revision of the package.

PR #1811, opened 2025-06-10
Newpkg: Lugaru

IntinteDAO:
# Known issues:
- The game (probably) cannot localize a sound / music data. The sound works if you delete CMAKE Install Dirs, but creating mess

@robertkirkman


IntinteDAO:
@robertkirkman @hansm629 All known bugs fixed, please check :)

tur/lugaru/build.sh
```diff
@@ -0,0 +1,18 @@
+TERMUX_PKG_HOMEPAGE=https://osslugaru.gitlab.io
+TERMUX_PKG_DESCRIPTION="Lugaru HD, free and open source ninja rabbit fighting game"
+TERMUX_PKG_LICENSE="GPL-2.0"
+TERMUX_PKG_MAINTAINER="@termux-user-repository"
```
licy183: It is better to set yourself as maintainer.

IntinteDAO: @licy183 
Thanks for reply. Should that be the Github nickname or the actual nickname? Probably the github one due to the fact that the bot will call me if the package doesn't work, but I prefer to inquire

licy183: It should be the nickname on GitHub.

robertkirkman:
This game does capture the Termux:X11 cursor inescapably and if you don't have a physical, hardware keyboard, it will unfortunately trap the Termux:X11 cursor inescapably. however, the Android OS cursor will still work and it will allow you to go to Termux App and type `pkill lugaru` which will escape this by force quitting the game.

Unfortunately, this problem is just the kind of thing that is common to all games generally that don't have any touchscreen buttons implemented. On a touchscreen-only device, including GNU/Linux as well, there are a lot of unreachable inputs and the game is unplayable.

tur/lugaru/build.sh
```diff
@@ -0,0 +1,19 @@
+TERMUX_PKG_HOMEPAGE=https://osslugaru.gitlab.io
+TERMUX_PKG_DESCRIPTION="Lugaru HD, free and open source ninja rabbit fighting game"
+TERMUX_PKG_LICENSE="GPL-2.0"
+TERMUX_PKG_MAINTAINER="@IntinteDAO"
+TERMUX_PKG_VERSION="1.2"
+TERMUX_PKG_SRCURL=https://github.com/osslugaru/lugaru/releases/download/${TERMUX_PKG_VERSION}/lugaru-${TERMUX_PKG_VERSION}.tar.xz
+TERMUX_PKG_SHA256=f3ea477caf78911c69939fbdc163f9f6517c7ef2267e716a0e050be1a166ef97
+TERMUX_PKG_DEPENDS="sdl2, glu, openal-soft, lugaru-data"
+TERMUX_PKG_EXTRA_CONFIGURE_ARGS="
+-DCMAKE_INSTALL_PREFIX=$TERMUX_PREFIX
```
robertkirkman: This line is automatically applied by `termux_step_configure_cmake()` and is not necessary to specify explicitly here.

tur/lugaru/build.sh
```diff
@@ -0,0 +1,19 @@
...
+-DSYSTEM_INSTALL=ON
+-DCMAKE_INSTALL_BINDIR=bin
+-DCMAKE_INSTALL_DATADIR=share/games
+-DCMAKE_POLICY_VERSION_MINIMUM=3.5
```
robertkirkman: This additional patch is also necessary in order to fix the second error that occurs with CMake 4:

```diff
--- a/CMakeLists.txt
+++ b/CMakeLists.txt
@@ -1,7 +1,6 @@
 project(lugaru)
 
 cmake_minimum_required(VERSION 3.0)
-cmake_policy(SET CMP0004 OLD)
 
 include(FindPkgConfig)
 include(GNUInstallDirs)
```

IntinteDAO:
@robertkirkman 
Check now. If everything will be OK, I will try to move to termux-packages (if will be not be merged to TUR)

tur/lugaru/build.sh
```diff
@@ -0,0 +1,18 @@
...
+TERMUX_PKG_MAINTAINER="@IntinteDAO"
+TERMUX_PKG_VERSION="1.2"
+TERMUX_PKG_SRCURL=https://github.com/osslugaru/lugaru/releases/download/${TERMUX_PKG_VERSION}/lugaru-${TERMUX_PKG_VERSION}.tar.xz
+TERMUX_PKG_SHA256=f3ea477caf78911c69939fbdc163f9f6517c7ef2267e716a0e050be1a166ef97
+TERMUX_PKG_DEPENDS="glu, lugaru-data, openal-soft, sdl2 | sdl2-compat"
+TERMUX_PKG_GROUPS="games"
+TERMUX_PKG_EXTRA_CONFIGURE_ARGS="
+-DSYSTEM_INSTALL=ON
+-DCMAKE_INSTALL_BINDIR=bin
+-DCMAKE_INSTALL_DATADIR=share/games
```
robertkirkman: If you don't change `CMAKE_INSTALL_BINDIR` and `CMAKE_INSTALL_DATADIR` here, what happens?

IntinteDAO: It is possible that the new version of cmake (because the game is probably not in the new version) fixes some problem. Previously, game files were installed in TERMUX_PREFIX/usr, which caused a lot of mess. Now they install correctly, so i removed those lines

tur/lugaru/build.sh
```diff
@@ -0,0 +1,17 @@
...
+TERMUX_PKG_MAINTAINER="@IntinteDAO"
+TERMUX_PKG_VERSION="1.2"
+TERMUX_PKG_SRCURL=https://github.com/osslugaru/lugaru/releases/download/${TERMUX_PKG_VERSION}/lugaru-${TERMUX_PKG_VERSION}.tar.xz
+TERMUX_PKG_SHA256=f3ea477caf78911c69939fbdc163f9f6517c7ef2267e716a0e050be1a166ef97
+TERMUX_PKG_DEPENDS="glu, lugaru-data, openal-soft, sdl2 | sdl2-compat"
+TERMUX_PKG_AUTO_UPDATE=true
+TERMUX_PKG_GROUPS="games"
+TERMUX_PKG_EXTRA_CONFIGURE_ARGS="
+-DSYSTEM_INSTALL=ON
+-DCMAKE_POLICY_VERSION_MINIMUM=3.5"
```
robertkirkman: Could you put this `"` symbol on the next line instead of the end of this line?

Most packages in Termux have that formatting.

PR #1818, opened 2025-06-13
addpkg(tur/topiary): 0.6.1

### LINKED ISSUE #1817: [Package]: topiary
### Name

topairy

### Home page URL

https://topiary.tweag.io/

### Source code URL

https://github.com/tweag/topiary

- [x] I understand [Tur Package Policy](https://github.com/termux-user-repository/tur/wiki/TUR-package-policy.).

### Additional information

_No response_

Juhan280:
Closes #1817

### DIFF
--- a/tur/topiary/build.sh
+++ b/tur/topiary/build.sh
@@ -1,0 +1,35 @@
+TERMUX_PKG_HOMEPAGE=https://topiary.tweag.io/
+TERMUX_PKG_DESCRIPTION="The universal code formatter"
+TERMUX_PKG_LICENSE="MIT"
+TERMUX_PKG_MAINTAINER="@termux-user-repository"
+TERMUX_PKG_VERSION="0.6.1"
+TERMUX_PKG_SRCURL=https://github.com/tweag/topiary/archive/refs/tags/v${TERMUX_PKG_VERSION}.tar.gz
+TERMUX_PKG_SHA256=e2bbac9cb46a3743cc41ca55245026580308722242c9df84bc0ef3cbb989aa81
+TERMUX_PKG_BUILD_IN_SRC=true
+TERMUX_PKG_AUTO_UPDATE=true
+TERMUX_PKG_UPDATE_TAG_TYPE=latest-release-tag
+
+termux_step_pre_configure() {
+	# clash with rust host build
+	# causes 32bit builds to fail if set
+	unset CFLAGS
+	termux_setup_rust
+}
+
+termux_step_make() {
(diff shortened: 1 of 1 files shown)


Juhan280:
This is the first time I wrote a build script from scratch. Let me know if I need to change anything

tur/topiary/build.sh
```diff
@@ -0,0 +1,34 @@
...
+
+termux_step_make() {
+	cargo build --jobs $TERMUX_PKG_MAKE_PROCESSES --target $CARGO_TARGET_NAME --release
+}
+
+termux_step_make_install() {
+	install -Dm755 -t $TERMUX_PREFIX/bin target/${CARGO_TARGET_NAME}/release/topiary
+
+	# Shell Completions
+	cargo run -r --target $CARGO_TARGET_NAME --bin topiary -- completion    zsh \
```
Biswa96: It failed because the host build container can not execute binaries for Android aarch64. Either it has to be built for the host system or the shell completions have to be generated in post-install script.

Biswa96: I prefer the later one because it reduces the build time and does not build the binary twice.

Juhan280: I am not sure how to do that, could you provide some assistance.

Juhan280: Just putting that part into `termux_step_post_make_install()` doesn't seem to work

Biswa96: see ripgrep build script for example.

tur/topiary/build.sh
```diff
@@ -0,0 +1,37 @@
...
+TERMUX_PKG_BUILD_IN_SRC=true
+TERMUX_PKG_AUTO_UPDATE=true
+TERMUX_PKG_UPDATE_TAG_TYPE=latest-release-tag
+
+termux_step_pre_configure() {
+	# clash with rust host build
+	# causes 32bit builds to fail if set
+	unset CFLAGS
+	# do not pin the version of the rust toolchain
+	rm -f rust-toolchain.toml
```
TomJo2000: That comment and `rm` command was specific to the `nerdfix` package.
`topiary` doesn't pin the rust version it uses, so this shouldn't be necessary.

Juhan280:
I think it is building fine now

tur/topiary/build.sh
TomJo2000: > see ripgrep build script for example.

Actually that currently generates those at install time.
I've opened a PR to fix that.
- termux/termux-packages#25052

PR #1857, opened 2025-07-01
nstool: new package

flosnvjx:


tur/nstool/build.sh
```diff
@@ -0,0 +1,29 @@
+TERMUX_PKG_HOMEPAGE="https://github.com/jakcron/nstool"
+TERMUX_PKG_DESCRIPTION="General purpose read/extract tool for Switch dumps"
+TERMUX_PKG_LICENSE="MIT"
+TERMUX_PKG_MAINTAINER="@flosnvjx"
+TERMUX_PKG_SRCURL="git+https://github.com/jakcron/nstool"
+TERMUX_PKG_VERSION="1.9.2"
+TERMUX_PKG_GIT_BRANCH=stable
```
licy183, line 7: Why not set `TERMUX_PKG_GIT_BRANCH` to `v$TERMUX_PKG_VERSION`?

flosnvjx: because the upstream doesn't always create versioned branch on release

Issue #1833, opened 2025-06-20
[Bug]: dosemu2 is outdated and have some issues


DuilioPerez:
### Problem description

The version available in TUR repo of dosemu have some issues, like double cursor, which was fixed [here](https://github.com/dosemu2/dosemu2/commit/aeb2d25be7632724d1e0c62382d69281034573fc) by the mantainer @stsp. It could be great if the available package is updated since, in myuse experoemce/ have better performance on Termux

### What steps will reproduce the bug?

Try to use a GUI over MS-DOS. It'll have double cursor and clicks won't work

### System information

termux-info:

```
Termux Variables:
TERMUX_APP_PACKAGE_MANAGER=apt
TERMUX_APP__APK_FILE=/data/app/~~LQzey9mCzS7FzKtqTYRMfw==/com.termux-GhIV57Ap-eqr_cw6kSgV0Q==/base.apk
TERMUX_APP__APK_RELEASE=F_DROID
TERMUX_APP__APP_VERSION_CODE=1020
TERMUX_APP__APP_VERSION_NAME=0.119.0-beta.1
TERMUX_APP__DATA_DIR=/data/user/0/com.termux
TERMUX_APP__IS_DEBUGGABLE_BUILD=false
TERMUX_APP__IS_INSTALLED_ON_EXTERNAL_STORAGE=false
TERMUX_APP__PACKAGE_NAME=com.termux
TERMUX_APP__PID=32298
TERMUX_APP__TARGET_SDK=28
TERMUX_VERSION=0.119.0-beta.1
TERMUX__SE_FILE_CONTEXT=u:object_r:app_data_file:s0:c33,c257,c512,c768
TERMUX__SE_INFO=default:targetSdkVersion=28:complete
TERMUX__SE_PROCESS_CONTEXT=u:r:untrusted_app_27:s0:c33,c257,c512,c768
TERMUX__UID=10289
TERMUX__USER_ID=0
Packages CPU architecture:
aarch64
Subscribed repositories:
# sources.list
deb https://packages-cf.termux.dev/apt/termux-main/ stable main
# tur-repo (sources.list.d/tur.list)
deb https://tur.kcubeterm.com tur-packages tur tur-on-device tur-continuous
# x11-repo (sources.list.d/x11.list)
deb https://packages-cf.termux.dev/apt/termux-x11/ x11 main
# glibc-repo (sources.list.d/glibc.list)
[...]

PR #1860, opened 2025-07-06
smallerc: fix PATH problems

stsp:
It was setting PATH in makefiles to just-built executables, failing the local build, as we need to use executables from smallerc-cross subpackage rather than the just-built ones.

Related: #1833


licy183:
Please also revbump this package, by setting `TERMUX_PKG_REVISION=1`

PR #1871, opened 2025-07-15
dosemu2: add gfx support

stsp:
It turned out dosemu2 was finding libx11 during build and was enabling the basic GUI support, even if I didn't ask it for that.

This patch enables the official GFX support via libsdl2. It disables outdated X support which currently needs libandroid-shmem and its not being linked in.

Related: #1833

### DIFF
--- a/tur/dosemu2/build.sh
+++ b/tur/dosemu2/build.sh
@@ -3,12 +3,12 @@ TERMUX_PKG_DESCRIPTION="Run DOS programs under linux."
 TERMUX_PKG_LICENSE="GPL-2.0"
 TERMUX_PKG_MAINTAINER="@stsp"
 TERMUX_PKG_VERSION="2.0pre9-git"
-TERMUX_PKG_REVISION=2
+TERMUX_PKG_REVISION=3
 TERMUX_PKG_SRCURL=git+https://github.com/dosemu2/dosemu2
 TERMUX_PKG_GIT_BRANCH=devel
 _COMMIT=58cd68c2bbbf1976bbb80d39c4ddd8edd2dd93ea
 TERMUX_PKG_SHA256=e2fcd6e79091442f22fc61f24cac3c9e21dbe47df31ca3ae3f65b8dc457c3c72
-TERMUX_PKG_BUILD_DEPENDS="libandroid-posix-semaphore, libandroid-glob, slang, libao, fluidsynth, ladspa-sdk, libslirp, libbsd, readline, json-c, libseccomp, libsearpc, fdpp, dj64dev"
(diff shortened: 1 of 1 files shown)


tur/dosemu2/build.sh
```diff
@@ -3,12 +3,12 @@ TERMUX_PKG_DESCRIPTION="Run DOS programs under linux."
...
 TERMUX_PKG_MAINTAINER="@stsp"
 TERMUX_PKG_VERSION="2.0pre9-git"
-TERMUX_PKG_REVISION=2
+TERMUX_PKG_REVISION=3
 TERMUX_PKG_SRCURL=git+https://github.com/dosemu2/dosemu2
 TERMUX_PKG_GIT_BRANCH=devel
 _COMMIT=58cd68c2bbbf1976bbb80d39c4ddd8edd2dd93ea
 TERMUX_PKG_SHA256=e2fcd6e79091442f22fc61f24cac3c9e21dbe47df31ca3ae3f65b8dc457c3c72
-TERMUX_PKG_BUILD_DEPENDS="libandroid-posix-semaphore, libandroid-glob, slang, libao, fluidsynth, ladspa-sdk, libslirp, libbsd, readline, json-c, libseccomp, libsearpc, fdpp, dj64dev"
+TERMUX_PKG_BUILD_DEPENDS="libandroid-posix-semaphore, libandroid-glob, slang, libao, fluidsynth, ladspa-sdk, libslirp, libbsd, readline, json-c, libseccomp, libsearpc, sdl2, sdl2-ttf, fontconfig, fdpp, dj64dev"
```
robertkirkman: Probably, it should be `sdl2 | sdl2-compat` instead of `sdl2`, like this:

https://github.com/termux-user-repository/tur/blob/0a14bec1692cf3069089c6ab59e75992433ff385/tur/wine-stable/build.sh#L10

robertkirkman (changes requested):

tur/dosemu2/build.sh
stsp: Will it help if I put just
`sdl2-compat` to avoid any
non-determinism?

robertkirkman: well, the problem is that I think this syntax is what allows people to choose to have `sdl2` installed if they don't want `sdl2-compat` or also for other people to have `sdl2-compat` if they don't want `sdl2`, at the same time.

stsp: Ok, done.

stsp:
Oops.

stsp:
Needs #1874

robertkirkman (approved):

PR #1876, opened 2025-07-26
icewm: moved to main

stsp:
[skip ci]


robertkirkman:
I'm sorry, actually I didn't know that IceWM was already in TUR and I made a different version.

@IntinteDAO if you like, could you please test the new one I made when you get a chance and let me know if there is anything wrong about it?

PR #1886, opened 2025-08-12
Newpkg: megaglest

IntinteDAO:
Things to improve:
- The game does not install game data files, as they are located in a different source. This needs to be fixed so that the game and game data are installed at the same time.
https://github.com/MegaGlest/megaglest-data/releases/download/3.13.0/megaglest-standalone-data-3.13.0.tar.xz

- Menu entries and icons are in the game data file, so they need to be copied during installation.

If you can help @robertkirkman 

Game works too slow for me, however Zink works slower every month :P Does not work on virglrenderer Android (for me), but works on llvmpipe (but unplayable)


tur/megaglest/build.sh
```diff
@@ -0,0 +1,32 @@
...
+TERMUX_PKG_VERSION=3.13.0
+TERMUX_PKG_SRCURL=https://github.com/MegaGlest/megaglest-source/archive/refs/tags/${TERMUX_PKG_VERSION}.tar.gz
+TERMUX_PKG_SHA256=e02e58c2329558cc5d67374b5e5f9b3cfaafc300b96feff71df8d4b0d39e1eaa
+TERMUX_PKG_DEPENDS="glib, openal-soft, sdl2, libvorbis, libjpeg-turbo, libpng, freetype, curl, libxml2, wxwidgets, lua52, fribidi, libandroid-glob, libftgl2, glew"
+
+TERMUX_PKG_EXTRA_CONFIGURE_ARGS="
+-DCMAKE_POLICY_VERSION_MINIMUM=3.5
+-DLUA_LIBRARIES=${TERMUX_PREFIX}/lib/liblua5.2.so
+-DLUA_INCLUDE_DIR=${TERMUX_PREFIX}/include/lua5.2
+-DLUA_MATH_LIBRARY=$TERMUX_STANDALONE_TOOLCHAIN/sysroot/usr/lib/$TERMUX_HOST_PLATFORM/$TERMUX_PKG_API_LEVEL/libm.so
```
robertkirkman, line 14: If you just write `-DLUA_MATH_LIBRARY=m` does that work instead and prevent the error in CI? Yes I know that some packages like `neovim` do this, but it just looks like a big mess, maybe it's fine with only `-DLUA_MATH_LIBRARY=m`. If it really does have to be like this then you can move this to `termux_step_pre_configure()` (like it is in `neovim`)

robertkirkman (changes requested):

tur/megaglest/build.sh
```diff
@@ -0,0 +1,32 @@
...
+"
+
+termux_step_pre_configure() {
+	export LDFLAGS+=" -landroid-glob -Wl,--no-as-needed,-lOpenSLES,--as-needed"
+	export CFLAGS+=" -D__OpenBSD__"
+	export CXXFLAGS+=" -D__OpenBSD__"
+}
+
+termux_step_post_make_install() {
+	echo "POST INSTALL for Menu entries"
```
robertkirkman, line 31: What does this do? (is it a placeholder for WIP function?)

tur/megaglest/build.sh
robertkirkman: Ok I see when I test, it does need either that or a patch, so put it in `termux_step_pre_configure()` like `neovim`:

```bash
termux_step_pre_configure() {
	TERMUX_PKG_EXTRA_CONFIGURE_ARGS+=" -DLUA_MATH_LIBRARY=$TERMUX_STANDALONE_TOOLCHAIN/sysroot/usr/lib/$TERMUX_HOST_PLATFORM/$TERMUX_PKG_API_LEVEL/libm.so"

```

tur/megaglest/build.sh
robertkirkman: > The game does not install game data files, as they are located in a different source. This needs to be fixed so that the game and game data are installed at the same time.

You can add additional URLs in `TERMUX_PKG_SRCURL`, like this:

```bash
TERMUX_PKG_SRCURL=(https://github.com/MegaGlest/megaglest-source/archive/refs/tags/${TERMUX_PKG_VERSION}.tar.gz
					https://github.com/MegaGlest/megaglest-data/releases/download/3.13.0/megaglest-standalone-data-${TERMUX_PKG_VERSION}.tar.xz)
TERMUX_PKG_SHA256=(e02e58c2329558cc5d67374b5e5f9b3cfaafc300b96feff71df8d4b0d39e1eaa
					996040acfb338cfcbd0612b13db9c1734a7588bcc60c7e9b87e9fd619629dda3)
```

this will make the megaglest-standalone-data be extracted somewhere in `$TERMUX_PKG_SRCDIR` too, in this case mixed into the other sources, and you can try to install the assets from there in `termux_step_post_make_install()`.

IntinteDAO: > What does this do? (is it a placeholder for WIP function?)

Yep. I had a issues with that. I usually create a package with 1 link, not 2s and I don't know how to work on 2 archives at once.

robertkirkman: > Yep. I had a issues with that. I usually create a package with 1 link, not 2s and I don't know how to work on 2 archives at once.

I have given the code block which you can copy and paste in order to start with it, but if you still don't understand it, let me know and later I can try to finish the entire thing and send it

PR #1901, opened 2025-09-17
blueprint-compiler: moved to main [skip ci]

stsp:


robertkirkman (approved):
Thank you, I did not know `blueprint-compiler` was in TUR but it is needed in the main repo now so thanks.

Also, I independently implemented my `build.sh` of `blueprint-compiler` differently, this `blueprint-compiler` looks kind of weird and I am not sure why it was implemented this way unless the custom steps were required for an old version of it, so I don't think the new `blueprint-compiler` needs to copy anything from this version.

PR #1948, opened 2025-10-04
python-scipy: rev-bump

stsp:
Version of 'python-scipy' has not changed.
Either 'TERMUX_PKG_REVISION' or 'TERMUX_PKG_VERSION' need to be modified in the build.sh when changing a package build. Alternatively you can add '[no version check]'.
To the commit message to skip this check.


stsp:
Seem fixed in master while
I was at it. :)

PR #2013, opened 2025-10-27
bump(tur/openjdk-11): 11.0.29

robertkirkman:
- Also add all applicable improvements that were introduced to the github.com/termux/termux-packages `openjdk-17`, `openjdk-21` and `openjdk-25` packages in between April 2024 and October 2025
  - Consult https://github.com/termux/termux-packages/commits/master/packages/openjdk-17/build.sh for a general overview
  - New `build.sh` follows a structure derived from a combination of the `openjdk-17` and `openjdk-25` packages
    - `termux_step_create_debscripts()` based on that of `openjdk-17`
    - `termux_step_host_build()` based on that of `openjdk-25`

- LTO unfortunately not working although it does in later JDKs
  - Error: `ld: error: /home/builder/.termux-build/openjdk-11/src/build/linux-aarch64-normal-server-release /buildjdk/hotspot/variant-server/libjvm/objs/abstractCompiler.o: Invalid record`


tur/openjdk-11/build.sh
```diff
@@ -81,53 +129,107 @@ termux_step_make_install() {
 }
 
 termux_step_post_make_install() {
-	# Install ca-certs
-	mkdir -p $TERMUX_PREFIX/lib/jvm/java-11-openjdk/lib/security/
-	ln -sfr $TERMUX_PREFIX/lib/jvm/java-17-openjdk/lib/security/jssecacerts \
```
robertkirkman, line 86: @licy183 I have a question, do you know what this part is for, and is there a way I can test if it is still required or not?

It looks like it references files in the package `openjdk-17`, if it is still necessary to keep, then does that mean that this should depend on `openjdk-17`?

robertkirkman: I am not sure what the difference is because in each OpenJDK version it currently looks the same, and to me it seemed unnecessary to keep that part of `termux_step_post_make_install()`. I just didn't want to remove it without letting you know

```
~ $ apt-cache policy openjdk-11 openjdk-17 openjdk-21 openjdk-25
openjdk-11:
  Installed: 11.0.29
  Candidate: 11.0.29
  Version table:
 *** 11.0.29 100
        100 /data/data/com.termux/files/usr/var/lib/dpkg/status
     11.0.23+9-1 500
        500 https://tur.kcubeterm.com tur-packages/tur aarch64 Packages
openjdk-17:
  Installed: 17.0.17
  Candidate: 17.0.17
  Version table:
 *** 17.0.17 500
        500 https://packages-cf.termux.dev/apt/termux-main stable/main aarch64 Packages
        100 /data/data/com.termux/files/usr/var/lib/dpkg/status
openjdk-21:
  Installed: 21.0.9
  Candidate: 21.0.9
  Version table:
 *** 21.0.9 500
        500 https://packages-cf.termux.dev/apt/termux-main stable/main aarch64 Packages
        100 /data/data/com.termux/files/usr/var/lib/dpkg/status
openjdk-25:
  Installed: 25.0.1
  Candidate: 25.0.1
  Version table:
 *** 25.0.1 500
        500 https://packages-cf.termux.dev/apt/termux-main stable/main aarch64 Packages
        100 /data/data/com.termux/files/usr/var/lib/dpkg/status
~ $ dpkg -L openjdk-11 openjdk-17 openjdk-21 openjdk-25 | grep cacerts
/data/data/com.termux/files/usr/lib/jvm/java-11-openjdk/lib/security/cacerts
/data/data/com.termux/files/usr/lib/jvm/java-17-openjdk/lib/security/cacerts
/data/data/com.termux/files/usr/lib/jvm/java-21-openjdk/lib/security/cacerts
/data/data/com.termux/files/usr/lib/jvm/java-25-openjdk/lib/security/cacerts
~ $ 
```

licy183: They are for `ca-certificates-java`. 

https://github.com/termux/termux-packages/blob/a511bf8dba3fe9a65bd33317e00d5539b33196c4/packages/ca-certificates/build.sh#L30

robertkirkman: Thanks, in this case, should `openjdk-21` and `openjdk-25` be modified to have these lines, instead of removing the lines from `openjdk-11`?

Before I did anything, `openjdk-21` never had these lines, so I did not know about this.

licy183: I'm not sure as I hardly use java on Termux...

robertkirkman: Ok, when I check what most desktop Linux distros do, I see that they actually completely delete the original `lib/security/cacerts` file in all their OpenJDK packages and then replace it with a symbolic link to an empty location that doesn't have any package providing it statically, but which has a `cacerts` file (not named `jssecacerts`) dynamically autogenerated into that location using a complicated `postinst` script in their `ca-certificates-java` packages,

https://salsa.debian.org/openjdk-team/openjdk/-/blob/91845e5b3b165526c1b7939595636e33cdf9070b/debian/rules#L1430

https://salsa.debian.org/java-team/ca-certificates-java/-/blob/4bb1fb455129ebcefd916271500b2121bf8f5cd9/debian/ca-certificates-java.postinst#L99

that isn't really the same as what Termux's `ca-certificates-java` package does and I don't want to mess anything up, so I will leave all the ca-certificates parts of Termux OpenJDK packages unchanged until there is a reason to change them, like a bug report regarding this. I have restored the `jssecacerts` lines in the `build.sh` of this `openjdk-11`.

PR #2019, opened 2025-10-28
bump(tur/ffplayout): 0.25.7

robertkirkman:
- Fixes https://github.com/termux-user-repository/tur/issues/2018

- Fixes https://github.com/termux-user-repository/tur/issues/1543

- Seems to work now. Lots of errors kept happening at runtime, so I just kept fixing the errors and eventually it started working. This mostly involved greatly expanding `0001-fix-hardcoded-paths.patch` and then implementing the `postinst` and `postrm` debscripts.

Screenshot_20251028-053754_Termux_X11


robertkirkman:
Useful notes:

Setup step 1: `ffplayout --drop-db`. Might be necessary if seeing an error `Error: DB("migration 1 was previously applied but has been modified")` after attempting to use the previous version that was not working.

Setup step 2: `ffplayout -i`. Follow the directions to configure initial settings.

Setup step 3: `ffplayout -l 127.0.0.1:8787`. Launches the server

PR #2037, opened 2025-11-05
revbump({tur,tur-on-device}): for `imath` 3.2.2, `openvdb` 13.0.0

robertkirkman:
- Necessary after https://github.com/termux/termux-packages/pull/27105


robertkirkman:
It built successfully, so my changes to `perl` and `setup-termux.sh` in github.com/termux/termux-packages worked, and this is ready now.

Issue #2043, opened 2025-11-09
[Bug]: blender4 needs rebuild


lfdevs:
### Problem description

@robertkirkman Hello, sir! I think the package `blender4` needs rebuild. Some shared libraries have been updated, causing version mismatches.

### What steps will reproduce the bug?

```
~ $ blender-4.5
CANNOT LINK EXECUTABLE "blender-4.5": library "libImath-3_1.so" not found: needed by main executable
~ $ apt list blender4
Listing... Done
blender4/tur-packages,now 4.5.4 aarch64 [installed]
~ $ apt list imath*
Listing... Done
imath/stable,now 3.2.2 aarch64 [installed]
~ $ ls -ahl $PREFIX/lib/libImath*
-rw-------. 1 u0_a410 u0_a410 325K Nov  4 17:14 /data/data/com.termux/files/usr/lib/libImath-3_2.so
lrwxrwxrwx. 1 u0_a410 u0_a410   15 Nov  9 00:02 /data/data/com.termux/files/usr/lib/libImath.so -> libImath-3_2.so
```

### System information

termux-info:

```
Termux Variables:
TERMUX_APP_PACKAGE_MANAGER=apt
TERMUX_APP__APK_FILE=/data/app/~~BGyvws6FD6I3_gJQ5HBsrw==/com.termux-k6UTceidu0oDxr2Ep42Maw==/base.apk
TERMUX_APP__APK_RELEASE=GITHUB
TERMUX_APP__APP_VERSION_CODE=1022
TERMUX_APP__APP_VERSION_NAME=0.119.0-beta.3
TERMUX_APP__DATA_DIR=/data/user/0/com.termux
TERMUX_APP__IS_DEBUGGABLE_BUILD=true
TERMUX_APP__IS_INSTALLED_ON_EXTERNAL_STORAGE=false
TERMUX_APP__LEGACY_DATA_DIR=/data/data/com.termux
TERMUX_APP__PACKAGE_NAME=com.termux
TERMUX_APP__PID=16168
TERMUX_APP__SE_FILE_CONTEXT=u:object_r:app_data_file:s0:c154,c257,c512,c768
TERMUX_APP__SE_INFO=default:targetSdkVersion=28:complete
TERMUX_APP__TARGET_SDK=28
TERMUX_VERSION=0.119.0-beta.3
TERMUX__APPS_DIR=/data/user/0/com.termux/termux/apps
TERMUX__HOME=/data/data/com.termux/files/home
[...]

PR #2044, opened 2025-11-09
fix(ci/package_on_device): do not conflict `gh release upload` `archive` filename with `packages.yml`

robertkirkman:
- Retry https://github.com/termux-user-repository/tur/pull/2037

- Attempt to fix https://github.com/termux-user-repository/tur/issues/2043

### DIFF
--- a/common-files/building-system-patches/0009-allow-lint-packages-version-check.patch
+++ b/common-files/building-system-patches/0009-allow-lint-packages-version-check.patch
@@ -1,0 +1,15 @@
+'git reset HEAD~' is TEMPORARY - work around bug in lint-packages.sh
+will discuss with Tomjo2000 and fix upstream later
+
+--- a/scripts/lint-packages.sh
++++ b/scripts/lint-packages.sh
+@@ -124,7 +124,8 @@ check_indentation() {
+ # We'll need the 'origin/master' as a base commit when running the version check.
+ # So try fetching it now if it doesn't exist.
++git reset HEAD~
+ if ! base_commit="HEAD~$(git rev-list --count FETCH_HEAD..)"; then
+-	git fetch https://github.com/termux/termux-packages.git
++	git fetch https://github.com/termux-user-repository/tur.git
+ 	base_commit="HEAD~$(git rev-list --count FETCH_HEAD..)"
+ fi
+ 


.github/workflows/package_on_device.yml
```diff
@@ -282,8 +282,10 @@ jobs:
       run: |
         GITHUB_SHA=${{ github.sha }}
         for archive in debs-*/debs-{aarch64,arm,i686,x86_64}-${{ github.sha }}.tar; do
-          gh release upload -R https://github.com/termux-user-repository/tur "0.1" $archive
-          echo "$archive  uploaded"
+          tur_on_device_archive="${archive//.tar/-tur-on-device.tar}"
+          mv "$archive" "$tur_on_device_archive"
+          gh release upload -R https://github.com/termux-user-repository/tur "0.1" "$tur_on_device_archive"
```
robertkirkman, line 287: @licy183 do you think this will work?

I don't want to separate packages of `tur` and `tur-on-device` folders into separate PRs,

instead, I want to try to fix it permanently so that the problem never occurs again.

However, unfortunately, I don't understand exactly what causes the problem, only that there was an error with this file conflicting because both workflows ran in 1 PR.

Can the code which takes the `unprocessed_debfile` https://github.com/termux-user-repository/tur/releases/tag/0.1 and sends it to https://tur.kcubeterm.com  automatically handle the filename sometimes being something different, like `debs-{aarch64,arm,i686,x86_64}-${{ github.sha }}-tur-on-device.tar`, or will that not work and it has to be exactly `debs-{aarch64,arm,i686,x86_64}-${{ github.sha }}.tar` every time?

common-files/building-system-patches/0009-allow-lint-packages-version-check.patch
```diff
@@ -0,0 +1,15 @@
+'git reset HEAD~' is TEMPORARY - work around bug in lint-packages.sh
+will discuss with Tomjo2000 and fix upstream later
```
robertkirkman, line 2: Have discussed now, and we think that it will get fixed by today's update to this:

- https://github.com/termux/termux-packages/pull/27095

this patch can be rebased to remove `git reset HEAD~` after that PR is merged.

.github/workflows/package_on_device.yml
licy183: This should work correctly. Currently, the scripts in the dists repo do not validate filenames.

licy183: It doesn't work as the CI reuse can't pick up the files uploaded by the on-device worflow. The best way to avoid this is avoid building tur-on-device packages and tur packages in the same MR.

robertkirkman: oh no... actually, I did not ever know whether CI reuse would actually work with tur-on-device or not.

The reason why I pasted the lines for it in this commit https://github.com/termux-user-repository/tur/commit/70cc0df401b48f24cbc4a562aa7f195367ff3e8a was because if I didn't do that, a build error occurred because of a variable or setting of some kind missing that prevented building all tur-on-device packages at all after the merge of CI reuse in the main repository.

I don't know how to fix it, so it is necessary to disable CI reuse for all PRs which modify tur-on-device packages.

I am trying again here

- https://github.com/termux-user-repository/tur/pull/2045

licy183: In fact, CI reuse can work on tur-on-device.

The situation where it doesn't work properly is when the packages from tur and tur-on-device are modified in the same PR.

I think there is no need to revert it. I've uploaded them manually.

robertkirkman: > In fact, CI reuse can work on tur-on-device.

I thought that it didn't because I saw this log, in an earlier PR where I modified only tur-on-device packages:

image

https://github.com/termux-user-repository/tur/actions/runs/19011632972/job/54293393376

Do you know of an example where it worked?

licy183: I think it should work, but I actually don't know whether it works. The most important thing is to avoid modifying `build.sh`s in both `tur` and `tur-on-device` in the same PR.

robertkirkman: I want to fix it so that modifying `build.sh` in both `tur` and `tur-on-device` in the same PR is possible,

because I do not want anyone else to ever encounter the same problem, and the draft PR I opened can do that

However, it does it by disabling the CI reuse for `tur-on-device` PRs.

I will try to figure out any possible way to do it without disabling the CI reuse.

I believe that part of the problem is that the CI reuse has never worked properly for `tur-on-device` packages (which can be seen in the log above), and fixing that might be where I can start, I don't know how to fix it yet but I will try

robertkirkman: See, it appears that the CI reuse still does not ever work in tur-on-device even if only tur-on-device packages are modified in the PR.

image

I still will continue trying to figure out how to fix it

PR #2057, opened 2025-11-18
tree-wide: rebuild for FFmpeg 8

robertkirkman:
- It is too hard to make Blender 3.6 work with FFmpeg 8, but this works
- It is easy to make Blender 4 work with FFmpeg 8
- All of the other packages in TUR that depend on `ffmpeg` appear to have source code that only invokes the `ffmpeg` executable, and they do not open the `.so` files of FFmpeg, so they do not need to be recompiled

### DIFF
--- a/tur-on-device/blender4/build.sh
+++ b/tur-on-device/blender4/build.sh
@@ -4,8 +4,7 @@ TERMUX_PKG_DESCRIPTION="A fully integrated 3D graphics creation suite"
 # https://www.blender.org/about/license/
 TERMUX_PKG_LICENSE="GPL-3.0-or-later"
 TERMUX_PKG_MAINTAINER="@termux-user-repository"
-TERMUX_PKG_VERSION=4.5.4
-TERMUX_PKG_REVISION=2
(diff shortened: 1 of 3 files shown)


common-files/building-system-patches/0002-fix-auto-update.patch
```diff
@@ -19,15 +19,6 @@ Only search packages from TUR repository when auto-updating.
  				if [ -d "${repo_dir}/${pkg}" ]; then
  					pkg="${repo_dir}/${pkg}"
  					break
-@@ -352,7 +352,7 @@
- 	local body
- 
- 	if [[ "${assignee:-termuxbot2}" == "termuxbot2" ]]; then
--		assignee="MrAdityaAlok" # Assign myself if termuxbot2 is the actor.
-+		assignee="licy183" # Assign myself if termuxbot2 is the actor.
```
robertkirkman, line 27: I am assuming that just removing this is the correct change, since MrAdityaAlok removed this upstream.

robertkirkman:
The cloudflare and github outages are gone, so this will be ready when it finishes building

PR #2066, opened 2025-11-22
New package: furnace

ghost:
@licy183 
Before attempting to build this, could you please review the buildscript and tell me if I'm missing something? Thanks in advance.


tur/furnace/build.sh
```diff
@@ -0,0 +1,35 @@
+TERMUX_PKG_HOMEPAGE=https://github.com/tildearrow/furnace
+TERMUX_PKG_DESCRIPTION="A multi-system chiptune tracker compatible with DefleMask modules"
+TERMUX_PKG_LICENSE="GPL-2.0-or-later"
+TERMUX_PKG_MAINTAINER="@ChunkyHamsterfly"
+TERMUX_PKG_VERSION=0.6.8.3
+TERMUX_PKG_SRCURL=https://github.com/tildearrow/furnace/archive/refs/tags/v${TERMUX_PKG_VERSION}.tar.gz
+TERMUX_PKG_SHA256=c54b93dc7211a8e86828ea6b436628920df7a7dc87b130b3752f2251f5466e9c
+TERMUX_PKG_DEPENDS="alsa-lib, fftw, fmt, freetype, libglvnd, libsndfile, libx11, pulseaudio, sdl2, zlib"
```
robertkirkman, line 8: It could have `sdl2 | sdl2-compat` instead of just `sdl2` so that it can be used by people using `sdl2-compat` also. Also, instead of `libglvnd`, it can probably have `opengl` specified like other packages which depend on that

PR #2087, opened 2025-12-02
bump(tur/code-server-extension-codelldb): 1.12.0

robertkirkman:
- Fixes https://github.com/termux-user-repository/tur/issues/2083


common-files/building-system-patches/0002-fix-auto-update.patch
```diff
@@ -1,18 +1,30 @@
...
 --- a/scripts/bin/update-packages
 +++ b/scripts/bin/update-packages
-@@ -307,7 +307,7 @@
+@@ -336,6 +336,11 @@ _should_update() {
+ 		termux_error_exit "directory '${pkg_dir}' is not a package."
+ 	fi
+ 
++	if [[ "$(realpath "$(dirname "${pkg_dir}")")" != *"tur"* ]]; then
++		echo "INFO: Skipping '$(basename "${pkg_dir}")', it is not a TUR package."
++		return 1
```
robertkirkman, line 10: I don't know if it will work or not until it is tried, but this is my attempt to fix the problem of TUR auto update opening issues of update failures for packages that are in github.com/termux/termux-packages, not in TUR.

PR #2115, opened 2025-12-13
fix(tur/gwc): fix all errors differently

robertkirkman:
- Cross-compile using `termux-proot-run`

- Lower source archive to stable, then cherry-pick back to upstream

- Include `stdlib.h` during two `configure` checks

- Include `gwc.h` in `tap_reverb.c`

- Initialize `enc_format` in `gwc.c` with `0` instead of `NULL`

- Use the default `configure` entrypoint instead of moving a different one around

- Do not bypass errors by disabling warnings

### DIFF
--- a/common-files/building-system-patches/0009-disable-broken-linter-function.patch
+++ b/common-files/building-system-patches/0009-disable-broken-linter-function.patch
@@ -1,0 +1,13 @@
+check_version() in lint-packages.sh is broken
+see: https://github.com/termux/termux-packages/pull/26980#discussion_r2456542227
+
+--- a/scripts/lint-packages.sh
++++ b/scripts/lint-packages.sh
+@@ -147,6 +147,7 @@ check_indentation() {
+ no_build="$(git log --fixed-strings --grep '%ci:no-build' --pretty=format:%H "$base_commit")"
+ 
+ check_version() {
++	return
+ 	local package_dir="${1%/*}"
+ 
+ 	[[ -z "$base_commit" ]] && {


common-files/building-system-patches/0008-allow-reuse-pr.patch
```diff
@@ -36,16 +44,34 @@
...
+-			jq -r '[.data.repository.object?.checkSuites?.nodes[]?
++			jq -r "[.data.repository.object?.checkSuites?.nodes[]?
+ 				| select(
+-				.workflowRun.event == "pull_request"
+-				and .workflowRun.file.path == ".github/workflows/packages.yml"
+-				and .conclusion == "SUCCESS"
+-				and .status == "COMPLETED")
+-				| .workflowRun.databaseId][0] // empty' <<< "$RESPONSE" || :
++				.workflowRun.event == \"pull_request\"
++				and .workflowRun.file.path == \"$WORKFLOW_PATH\"
```
robertkirkman, line 63: @licy183 this is my attempt to implement the ability to edit `tur` folder and `tur-on-device` folder in the same PR. i hope it will work but I don't know for sure unless tried, I want to try it using this PR.

flosnvjx:
Thanks for working on this. I'm not familiar with CI so for testing this I cherrypicked the gwc packaging part in my fork. And for cherrypicked commit 66fcbd1e7bb92b3a2aa8a413d9fc2a1df7bd5762, it does need to comment out the `
TERMUX_PKG_REVISION=1
` line to build. The build works fine, so for packaging part, I'm principally OK with it.

While attempting to understand the CI part I realized that the `TERMUX_PKG_VERSION` in my initial commit contains an unallowed hypen (`-`) character, as per [`deb-version(7)`](https://www.man7.org/linux/man-pages/man7/deb-version.7.html):

> If there is no *debian-revision* then hyphens are not allowed;

So if I understand this correctly, we should replace the hypen in `TERMUX_PKG_VERSION`. And it seems that Debian folks use a dot (as per [repology](https://repology.org/project/gwc/versions))

The lint script (specifically, `check_version()`) may also need some improvement to check this.

robertkirkman:
> So if I understand this correctly, we should replace the hypen in TERMUX_PKG_VERSION. And it seems that Debian folks use a dot (as per [repology](https://repology.org/project/gwc/versions))

Unfortunately, since you already set the version to be like that, we can't, without doing something called "epoch bump" which adds some annoyance. That is why I didn't change the version in this - I have to use the version you already chose, otherwise people wouldn't be able to upgrade correctly if they already installed it.

overall, it having a hyphen is not that bad, because Termux does not follow those rules of Debian or Arch Linux that you linked. There are dozens of packages in the main Termux repository that have hyphens in their versions. `tcc`, `gnushogi`, `prover9`, `dialog`, `sdl-image`, `usbmuxd`, it's a long list.

> The lint script (specifically, check_version()) may also need some improvement to check this.

The reason I disable that here is because of a problem unrelated to this specific package or the version format of this package, see the link in the comment of that patch for more details.

In this case, the problem there is that `lint-packages.sh` does not allow moving packages between folders, which is not an intended behavior of `lint-packages.sh` as explained in the linked thread, so that function of `lint-packages.sh` is broken and needs to be disabled until it's fixed.

robertkirkman:
> It appears to be that if this package triggers both `tur` and `tur-on-device` workflow, both will perform `ci reuse` and fail due to the file confliction when uploading to GitHub Release.

Yes, you are right. I thought that the code which renamed the archive in the tur-on-device workflow would not be necessary anymore, but actually it is.

image

image

robertkirkman:
Actually, now I understand that the name of the archive is not based on the workflow number, but actually it's just based on the number of the last commit from the branch after merging to master, I didn't notice this before, but if I had then I would have understood that of course the number in the archive name would be the same for both workflows..

image

PR #2149, opened 2026-01-04
Newpkg: search and rescue 2 (draft before pushing to termux-packages)

IntinteDAO:
The game should work without any problems. It is possible that the “fix_sconstruct.patch” file should be split into several smaller ones.

I must admit that I have never used scons, so I don't know if it can be written more elegantly. However, it works :) I will definitely work on it, because the serf package also uses scons, but it installs everything in termux_step_make_install.

### DIFF
--- a/tur/searchandrescue2/fix_sconstruct.patch
+++ b/tur/searchandrescue2/fix_sconstruct.patch
@@ -1,0 +1,68 @@
+diff -ur orig/src/config.h mod/src/config.h
+--- orig/src/config.h	2023-08-11 00:46:34.000000000 +0200
++++ mod/src/config.h	2025-12-24 12:01:55.064016834 +0100
+@@ -502,7 +502,7 @@
+ #define SAR_DEF_LOCAL_DATA_DIR  ".config/sar2"
+ 
+ /* Game data dir */
+-#define SAR_DEF_GLOBAL_DATA_DIR "/usr/share/sar2"
++#define SAR_DEF_GLOBAL_DATA_DIR "@TERMUX_PREFIX@/share/games/sar2"
+ 
+ /* Subdirs, one of each in the local and global game dirs */
+ #define SAR_DEF_AIRCRAFTS_DIR		"aircrafts"
+@@ -564,7 +564,7 @@
+ #ifdef __MSW__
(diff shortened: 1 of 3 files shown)


tur/searchandrescue2/searchandrescue2-data.subpackage.sh
```diff
@@ -0,0 +1,6 @@
+TERMUX_SUBPKG_DESCRIPTION="Data files for search and rescue 2 (metapackage)"
+TERMUX_SUBPKG_PLATFORM_INDEPENDENT=true
+TERMUX_SUBPKG_DEPEND_ON_PARENT=no
```
robertkirkman, line 3: This line has no effect for 2 reasons:

1. it would be `false` instead of `no`
2. There is code that automatically makes `TERMUX_SUBPKG_DEPEND_ON_PARENT=false` when the subpackage is in `TERMUX_PKG_DEPENDS` of the parent package so you don't need to specify this anyway

robertkirkman: Also this is not technically a "metapackage" so maybe the description should be adjusted to remove that

tur/searchandrescue2/fix_sconstruct.patch
```diff
@@ -0,0 +1,68 @@
...
++
++env.Append(CPPFLAGS = ['-I' + sys_prefix + '/include'])
++env.Append(LIBPATH = [sys_prefix + '/lib'])
++
++env.Append(LINKFLAGS = [
++    '-L' + sys_prefix + '/lib',
++    '-Wl,-rpath,' + sys_prefix + '/lib',
++    '-Wl,--enable-new-dtags',
++    '-landroid-spawn',
++    '-Wl,--no-as-needed,-lOpenSLES,--as-needed'
```
robertkirkman, line 45: Maybe it's possible to get `LDFLAGS` using `os.environ.get()` like you did for `PREFIX` instead of having to write `-landroid-spawn`, etc. twice?

Also, similar for `CPPFLAGS`,

Also, For `PREFIX` itself, I recommend writing this: `@TERMUX_PREFIX@` instead of `os.environ.get('PREFIX'` because `@TERMUX_PREFIX@` will be automatically replaced in `.patch` files for slightly less code required, but that's up to you

PR #2163, opened 2026-01-10
tree-wide: disable packages that exist in upstream termux-packages

robertkirkman:
- These 64-bit packages exist in https://github.com/termux/termux-packages, so they should have 64-bit disabled in TUR to avoid causing dependency conflicts and confusion.

```
arpack-ng
octave-x
octave
python-scipy
qrupdate-ng
suitesparse
sundials
```

- This package exists in both repositories with all architectures except for 32-bit ARM:

```
wine-stable
```

- This package exists in both repositories with all architectures, so it should be completely disabled in TUR:

```
virglrenderer-android
```

- This package depends on `python-scipy` in both repositories at the same time, so it needs to be disabled for 64-bit targets in TUR, then re-enabled for 64-bit targets in github.com/termux/termux-packages by moving it there later:

```
python-scikit-image
```

Packages that do not target aarch64 for dependency infrastructure reasons cannot auto update because of '`ERROR: Obtaining update order failed`'. They can have auto update re-enabled for 64-bit targets after those are moved to github.com/termux/termux-packages.


tur/octave-x/build.sh
robertkirkman, line 1: I see this:

- https://github.com/termux/termux-packages/pull/21976#issuecomment-2438755233

however, it has `TERMUX_PKG_EXCLUDED_ARCHES="arm"` for a long time, seeming to defeat the purpose.

Can it be disabled in TUR until the problem with building for 32-bit ARM is resolved?

If not, can it be renamed + marked as `TERMUX_PKG_CONFLICTS` (including dependencies) to `octave-x-gnu-abi`, or have `TERMUX_PKG_EXCLUDED_ARCHES="aarch64, x86_64"` added, so that it does not silently conflict with Octave in upstream termux-packages repository?

robertkirkman: I can see that some of the others, except for `virglrenderer-android`, do exist here for 32-bit because they don't have 32-bit in github.com/termux/termux-packages.

however, because of that, I'd like to disable the 64-bit builds of them here using `TERMUX_PKG_EXCLUDED_ARCHES="aarch64, x86_64"` so that the 64-bit builds can be removed from TUR, so that they stop conflicting with the 64-bit builds of the same packages in github.com/termux/termux-packages.

Would that be acceptable? I'll try to rebuild them for that now.

tur/python-scikit-image/build.sh
```diff
@@ -3,9 +3,12 @@ TERMUX_PKG_DESCRIPTION="Image processing in Python"
 TERMUX_PKG_LICENSE="BSD 2-Clause, BSD 3-Clause, MIT"
 TERMUX_PKG_MAINTAINER="@termux-user-repository"
 TERMUX_PKG_VERSION="0.26.0"
+TERMUX_PKG_REVISION=1
 TERMUX_PKG_SRCURL=https://github.com/scikit-image/scikit-image/archive/refs/tags/v$TERMUX_PKG_VERSION.tar.gz
 TERMUX_PKG_SHA256=09bb8cbb2a7b0c4f24fd4caeda6357732c64c2556520bdbad1dc15fb498e6a08
-TERMUX_PKG_AUTO_UPDATE=true
+# packages that do not target aarch64 for dependency infrastructure reasons
+# cannot auto update because of 'ERROR: Obtaining update order failed'
+TERMUX_PKG_AUTO_UPDATE=false
```
robertkirkman, line 11: I want to fix this by moving only the 64-bit builds of `python-scikit-image` to github.com/termux/termux-packages and then allowing them to auto update only there.

That way, the 64-bit builds of `python-scikit-image` will not have a chance of attempting to run while a _different_ build of `python-scipy` from the one they were built with is installed in the device.

Is that ok?

tur/octave-x/build.sh
licy183: > however, because of that, I'd like to disable the 64-bit builds of them here using `TERMUX_PKG_EXCLUDED_ARCHES="aarch64, x86_64"` so that the 64-bit builds can be removed from TUR, so that they stop conflicting with the 64-bit builds of the same packages in github.com/termux/termux-packages.
> 
> Would that be acceptable? I'll try to rebuild them for that now.

The reason I didn't do this before was that if the build.sh in the termux-packages had `TERMUX_PKG_EXCLUDED_ARCHES="arm, i686"`, while the build.sh in TUR had `TERMUX_PKG_EXCLUDED_ARCHES="aarch64, x86_64"`, then it would result in the inability to build arm and i686.

Has you confirmed that it has been fixed now?

robertkirkman:
Ok, that sounds good then. Thank you for explaining. I will merge the PR in the main repository and close this PR, since it is probably unnecessary after `octave`, `octave-x` and all duplicated dependencies are epoch-bumped to 2 in github.com/termux/termux-packages.

PR #2183, opened 2026-01-19
addpkg(tur-on-device/awscli): 2.32.28

Kapral67:
https://github.com/termux-user-repository/tur/pull/863

https://github.com/termux/termux-packages/pull/28092

### DIFF
--- a/tur-on-device/awscli/build.sh
+++ b/tur-on-device/awscli/build.sh
@@ -1,0 +1,27 @@
+TERMUX_PKG_HOMEPAGE=https://aws.amazon.com/cli
+TERMUX_PKG_DESCRIPTION="Universal Command Line Interface for Amazon Web Services"
+TERMUX_PKG_LICENSE="Apache-2.0"
+TERMUX_PKG_MAINTAINER="@termux-user-repository"
+TERMUX_PKG_VERSION="2.32.28"
+TERMUX_PKG_SRCURL="https://github.com/aws/aws-cli/archive/refs/tags/$TERMUX_PKG_VERSION.tar.gz"
+TERMUX_PKG_SHA256=85ecd5baf4d5cdf4d9e436dd76e023d0126cdcd0e5cbf97a5cd15b33f8829b05
+TERMUX_PKG_UPDATE_VERSION_REGEXP="\d+.\d+.\d+"
+TERMUX_PKG_AUTO_UPDATE=true
+TERMUX_PKG_SETUP_PYTHON=true
+TERMUX_PKG_DEPENDS="libffi, libsqlite, groff"
+TERMUX_PKG_BUILD_DEPENDS="ldd"
+TERMUX_PKG_EXTRA_CONFIGURE_ARGS="
+--with-install-type=portable-exe
+--with-download-deps
+"
+TERMUX_PKG_MAKE_PROCESSES=1
+
+termux_step_pre_configure() {
+	export LDFLAGS="$LDFLAGS -lm"
+}
+
+termux_step_make() {
+	export PIP_NO_BINARY='awscrt'
+	export AWS_CRT_BUILD_FORCE_STATIC_LIBS=1
+	make
+}


tur-on-device/awscli/build.sh
```diff
@@ -0,0 +1,25 @@
...
+TERMUX_PKG_DESCRIPTION="Universal Command Line Interface for Amazon Web Services"
+TERMUX_PKG_LICENSE="Apache-2.0"
+TERMUX_PKG_MAINTAINER="@termux-user-repository"
+TERMUX_PKG_VERSION="2.32.28"
+TERMUX_PKG_SRCURL=git+https://github.com/aws/aws-cli.git
+TERMUX_PKG_GIT_BRANCH="$TERMUX_PKG_VERSION"
+TERMUX_PKG_AUTO_UPDATE=true
+TERMUX_PKG_SETUP_PYTHON=true
+TERMUX_PKG_DEPENDS="libffi, libsqlite, groff"
+TERMUX_PKG_BUILD_DEPENDS="aosp-libs, python, python-pip, clang, make, cmake, binutils, ndk-sysroot, ldd"
```
robertkirkman: Great, 

In a TUR-on-device package, all of these `TERMUX_PKG_BUILD_DEPENDS` are not necessary, I think, except for maybe `python` and `python-pip`. I think you should try removing them

robertkirkman: Ok, `ldd` is necessary. Sorry, I didn't notice that one until you tried removing it. So, just add back `ldd` and try again, all the others should be automatically present.

tur-on-device/awscli/build.sh
```diff
@@ -0,0 +1,26 @@
+TERMUX_PKG_HOMEPAGE=https://aws.amazon.com/cli
+TERMUX_PKG_DESCRIPTION="Universal Command Line Interface for Amazon Web Services"
+TERMUX_PKG_LICENSE="Apache-2.0"
+TERMUX_PKG_MAINTAINER="@termux-user-repository"
+TERMUX_PKG_VERSION="2.32.28"
+TERMUX_PKG_SRCURL=git+https://github.com/aws/aws-cli.git
```
robertkirkman: What goes wrong if https://github.com/aws/aws-cli/archive/refs/tags/2.33.2.tar.gz is used instead of git+https://github.com/aws/aws-cli.git ? 

Does it attempt to run `git` commands on the cloned repository during the build and fail? There aren't any submodules, so it doesn't seem to need submodules.

Kapral67: I haven't tried that way before, is it preferred over git?
currently it uses the TERMUX_PKG_GIT_BRANCH to checkout the version tag

robertkirkman: It's preferred to try this if it works:

```
TERMUX_PKG_VERSION="2.32.28"
TERMUX_PKG_SRCURL="https://github.com/aws/aws-cli/archive/refs/tags/$TERMUX_PKG_VERSION.tar.gz"
TERMUX_PKG_SHA256=123456...
TERMUX_PKG_UPDATE_VERSION_REGEXP="\d+.\d+.\d+"
```

(and put the number in `TERMUX_PKG_SHA256` that it will say to after the first time using it with a random number like "1")

Kapral67: ok will try that

would it still need `TERMUX_PKG_AUTO_UPDATE=true`?

robertkirkman: Yes, keep `TERMUX_PKG_AUTO_UPDATE=true`

robertkirkman: TUR-on-device doesn't currently have working auto update, but it will still create a notification issue when there is an update, and TUR-on-device auto update implementation could be finished in the future

robertkirkman: It needs `TERMUX_PKG_SHA256` now, I edited my message to explain how to add that

robertkirkman (approved):

Kapral67:
#2187 for a cross compiled one that can be in tur instead

PR #2199, opened 2026-01-23
addpkg(tur/xbill): 2.1

IntinteDAO:


tur/xbill/02_hurd_logos.patch
```diff
@@ -0,0 +1,245 @@
+Description: Hurd logos for the game
+Author: Barry deFreese <bddebian@comcast.net>
```
robertkirkman, line 2: This is a Hurd-related patch, which is not relevant to Android

robertkirkman: Also, these patches are `.diff` files, which have no effect in Termux packages if not manually applied during `build.sh` (unlike `.patch` files)

IntinteDAO: > This is a Hurd-related patch, which is not relevant to Android

It's just a game content. In xbill we "kill" Bill Gates and Windows to protect Unix computers :P

IntinteDAO: > Also, these patches are `.diff` files, which have no effect in Termux packages if not manually applied during `build.sh` (unlike `.patch` files)

Damn, I just copied it without checking. I going to fix it

robertkirkman: > > This is a Hurd-related patch, which is not relevant to Android
> 
> It's just a game content. In xbill we "kill" Bill Gates and Windows to protect Unix computers :P

Wow, if you're right that's quite surprising and it's my mistake.

So, it's "Hurd in the game not Hurd in real life", a "meta" thing.

I had just checked it and thought "This is ASCII of the real Hurd logo, I assume this will not be displayed on Android" but I wasn't thinking with the correct "meta" logic.

robertkirkman:
It looks good, but could you squash the commits so that there is one commit?

PR #2207, opened 2026-01-25
addpkg(tur/freegish): 1.53-git

IntinteDAO:
Don't know what do to if the game is not released officially.


tur/freegish/build.sh
```diff
@@ -0,0 +1,10 @@
+TERMUX_PKG_HOMEPAGE=https://github.com/freegish/freegish
+TERMUX_PKG_DESCRIPTION="A physics based platformer"
+TERMUX_PKG_LICENSE="GPL-2.0"
+TERMUX_PKG_MAINTAINER="IntinteDAO"
+TERMUX_PKG_VERSION=1.53
+TERMUX_PKG_SRCURL=https://github.com/freegish/freegish/archive/refs/heads/master.zip
```
robertkirkman: I suggest this:

```
_COMMIT=f6a92933c9ba584214304b12405a6fd6ab0c3cdf
TERMUX_PKG_VERSION="1.53-p20240323"
TERMUX_PKG_SRCURL="https://github.com/freegish/freegish/archive/$_COMMIT.tar.gz"
```

this way, if the repository gets any updates, the version that last built successfully for Termux will be stored, and can be updated manually when ready.

tur/freegish/build.sh
```diff
@@ -0,0 +1,12 @@
...
+TERMUX_PKG_DESCRIPTION="A physics based platformer"
+TERMUX_PKG_LICENSE="GPL-2.0"
+TERMUX_PKG_MAINTAINER="IntinteDAO"
+_COMMIT=f6a92933c9ba584214304b12405a6fd6ab0c3cdf
+TERMUX_PKG_VERSION="1.53-p20240323"
+TERMUX_PKG_SRCURL="https://github.com/freegish/freegish/archive/$_COMMIT.tar.gz"
+TERMUX_PKG_SHA256=39f26294705e845e4c9bc0c6e55a11b93e9ec8ce984ec783835336f5f6679629
+TERMUX_PKG_DEPENDS="libvorbis, openal-soft, sdl2, libpng, zlib"
+TERMUX_PKG_EXTRA_CONFIGURE_ARGS="-DINSTALL_FHS=ON -DCMAKE_POLICY_VERSION_MINIMUM=3.5"
+TERMUX_PKG_AUTO_UPDATE=true
```
robertkirkman, line 11: It says that it needs `TERMUX_PKG_UPDATE_TAG_TYPE=newest-tag`

tur/freegish/build.sh
```diff
@@ -0,0 +1,13 @@
...
```
robertkirkman, line 11: Oh, I see the issue. You cannot auto update this package because it does not have ANY tags or releases in the upstream repository.

If you have a theory/idea of how you can use arbitrary code scripting to check for updates, you can implement your idea in a new `termux_pkg_auto_update()` function in `build.sh`, but generally that is hard and only possible if there is some way that upstream will actually tell us when they're going to release it.

PR #2213, opened 2026-01-26
addpkg(tur/numptyphysics): 0.3.10

IntinteDAO:


tur/numptyphysics/build.sh
```diff
@@ -0,0 +1,20 @@
+TERMUX_PKG_HOMEPAGE=https://gitlab.com/thp/numptyphysics
+TERMUX_PKG_DESCRIPTION="Crayon based physics puzzle game"
+TERMUX_PKG_LICENSE="GPL-3.0"
+TERMUX_PKG_MAINTAINER="@IntinteDAO"
+TERMUX_PKG_VERSION=0.3.10
+TERMUX_PKG_SRCURL=https://gitlab.com/thp/numptyphysics/-/archive/0.3.10/numptyphysics-0.3.10.tar.bz2
```
robertkirkman, line 6: It should probably have `$TERMUX_PKG_VERSION` used in `TERMUX_PKG_SRCURL` instead of the version number hardcoded.

tur/numptyphysics/usr-games.patch
```diff
@@ -0,0 +1,17 @@
...
+
+diff --git a/src/Os.cpp b/src/Os.cpp
+index a741110..6de71a7 100644
+--- a/src/Os.cpp
++++ b/src/Os.cpp
+@@ -290,6 +290,8 @@ Os::appName()
+ std::string
+ Os::globalDataDir()
+ {
++    return "@TERMUX_PREFIX@/share/games/numptyphysics/data/";
```
robertkirkman, line 13: This patch, and the `install-path.patch` above, are in my opinion both "Debianisms". I think that all they do is move the app data path around to fit a Debian policy that I don't think necessarily exists in Termux; at least in my opinion, allowing the game to use its default location (if that's `$TERMUX_PREFIX/share/numptyphysics/data`) is probably fine. 

I understand now that you did this to `xmoto`, but I admit I did not think about what you were doing very critically at that time, and I assumed that maybe you had changed the paths because the game otherwise failed to launch.

Does this game fail to work if you remove the patches that create the Debianism `games` folder?

robertkirkman: I understand that there are a couple of other packages in Termux that do already install a `games` folder, for example `xboard`, but in those cases that happens because the upstream code already contains references to a `games` folder, meaning that upstream intended that location relative to `$PREFIX` for their project.

In the case of `numptyphysics`, upstream shows that they intend the location relative to `$PREFIX` to be `share/numptyphysics`.

The reason I just feel concerned about allowing this in Termux is because:

Termux should, in my opinion, prioritize maintaining patches that are necessary for the functioning of the software on Android and also for the functioning of the software in harmony with the other packages in Termux, and should avoid exactly duplicating patches from other distros that only serve a purpose in the other distro and have no measurable benefits for Termux. This keeps the software behavior as close as possible to the intentions of upstream developers, while also reducing the amount of maintenance work required for bumping the packages in Termux.

PR #2248, opened 2026-02-12
[Experimental] 0AD

IntinteDAO:
@robertkirkman 

The code is compiled, but have a lots of Segmentation Fault. Gemini believed it's caused of MTE. I can try to fix all issues if you also confirm that, but maybe you have better idea


tur/0ad/build.sh
```diff
@@ -0,0 +1,113 @@
...
+	termux_setup_rust
+	termux_setup_nodejs
+	(
+		local PREMAKE_BASE=$TERMUX_PKG_SRCDIR/libraries/source/premake-core
+		local PV=5.0.0-beta8
+		mkdir -p $PREMAKE_BASE
+		cd $PREMAKE_BASE
+		[ ! -f "v${PV}.tar.gz" ] && curl -Lo "v${PV}.tar.gz" "https://github.com/premake/premake-core/archive/refs/tags/v${PV}.tar.gz"
+		tar -xf "v${PV}.tar.gz" --strip-components=1
+		env -i PATH="$PATH" HOME="$HOME" CC=gcc make -f Bootstrap.mak linux
```
robertkirkman: can this part be in `termux_step_host_build()` so it can be reused for repeated builds, or more convenient for it to be in `termux_step_pre_configure()`?

tur/0ad/patch-028.patch
```diff
@@ -0,0 +1,534 @@
...
+@@ -166,10 +166,10 @@
+ }
+ 
+ 
+-/*static*/ OsPath Paths::Root(const OsPath& argv0)
++/*static*/ OsPath Paths::Root([[maybe_unused]] const OsPath& argv0)
+ {
+ #if OS_ANDROID
+-	return OsPath("/sdcard/0ad"); // TODO: this is kind of bogus
++	return OsPath("/data/data/com.termux/files/usr/share/0ad/data"); // TODO: this is kind of bogus
```
robertkirkman: If you disable `OS_ANDROID` define at the first place it is defined, instead of editing things within `OS_ANDROID` blocks, what happens? Does it work that way or does a problem happen?

robertkirkman: I see you made a lot of edits that rely on `OS_ANDROID`,

I suggest you try making another version with `OS_ANDROID` globally disabled, and check what patches are necessary for that one. It might result in a version that requires smaller patches.

IntinteDAO: Well, fuck

I reworked all the “ANDROID” parts according to best practices, and the game works now. I'll try to clean up the entire docker and udocker environment, because maybe there's some cache...

IntinteDAO: @robertkirkman 
Well, not fully works, one bug occured, but maybe fixable somehow (with 0ad team?)

robertkirkman: If you show me the code and what bug happened then I can try to help with it when I have time

IntinteDAO: @robertkirkman 

https://github.com/IntinteDAO/tur/tree/0ad2

You need to build it ofc, I recommend
./scripts/run-docker.sh ./build-package.sh -I -d -f 0ad

Then you have 0ad with debug symbols

Run it via GDB, choose a game scenario and wait 10 seconds

```
#0  0x0000007fd638e388 in ?? ()
   from /data/data/com.termux/files/usr/lib/libmozjs-128.so
#1  0x0000007fd638ebdc in ?? ()
   from /data/data/com.termux/files/usr/lib/libmozjs-128.so
#2  0x0000007fd6421478 in JS_CallFunctionValue(JSContext*, JS::Handle<JSObject*>, JS::Handle<JS::Value>, JS::HandleValueArray const&, JS::MutableHandle<JS::Value>) () from /data/data/com.termux/files/usr/lib/libmozjs-128.so
#3  0x0000005555b8255c in IGUIObject::ScriptEventWithReturn (
    this=this@entry=0xb400007e33bcf640, eventName=..., paramData=...)
    at ../../../source/gui/ObjectBases/IGUIObject.cpp:481
#4  0x0000005555b822d8 in IGUIObject::ScriptEventWithReturn (
    this=0xb400007e33bcf640, eventName=...)
    at ../../../source/gui/ObjectBases/IGUIObject.cpp:462
#5  0x0000005555b35088 in CGUI::SendEventToAll (this=<optimized out>, 
    eventName=...) at ../../../source/gui/CGUI.cpp:362
#6  0x0000005555b30624 in CGUIManager::SendEventToAll (this=<optimized out>, 
    eventName=...) at ../../../source/gui/GUIManager.cpp:395
#7  0x00000055558cceb4 in CGame::Update (this=0xb400007da4a5ebc0, 
    deltaRealTime=<optimized out>, doInterpolate=true)
    at ../../../source/ps/Game.cpp:438
#8  0x0000005555725990 in Frame (fixedFrameFrequency=0, 
    rlInterface=<optimized out>, dapInterface=<optimized out>)
    at ../../../source/main.cpp:470
--Type <RET> for more, q to quit, c to continue without paging--c
#9  RunGameOrAtlas (argv=...) at ../../../source/main.cpp:760
#10 0x0000005555723854 in main (argc=1, argv=0x7fffffdf88)
    at ../../../source/main.cpp:834
```

IntinteDAO: > If you show me the code and what bug happened then I can try to help with it when I have time

[fix.patch](https://github.com/user-attachments/files/25459121/fix.patch)

This might be a fix for that, but need more tests

IntinteDAO: @robertkirkman 

I found another problem—the game crashes if you close it too quickly. It's not a critical error, but it might be worth looking into :D

It seems that 0ad version 0.28, even though it's rc5, has a lot of incompatibilities with SearchMonkey, even on Linux, and maybe that's what we're dealing with. That it's not a problem with Termux, libraries, Android, but with the game code itself.

IntinteDAO: @robertkirkman 
I talk with 0ad devs about it and the problem is that Searchmonkey required patches. 

I wonder whether it is better to create patches that solve some or all of the problems, or whether it is better to patch Searchmonkey.

robertkirkman: > @robertkirkman I talk with 0ad devs about it and the problem is that Searchmonkey required patches.
> 
> I wonder whether it is better to create patches that solve some or all of the problems, or whether it is better to patch Searchmonkey.

Could you send the link to "Searchmonkey"? I'm not 100% sure what it is

IntinteDAO: @robertkirkman 
Sorry, spidermonkey. I don't know who comes up with Mozilla's names, but they should fire him :P

In my spare time, I'll add patches to SpiderMonkey and see if they solve the problems.

IntinteDAO: @robertkirkman 
Ok, I just modify build.sh to build a Spidermonkey for 0ad (from 0ad package)
* Spidermonkey from Termux might be upgraded, 0AD required "definied" version, that's why it should be build with 0ad
* Spidermonkey (from 0ad) have some patches. We can backport to our version, but maybe those patches can be unstable (or something) with spidermonkey apps

In short - it works. Now I must add menu entries, analyze binary file from other distro, do clean-up and push here for verification and proposals

robertkirkman: > Spidermonkey (from 0ad) have some patches

I think those patches are for 0ad (0ad version of spidermonkey) only, right?

IntinteDAO: > > Spidermonkey (from 0ad) have some patches
> 
> I think those patches are for 0ad (0ad version of spidermonkey) only, right?

Yes. In theory, it seems that it will work fine for SpiderMonkey (it's not a big deal), but the question is whether we should implement such fixes when we can provide a special version of the SpiderMonkey library for 0AD.

In my opinion, it is better not to combine these libraries just in case. Especially since 0AD may run on a different version of SM than the one in the Termux repository.

robertkirkman: > Yes. In theory, it seems that it will work fine for SpiderMonkey (it's not a big deal), but the question is whether we should implement such fixes when we can provide a special version of the SpiderMonkey library for 0AD.

> In my opinion, it is better not to combine these libraries just in case. Especially since 0AD may run on a different version of SM than the one in the Termux repository.

oh, in that case, yes, I think the patches should be applied, and also I think that the 0AD version of spidermonkey should be inside the 0AD package, and also it should be the same exact version of spidermonkey matching the version that the 0AD developers assigned that the version of 0AD, preferably built from SpiderMonkey source code provided by the 0AD developers along with the 0AD source code

robertkirkman: (but also all Termux spidermonkey patches from Termux spidermonkey package also applied to it)

IntinteDAO: @robertkirkman Small cleanups created, but we can do more soon. This is pre-release version (without cleaning, but create menus, use spidermonkey from 0ad etc)

IntinteDAO:
@robertkirkman 
I cleaned it up as much as possible.

tur/0ad/build.sh
```diff
@@ -0,0 +1,133 @@
...
+TERMUX_PKG_BUILD_IN_SRC=true
+TERMUX_PKG_HOSTBUILD=true
+
+termux_step_host_build() {
+	local PREMAKE_SRC=$TERMUX_PKG_SRCDIR/libraries/source/premake-core
+	local PV=5.0.0-beta8
+	mkdir -p "$PREMAKE_SRC"
+	cd "$PREMAKE_SRC"
+	if [ ! -f "v${PV}.tar.gz" ]; then
+		curl -Lo "v${PV}.tar.gz" "https://github.com/premake/premake-core/archive/refs/tags/v${PV}.tar.gz"
```
robertkirkman: Can this use `termux_download`?

tur/0ad/build.sh
```diff
@@ -0,0 +1,133 @@
...
+
+termux_step_post_get_source() {
+	local FCOLLADA_BASE="libraries/source/fcollada"
+	mkdir -p "$FCOLLADA_BASE"
+	if [ ! -f "$FCOLLADA_BASE/fcollada-28209.tar.xz" ]; then
+		tar -xf "$TERMUX_PKG_SRCDIR/../0ad-${TERMUX_PKG_VERSION}-unix-build.tar.xz" -O "0ad-${TERMUX_PKG_VERSION}/libraries/source/fcollada/fcollada-28209.tar.xz" > "$FCOLLADA_BASE/fcollada-28209.tar.xz" || true
+	fi
+
+	# Download and Extract Data Tarball (Crucial for assets)
+	local DATA_URL="https://releases.wildfiregames.com/0ad-${TERMUX_PKG_VERSION}-unix-data.tar.xz"
```
robertkirkman: Could this be in `TERMUX_PKG_SRCURL` as the 2nd element in an array, because it is part of the game and matched with `TERMUX_PKG_VERSION`?

tur/0ad/build.sh
```diff
@@ -0,0 +1,133 @@
...
+	local DATA_FILE=$TERMUX_PKG_CACHEDIR/0ad-${TERMUX_PKG_VERSION}-unix-data.tar.xz
+	termux_download $DATA_URL $DATA_FILE e844b30ae2102c47e0a4fff2f0e0ef05ba0cebb1890aa72276fa12457c39526f
+
+	tar -xf "$DATA_FILE" -C "$TERMUX_PKG_SRCDIR" --strip-components=1
+
+	# SpiderMonkey patches from standalone package (only the ones that apply to the JS engine)
+	local SM_PATCH_DIR="$TERMUX_SCRIPTDIR/packages/spidermonkey"
+	mkdir -p libraries/source/spidermonkey/patches/termux
+	for p in 0001-configure-treat-termux-as-linux.patch 0002-configure-fix-arm-options.patch 0003-configure-fix-rustflags.patch 0004-configure-do-not-check-libstdcxx-headers.patch 0005-configure-do-not-append-sysroot.patch 0012-do-not-use-adaptive-mutex.patch 0014-unset-thread-priority-scheduling-macro.patch 0021-fix-missing-include-for-zlib-module.patch 0024-fix-missing-include-for-macro-MOZ_TRY-in-TimeZone.cpp.patch; do
```
robertkirkman: Could you implement this by copying and pasting all these patches to the `tur/tur/0ad` folder before committing the change, instead of copying them at build-time?

(I am concerned that if the spidermonkey version in `$TERMUX_SCRIPTDIR/packages/spidermonkey` changes and its patches change, but the spidermonkey version inside 0ad does not change, the 0ad package will stop building and have errors)

tur/0ad/build.sh
```diff
@@ -0,0 +1,133 @@
...
+	export PATH="$TERMUX_PKG_HOSTBUILD_DIR/bin:$PATH"
+	if [ ! -f "$TERMUX_PKG_HOSTBUILD_DIR/bin/premake5" ]; then
+		echo "ERROR: premake5 not found in $TERMUX_PKG_HOSTBUILD_DIR/bin"
+		ls -R "$TERMUX_PKG_HOSTBUILD_DIR"
+		exit 1
+	fi
+	cd $TERMUX_PKG_SRCDIR/libraries
+	./build-source-libs.sh --without-nvtt --with-system-premake -j${TERMUX_MAKE_PROCESSES:-4}
+	export PKG_CONFIG_PATH="$TERMUX_PREFIX/lib/pkgconfig:${PKG_CONFIG_PATH:-}"
+	export LDFLAGS="${LDFLAGS:-} -L$TERMUX_PREFIX/lib -liconv -llog -lgloox -Wl,--no-as-needed,-lOpenSLES,--as-needed -Wl,-rpath=$TERMUX_PREFIX/lib/0ad"
```
robertkirkman: It can be written as `LDFLAGS+="` instead of ` LDFLAGS="${LDFLAGS:-}` right?

tur/0ad/build.sh
```diff
@@ -0,0 +1,133 @@
...
+EOF
+
+	export PATH="$TERMUX_PKG_HOSTBUILD_DIR/bin:$PATH"
+	if [ ! -f "$TERMUX_PKG_HOSTBUILD_DIR/bin/premake5" ]; then
+		echo "ERROR: premake5 not found in $TERMUX_PKG_HOSTBUILD_DIR/bin"
+		ls -R "$TERMUX_PKG_HOSTBUILD_DIR"
+		exit 1
+	fi
+	cd $TERMUX_PKG_SRCDIR/libraries
+	./build-source-libs.sh --without-nvtt --with-system-premake -j${TERMUX_MAKE_PROCESSES:-4}
```
robertkirkman: Does not need `:-4` symbols here, I think

tur/0ad/build.sh
```diff
@@ -0,0 +1,133 @@
...
+	export PKG_CONFIG_PATH="$TERMUX_PREFIX/lib/pkgconfig:${PKG_CONFIG_PATH:-}"
```
robertkirkman: I think this line is unnecessary, right? If this line is necessary, what error happens if it is removed?

tur/0ad/build.sh
```diff
@@ -0,0 +1,133 @@
...
+	export LDFLAGS="${LDFLAGS:-} -L$TERMUX_PREFIX/lib -liconv -llog -lgloox -Wl,--no-as-needed,-lOpenSLES,--as-needed -Wl,-rpath=$TERMUX_PREFIX/lib/0ad"
+
+	export CXXFLAGS=$(echo "$CXXFLAGS" | sed 's/-D_FORTIFY_SOURCE=2//g; s/-D__USE_FORTIFY_LEVEL=2//g')
+	export CFLAGS=$(echo "$CFLAGS" | sed 's/-D_FORTIFY_SOURCE=2//g; s/-D__USE_FORTIFY_LEVEL=2//g')
+
+	export CXXFLAGS="${CXXFLAGS:-} -D_GNU_SOURCE -DCONFIG_ENABLE_MOCKS=0"
+	export CXXFLAGS="$CXXFLAGS -U__ANDROID__ -D_LIBCPP_HAS_NO_C11_ALIGNED_ALLOC -DINSTALLED_DATADIR=\"$TERMUX_PREFIX/share/0ad\""
```
robertkirkman: These can be written as `CXXFLAGS+="` and combined into one `export` across multiple lines (I think)

tur/0ad/build.sh
```diff
@@ -0,0 +1,133 @@
...
+	cd $TERMUX_PKG_SRCDIR/build/workspaces
+	./update-workspaces.sh --with-system-premake5 --without-nvtt --without-atlas --without-tests -j${TERMUX_MAKE_PROCESSES:-4}
```
robertkirkman: Does not need `:-4` symbols here, I think

tur/0ad/build.sh
```diff
@@ -0,0 +1,133 @@
...
+	make -j${TERMUX_MAKE_PROCESSES:-4} pyrogenesis CC="$CC" CXX="$CXX" LD="$CC" debug=1 LDFLAGS="${LDFLAGS}"
+}
+
+termux_step_make_install() {
+	mkdir -p $TERMUX_PREFIX/bin
+	cp $TERMUX_PKG_SRCDIR/binaries/system/pyrogenesis $TERMUX_PREFIX/bin/pyrogenesis
+
+	# Isolate 0AD shared libraries into their own directory
+	mkdir -p $TERMUX_PREFIX/lib/0ad
+	[ -f "$TERMUX_PKG_SRCDIR/binaries/system/libCollada.so" ] && install -Dm755 $TERMUX_PKG_SRCDIR/binaries/system/libCollada.so $TERMUX_PREFIX/lib/0ad/libCollada.so
```
robertkirkman, line 116: Why do you write `[ -f` for these?

Shouldn't it be written as just `install` without `[ -f` because, if the file isn't there, that's bad and it should make an error?

tur/0ad/build.sh
```diff
@@ -0,0 +1,133 @@
...
+	[ -f "$TERMUX_PKG_SRCDIR/binaries/system/libmozjs128-release.so" ] && install -Dm755 $TERMUX_PKG_SRCDIR/binaries/system/libmozjs128-release.so $TERMUX_PREFIX/lib/0ad/libmozjs128-release.so
+
+	# Install desktop resources
+	install -Dm644 $TERMUX_PKG_SRCDIR/build/resources/0ad.desktop $TERMUX_PREFIX/share/applications/0ad.desktop
+	install -Dm644 $TERMUX_PKG_SRCDIR/build/resources/0ad.png $TERMUX_PREFIX/share/icons/hicolor/128x128/apps/0ad.png
+	install -Dm644 $TERMUX_PKG_SRCDIR/build/resources/0ad.appdata.xml $TERMUX_PREFIX/share/metainfo/0ad.appdata.xml
+	install -Dm644 $TERMUX_PKG_SRCDIR/build/resources/pyrogenesis.xml $TERMUX_PREFIX/share/mime/packages/pyrogenesis.xml
+
+	# Install game data (contents of binaries/data into share/0ad)
+	mkdir -p $TERMUX_PREFIX/share/0ad
```
robertkirkman: Should have `rm -rf "$TERMUX_PREFIX/share/0ad"` right before this line, to avoid an error if the build is built twice in the same container without deleting the container, right?

tur/0ad/build.sh
```diff
@@ -0,0 +1,133 @@
...
+	cp -r $TERMUX_PKG_SRCDIR/binaries/data/. $TERMUX_PREFIX/share/0ad/
```
robertkirkman: It is strange to write `/.` at the end of the path, should it be `/*`?

tur/0ad/build.sh
```diff
@@ -0,0 +1,133 @@
...
+
+	install -Dm755 $TERMUX_PKG_SRCDIR/build/resources/0ad.sh $TERMUX_PREFIX/bin/0ad
+	#$TERMUX_PREFIX/bin/termux-fix-shebang $TERMUX_PREFIX/bin/0ad
```
robertkirkman: This comment is left over and can be removed, right?

tur/0ad/disable_android_detection.patch
```diff
@@ -0,0 +1,293 @@
...
+-#if OS_MACOSX || OS_ANDROID
++#if OS_MACOSX || 0
+ static ErrorReactionInternal try_gui_display_error(const wchar_t* /*text*/, bool /*manual_break*/,
+ 	bool /*allow_suppress*/, bool /*no_continue*/)
+ {
+@@ -224,7 +225,7 @@
+ 	if (ret != ERI_NOT_IMPLEMENTED)
+ 		return ret;
+ 
+-#if OS_ANDROID
```
robertkirkman, line 126: You disabled `OS_ANDROID` globally now, so any patch hunk that had replaced `if OS_ANDROID` with `if 0` can be removed now, right?

tur/0ad/build.sh
```diff
@@ -0,0 +1,133 @@
...
+	./update-workspaces.sh --with-system-premake5 --without-nvtt --without-atlas --without-tests -j${TERMUX_MAKE_PROCESSES:-4}
+}
+
+termux_step_configure() {
+	:
+}
+
+termux_step_make() {
+	cd $TERMUX_PKG_SRCDIR/build/workspaces/gcc
+	make -j${TERMUX_MAKE_PROCESSES:-4} pyrogenesis CC="$CC" CXX="$CXX" LD="$CC" debug=1 LDFLAGS="${LDFLAGS}"
```
robertkirkman: Does not need `:-4` symbols here, I think

tur/0ad/build.sh
```diff
@@ -0,0 +1,133 @@
...
+	termux_setup_rust
+
+	cargo install cbindgen
+
+	if [ "$TERMUX_DEBUG_BUILD" = false ]; then
+		local env_host=$(printf $CARGO_TARGET_NAME | tr a-z A-Z | sed s/-/_/g)
+		export CARGO_TARGET_${env_host}_RUSTFLAGS+=" -C debuginfo=1"
+	fi
+
+	export BINDGEN_CFLAGS="--target=$CCTERMUX_HOST_PLATFORM --sysroot=$TERMUX_STANDALONE_TOOLCHAIN/sysroot -D_FORTIFY_SOURCE=0"
```
robertkirkman, line 69: `BINDGEN_CFLAGS="--target=` for 32-bit arm should be `armv7a-linux-androideabi$TERMUX_PKG_API_LEVEL` rather than `$CCTERMUX_HOST_PLATFORM`, you can look at a package like `mesa` in the main repository to see how that usually is

tur/0ad/build.sh
```diff
@@ -0,0 +1,133 @@
...
+	tar -xf "$DATA_FILE" -C "$TERMUX_PKG_SRCDIR" --strip-components=1
+
+	# SpiderMonkey patches from standalone package (only the ones that apply to the JS engine)
+	local SM_PATCH_DIR="$TERMUX_SCRIPTDIR/packages/spidermonkey"
+	mkdir -p libraries/source/spidermonkey/patches/termux
+	for p in 0001-configure-treat-termux-as-linux.patch 0002-configure-fix-arm-options.patch 0003-configure-fix-rustflags.patch 0004-configure-do-not-check-libstdcxx-headers.patch 0005-configure-do-not-append-sysroot.patch 0012-do-not-use-adaptive-mutex.patch 0014-unset-thread-priority-scheduling-macro.patch 0021-fix-missing-include-for-zlib-module.patch 0024-fix-missing-include-for-macro-MOZ_TRY-in-TimeZone.cpp.patch; do
+		[ -f "$SM_PATCH_DIR/$p" ] && cp "$SM_PATCH_DIR/$p" libraries/source/spidermonkey/patches/termux/ || true
+	done
+
+	cp "$TERMUX_PKG_BUILDER_DIR/fcollada-build.sh" "$FCOLLADA_BASE/build.sh"
```
robertkirkman: Could this be a `cat > "$FCOLLADA_BASE/build.sh" << 'EOF'` Heredoc in the `tur/0ad/build.sh` instead of a separate file? I think it is easier to see what is happening like that because the script is small.

robertkirkman:
Sorry for making the biggest review I have ever written on this, it is just that there are some parts that seemed to me like they could be more cleaned here so I just needed to ask about them?

PR #2250, opened 2026-02-12
codex: isolate target CFLAGS for cargo cross-builds

MatiasHiltunen:
Codex build fails in CI, this should fix it.

## Summary
- preserve target-specific CFLAGS for Cargo cross-compiles
- unset global CFLAGS before `cargo build` so host-side C compilation (e.g. `ring` build script) does not receive ARM-only flags

## Validation
- Docker build for all targets succeeded locally

### DIFF
--- a/tur/codex/build.sh
+++ b/tur/codex/build.sh
@@ -4,12 +4,34 @@ TERMUX_PKG_LICENSE="Apache-2.0, MIT"
 TERMUX_PKG_LICENSE_FILE="../LICENSE, ../NOTICE"
 TERMUX_PKG_MAINTAINER="@termux-user-repository"
 TERMUX_PKG_VERSION="0.104.0"
+TERMUX_PKG_REVISION=1
 TERMUX_PKG_SRCURL="https://github.com/openai/codex/archive/refs/tags/rust-v$TERMUX_PKG_VERSION.tar.gz"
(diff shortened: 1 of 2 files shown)


tur/codex/build.sh
```diff
@@ -18,6 +18,12 @@ termux_step_configure() {
 
 termux_step_make() {
 	termux_setup_rust
+	# Keep target CFLAGS for cross-compiles, but clear global CFLAGS so host-side
+	# C builds (e.g. ring build script) don't receive ARM-only flags.
+	local target_cflags_var=CFLAGS_${CARGO_TARGET_NAME//-/_}
+	printf -v "$target_cflags_var" '%s' "$CFLAGS"
+	export "$target_cflags_var"
+	unset CFLAGS
```
robertkirkman: rather than `unset CFLAGS`, to solve the error (`error: unrecognized command-line option '-mfloat-abi=softfp'`), licy183 prefers an alternative solution called "[`rust-cc-do-not-concatenate-all-the-CFLAGS.diff`](https://github.com/termux/termux-packages/blob/e42eda78737841aa8e9cfba973c799adf79fc71b/packages/pyrefly/rust-cc-do-not-concatenate-all-the-CFLAGS.diff)".

Could you try applying that here instead?

MatiasHiltunen: Thanks, I'll check that

PR #2294, opened 2026-03-03
newpkg: 0ad

IntinteDAO:
@robertkirkman I something fucked up - sorry :P

https://github.com/termux-user-repository/tur/pull/2248

For now, I'm thinking about those spidermonkey patches. Should they be in a separate directory or at the root of the project? I made a separate directory to keep things cleaner, but I don't know which is preferable.

PS. This is a bug in TUR / Termux Build, right? :P


robertkirkman:
> PS. This is a bug in TUR / Termux Build, right? :P

termux-packages is currently paused while thunder-coding does something. Wait for this commit to be reverted https://github.com/termux/termux-packages/commit/ac417206a86a7259afe50c8159100fec3a738e80

when that commit is reverted, it will signal that termux-packages is ready to continue development

PR #2296, opened 2026-03-04
chore(common-files/building-system-patches): remove `do-not-use-python-from-usr-bin.patch`

robertkirkman:
- After https://github.com/termux/termux-packages/pull/27739


NoNameWasDefined:
> the fuse thing is not working in TUR, but it has been speculated that the fuse thing might be removed from the main repository soon also, not sure yet, i'm going to wait to see what happens in the main repository.

Please! It doesn't works in rootless Podman even on upstream termux/termux-packages repo

PR #2316, opened 2026-03-08
revbump(tur-on-device/blender{4,5}): for `python` 3.13

robertkirkman:
- Also all other tur-on-device packages that need to be rebuilt for Python 3.13 except for Blender 3.6, which is too much maintenance work to work with Python 3.13 quickly

### DIFF
--- a/.github/workflows/package_on_device.yml
+++ b/.github/workflows/package_on_device.yml
@@ -1,6 +1,6 @@
-# Based on https://github.com/termux/termux-packages/blob/595969f581655e8cbf65182cf84bd5ffb2cf7b89/.github/workflows/packages.yml
+# Based on https://github.com/termux/termux-packages/blob/d20a0284b75b8b580ebf4e6b7b46c229748d989e/.github/workflows/packages.yml
 ##
-##  Copyright 2020 Termux
+##  Copyright 2026 Termux
 ##
 ##  Licensed under the Apache License, Version 2.0 (the "License");
 ##  you may not use this file except in compliance with the License.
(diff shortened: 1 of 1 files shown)


tur-disabled/blender/0016-python-3.13-android-as-linux.patch
```diff
@@ -0,0 +1,13 @@
+diff --git a/build_files/utils/make_update.py b/build_files/utils/make_update.py
+index 3129ade..5041b1e 100755
+--- a/build_files/utils/make_update.py
++++ b/build_files/utils/make_update.py
+@@ -71,6 +71,8 @@ def get_effective_platform(args: argparse.Namespace) -> str:
+         platform = "macos"
+     elif sys.platform == "win32":
+         platform = "windows"
++    elif sys.platform == "android":
++        platform = "linux"
```
robertkirkman, line 10: I include some changes that would be necessary to start building Blender 3.6 with Python 3.13, but Blender 3.6 is end-of-life, and officially does not support Python 3.13 and takes a lot of maintenance work to make it work with Python 3.13, so I probably will not make any more updates for it unless someone asks for updates for Blender 3.6.

robertkirkman:
> Hello! Sir! Thank you for your continued efforts in developing the Termux package!
> 
> Are there any plans for a `Blender 5.0` update? I heard that many of the Vulkan backend improvements were incorporated into version 5.0!
> 
> Currently, on `Turnip`, the Vulkan backend in `Blender 4.5` doesn't seem to work well with `A7xx/8xx`.

There is already blender 5.0, you can get it from `pkg install blender5`.. did you try it yet?

It's currently rebuilt for Python 3.13 in this PR.

robertkirkman:
@licy183 unfortunately I broke it by editing packages in `tur` and `tur-on-device` folders in a single PR, but I will try to fix it using this:

- https://github.com/termux-user-repository/dists/pull/5

robertkirkman:
I did it!!

Now TUR supports editing packages in the `tur-on-device` folder and the `tur` folder at the same time in a single PR!

image

PR #2329, opened 2026-03-12
bump(tur/rustc-nightly): 1.96.0-2026.03.12-nightly

robertkirkman:
- Fixes https://github.com/termux-user-repository/tur/issues/2328

- Copy `force-allow-edit-vendor.diff` from https://github.com/termux/termux-packages

- Rebase `cross-compile-error_index_generator.patch`

### DIFF
--- a/tur/rustc-nightly/build.sh
+++ b/tur/rustc-nightly/build.sh
@@ -2,12 +2,11 @@ TERMUX_PKG_HOMEPAGE=https://www.rust-lang.org/
 TERMUX_PKG_DESCRIPTION="Rust compiler and utilities (nightly version)"
 TERMUX_PKG_LICENSE="MIT"
 TERMUX_PKG_MAINTAINER="@termux-user-repository"
-TERMUX_PKG_VERSION="1.96.0-2026.03.11-nightly"
+TERMUX_PKG_VERSION="1.96.0-2026.03.12-nightly"
 _RUST_VERSION=$(echo $TERMUX_PKG_VERSION | cut -d- -f1)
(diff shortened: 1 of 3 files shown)


tur/rustc-nightly/build.sh
```diff
@@ -108,14 +122,6 @@ termux_step_pre_configure() {
 }
 
 termux_step_configure() {
-	# Install llvm-20
```
licy183, line 111: There is a historical reason for this. The LLVM version used by `rustc-nightly` is often newer than the version in the Termux-Package docker images, therefore there is no need to delete these codes as it may be useful in the future.

tur/rustc-nightly/bootstrap.toml
```diff
@@ -48,7 +48,7 @@ lld = false
 codegen-backends = ["llvm", "cranelift"]
 
 [target.x86_64-unknown-linux-gnu]
-llvm-config = "/usr/bin/llvm-config-20"
+llvm-config = "@TERMUX_HOST_LLVM_BASE_DIR@/bin/llvm-config"
```
licy183: Same here, `@TERMUX_HOST_LLVM_BASE_DIR@` should be `@TUR_RUSTC_NIGHTLY_LLVM_BASE_DIR@` or something else.

tur/rustc-nightly/build.sh
robertkirkman: hmm, I did not know that it was necessary, I'm sorry. I thought that it was just outdated because the stable rust package changed the way it detects LLVM. I will remove this change

PR #2361, opened 2026-03-21
bump(tur-on-device/blender{4,5}): 4.5.8, 5.1.0

robertkirkman:
- `blender5` now has a build dependency on `eigen5`


tur-on-device/blender5/build.sh
```diff
@@ -4,13 +4,12 @@ TERMUX_PKG_DESCRIPTION="A fully integrated 3D graphics creation suite"
...
-TERMUX_PKG_BUILD_DEPENDS="boost-headers, git-lfs, mold, sse2neon"
+TERMUX_PKG_DEPENDS="alembic, boost, brotli, desktop-file-utils, draco, ffmpeg, fftw, freetype, glew, hicolor-icon-theme, imath, libandroid-execinfo, libandroid-posix-semaphore, libblosc, libc++, libceres-solver, libepoxy, libharu, libpng, libpugixml, libraw, libsndfile, libspnav, libtbb, libtiff, libwebp, libx11, libxfixes, libxi, libxkbcommon, libyaml-cpp, oidn, openal-soft, opencolorio, openexr, openimageio, openjpeg, openpgl, openshadinglanguage, opensubdiv, openvdb, openxr, potrace, ptex, python, python-numpy, python-pip, shaderc, shared-mime-info, usd, zlib, zstd"
+TERMUX_PKG_BUILD_DEPENDS="boost-headers, eigen5, git-lfs, mold, sse2neon"
```
robertkirkman, line 12: So, since this is pretty confusing now, to be clear I should write this down:

`blender5` now has a build dependency on `eigen5`, and also a runtime dependency on `libceres-solver`, but `libceres-solver` also has a dependency on `eigen`, which is not API-compatible with `eigen5`, so `eigen5` and `eigen` must definitely both be installed at the same time during the build of `blender5`, but `eigen5` still has folders in it named "`eigen3`" (yes), so the best thing to do is to provide `eigen5` and install it into `$TERMUX_PREFIX/opt/eigen5` instead of `$TERMUX_PREFIX` so that it does not conflict with `eigen` at all.

robertkirkman: Termux will now have Eigen packages like Arch Linux:

|Eigen|Arch Linux|Termux|
|:--- |:--- |:--- |
|Version 5|`eigen`|`eigen5`|
|Version 3|`eigen3`|`eigen`|
|Conflicting files?|Yes|No|
|Primary version|5|3|

PR #1599, opened 2025-03-15
fix(tur/code-server): mark as unsafe for on-device builds

robertkirkman:
- Fixes #1598

PR #2386, opened 2026-03-30
fix(tur): some cleanup of python and rust packages

robertkirkman:
- Implement on-device build of all python packages except for `python-scipy`

- Build all rust packages with stable toolchain except for `rustc-nightly`

- Apply x11 support patch to all known uses of `arboard` in rust packages

- Version `Cython` dependency in `python-pandas` to avoid errors if a previously-built package required an old version of `Cython`

- Revision-bump all reverse dependencies of `python` that have not yet been rebuilt after the upgrade of `python` to 3.13


tur/instfd/build.sh
```diff
@@ -3,7 +3,7 @@ TERMUX_PKG_DESCRIPTION="FreeDOS installed for dosemu2"
 TERMUX_PKG_LICENSE="GPL-3.0"
 TERMUX_PKG_MAINTAINER="@stsp"
 TERMUX_PKG_VERSION="0.4"
-TERMUX_PKG_REVISION=1
+TERMUX_PKG_REVISION=2
```
licy183: It is just a python script. There is no need to revbump it.

tur/python-tokenizers/build.sh
```diff
@@ -3,74 +3,78 @@ TERMUX_PKG_DESCRIPTION="Fast State-of-the-Art Tokenizers optimized for Research
...
+	local TERMUX_RUST_VERSION="$(. "${TERMUX_SCRIPTDIR}"/packages/rust/build.sh; echo "${TERMUX_PKG_VERSION}")"
+	curl https://sh.rustup.rs -sSfo "${TERMUX_PKG_TMPDIR}"/rustup.sh
+	sh "${TERMUX_PKG_TMPDIR}"/rustup.sh -y
+	export PATH="${HOME}/.cargo/bin:${PATH}"
+	rustup target add "${CARGO_TARGET_NAME}"
+}
 
-	termux_setup_rust
+termux_step_pre_configure() {
+	# prevents:
```
licy183, line 30: I don't understand this change. In most of the circumstances, the `nightly` channel of rust should be used to provide `std` and `core`.

robertkirkman: This PR is WIP and might not be finished yet, but when it is finished, I wanted to ask several questions about certain things here.

One of the questions I was going to ask is: I have noticed that using this code is working as an alternative to using `rustup toolchain install nightly`:

```bash
	: "${CARGO_HOME:="$TERMUX_PKG_TMPDIR/.cargo"}"
	: "${RUSTUP_HOME:="$TERMUX_PKG_TMPDIR/.rustup"}"
	export CARGO_HOME RUSTUP_HOME
	local TERMUX_RUST_VERSION="$(. "${TERMUX_SCRIPTDIR}"/packages/rust/build.sh; echo "${TERMUX_PKG_VERSION}")"
	curl https://sh.rustup.rs -sSfo "${TERMUX_PKG_TMPDIR}"/rustup.sh
	sh "${TERMUX_PKG_TMPDIR}"/rustup.sh -y
	export PATH="${CARGO_HOME}/bin:${PATH}"
	rustup target add "${CARGO_TARGET_NAME}"
```

if the correct thing to do is to use `rustup toolchain install nightly`, then why does that work here? I wanted to understand that also.

tur/python-scikit-image/build.sh
```diff
@@ -20,10 +20,6 @@ TERMUX_PKG_EXTRA_CONFIGURE_ARGS="
 "
 
 termux_step_pre_configure() {
-	if $TERMUX_ON_DEVICE_BUILD; then
-		termux_error_exit "Package '$TERMUX_PKG_NAME' is not available for on-device builds."
```
robertkirkman, line 24: In order to achieve my plan to implement as many on-device builds in packages as possible (except for packages where something is not compatible with on-device build that I don't know how to fix) that I outlined here:

- https://github.com/termux-user-repository/tur/pull/1599#issuecomment-2726971933

I am starting to remove more of these checks if I tested the build while `$TERMUX_ON_DEVICE_BUILD` is `true` and I couldn't find anything bad happening,

do you know if there was an important reason these two packages have the on-device build disabled?

licy183: Previously, `meson-build` could not work correctly on device. It may work now.

tur/python-tokenizers/build.sh
```diff
@@ -3,74 +3,80 @@ TERMUX_PKG_DESCRIPTION="Fast State-of-the-Art Tokenizers optimized for Research
...
 	esac
+	local native_wheel_ext="${TERMUX_PKG_VERSION}-cp39-abi3-android_${ANDROID_API_LEVEL}_${native_wheel_arch}.whl"
 
-	# Fix wheel name, although it it built with tag `cp39-abi3`, but it is linked against `python3.x.so`
-	# so it will not work on other pythons.
-	mv "target/wheels/tokenizers-${TERMUX_PKG_VERSION}-${_tag}-android_${TERMUX_PKG_API_LEVEL}_${wheel_arch}.whl" \
-		"target/wheels/tokenizers-${TERMUX_PKG_VERSION}-py${_pyver}-none-any.whl"
-
-	pip install --force-reinstall --no-deps ./target/wheels/*.whl --prefix $TERMUX_PREFIX
+	# Avoid 'ERROR: polars_runtime_32-1.38.1-cp310-abi3-android_24_arm64_v8a.whl is not a supported wheel on this platform.'
```
robertkirkman: I did not understand the original comment, so I changed the comment to the error message that I see that the command actually fixes. However, I am not sure if changing that comment is correct.

At this point, the behavior of Python when it detects Android has changed greatly since the time when you wrote that comment. I was wondering, is the comment you wrote still accurate? also, if it is, could you explain what it means a little bit?

tur/python-scikit-image/build.sh
robertkirkman: > Previously, `meson-build` could not work correctly on device. It may work now.

For me, I think it is working fine. If it is still not working for you while using this PR version of it, I would be interested to find out what problem happens, and what kind of device you have.

tur/python-tokenizers/build.sh
robertkirkman: well, I know that it is linked to `python3.x.so`, so it does not work with other pythons, but what I don't understand is how that is connected to the problem of the `not a supported wheel on this platform` error that keeps happening and has to be avoided by generating a working filename and renaming the wheel file to that.

robertkirkman: When I think about it, this problem seems like something that might have a root cause that should be reported somewhere as an upstream issue, but I'm not sure where that would be, so I wanted to find out more.

licy183: Just ignore it, as it's a comment that wasn't deleted when it was copied from TUR's PyPI builder.

I suspect the reason you're asking is because crossenv changes the platform and arch.

tur/python-scikit-image/build.sh
licy183: This was so long ago that I can't remember exactly, but I vaguely recall it being a problem related to meson-python.

tur/python-tokenizers/build.sh
robertkirkman: Ok, I have left the comment unchanged,

if that is not what you meant, please let me know, since I am not sure if I understand.

tur/python-tokenizers/build.sh
licy183: I'll take a look at this over the weekend; I don't have much time to go through it right now.

PR #2391, opened 2026-03-31
Cork: new package

sapirrior:
# C.O.R.K. (Core Operations & Runtime Kernel)
### Info
Cork is a minimalist, language-agnostic task orchestrator and command runner developed in C, designed as a **lightweight, predictable** alternative to *complex* build tools like make or cmake, specifically optimized for resource-constrained environments like Termux.

### Extensible Shell Integration
Unlike traditional build systems that are strictly limited to compilation, C.O.R.K. serves as a universal wrapper for any system-level operations.

- Native Shell Support: Can execute any command available in the user's $PATH (e.g., pkg, git, python, curl).
- ​Environment Agnostic: Optimized to orchestrate complex workflows—from deploying AI agents to managing micro-services—all within a single Corkfile.
- ​Zero Abstraction: Commands are passed directly to the system shell, ensuring that there is no "middleware" delay or proprietary syntax to learn.


sapirrior:
Sir/man please review my package.

Issue #2394, opened 2026-03-31
[Bug]: frustrating AppArmor-related error `user builder is currently used by process 1`


robertkirkman:
### Problem description

Unfortunately, I have been experiencing an error that I am not sure how to reproduce, but is difficult to bypass once it starts happening:

```
tacokoneko@CORSAIR ~/code/termux/tur $ scripts/run-docker.sh ./build-package.sh -I -f gcc-15
Running container 'tur-termux-package-builder' from image 'ghcr.io/termux/package-builder'...
WARNING: apparmor_parser not found, AppArmor profiles will not be loaded!
         This is not recommended, as it may cause security issues and unexpected behavior
         Avoid executing untrusted code in the container
Creating new container...
df1a8650290083aa0e9295989b9484a84faba0c743211ecd0cdec061922811d7
Changed builder uid/gid... (this may take a while)
usermod: user builder is currently used by process 1
tacokoneko@CORSAIR ~/code/termux/tur $ 
```

It bisects to this, so I would like to apply a more complete condensed reversion of this PR that I have developed and tested:

- https://github.com/termux/termux-packages/pull/28602

This is also related to this:

- https://github.com/termux-user-repository/tur/pull/2296

Unfortunately, I did not completely finish that PR before it was merged, so I will work on a more complete version now.

This is also related to this:

- https://github.com/termux/termux-packages/issues/29118

Deleting the container and **even deleting all containers and images and trying again** does not help the problem once it has started, nor does restarting the computer:

```
docker container kill termux-package-builder
docker container rm termux-package-builder
# remove any other containers also using the termux/package-builder image
docker container rm other-termux-package-builder other2-termux-package-builder
for c in $(docker container ls -a --format "table {{.Names}}" | grep termux-package-builder); do docker container rm $c; done
docker image prune -a
# like sudo systemctl restart docker
sudo rc-service docker restart
sudo reboot
```

There may be an unknown factor required to reproduce it, but my patch can prevent it and it also can fix it once it has started happening

https://github.com/robertkirkman/termux-generator/blob/54be8e4f18c5ed4eb82cb9a9e7fe09db92b953de/play-store-patches/bootstrap-patches/docker-fixes.patch

### What steps will reproduce the bug?

1. Build some TUR packages using regular commands (this is the command I was using for a while before the problem happened):

```bash
git clone https://github.com/termux-user-repository/tur.git
cd tur
./setup-environment.sh
[...]

PR #2395, opened 2026-04-01
fix(common-files/building-system-patches/disable-apparmor): more aggressive method of disabling AppArmor

robertkirkman:
- Fixes https://github.com/termux-user-repository/tur/issues/2394 for me

- Related (though `termux-am` is not in TUR, the same patch also currently allows building `termux-am`) https://github.com/termux/termux-packages/issues/29118

### DIFF
--- a/common-files/building-system-patches/0008-disable-apparmor.patch
+++ b/common-files/building-system-patches/0008-disable-apparmor.patch
@@ -1,19 +1,48 @@
+--- a/scripts/build/toolchain/termux_setup_toolchain_29.sh
++++ b/scripts/build/toolchain/termux_setup_toolchain_29.sh
+@@ -133,7 +133,7 @@ termux_setup_toolchain_29() {
+ 	[ -d "${TERMUX_STANDALONE_TOOLCHAIN}-work" ] || mkdir -p "${TERMUX_STANDALONE_TOOLCHAIN}-work"
+ 
+ 
+-	if ! mountpoint -q "${TERMUX_STANDALONE_TOOLCHAIN}"; then
++	if false; then
+ 		fuse-overlayfs \
+ 			"${TERMUX_STANDALONE_TOOLCHAIN}" \
+ 			-o lowerdir="${NDK}/toolchains/llvm/prebuilt/linux-x86_64" \
+@@ -145,12 +145,17 @@ termux_setup_toolchain_29() {
+ 		return
+ 	fi
+ 
++	rm -rf "${TERMUX_STANDALONE_TOOLCHAIN}"
++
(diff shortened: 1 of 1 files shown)


common-files/building-system-patches/0008-disable-apparmor.patch
```diff
@@ -1,19 +1,48 @@
+--- a/scripts/build/toolchain/termux_setup_toolchain_29.sh
++++ b/scripts/build/toolchain/termux_setup_toolchain_29.sh
+@@ -133,7 +133,7 @@ termux_setup_toolchain_29() {
+ 	[ -d "${TERMUX_STANDALONE_TOOLCHAIN}-work" ] || mkdir -p "${TERMUX_STANDALONE_TOOLCHAIN}-work"
+ 
+ 
+-	if ! mountpoint -q "${TERMUX_STANDALONE_TOOLCHAIN}"; then
```
licy183, line 7: This PR ensures that TUR entirely doesn't depend on FUSE. I think a more robust approach would be to unmount the filesystem first, to check if it is currently mounted.

Currently, I use a virtual machine to build packages. There will be some problems if I attempt to build a package using TUR immediately after having previously built a package using `termux-packages`.

robertkirkman: > There will be some problems if I attempt to build a package using TUR immediately after having previously built a package using termux-packages.

Thank you for explaining that.

in my workflow, problems do not occur when I build a TUR package immediately after building a termux-packages package. That might be because of my practice of invariably using a unique `CONTAINER_NAME` with every different `scripts` folder that I manually set every time I download a new copy of the `scripts` folder. The reason I do that is because I consistently observed many problems occurring in the past as a result of attempting to share the same container for multiple separate `scripts` folders, that do not only have to do with AppArmor or fuse-overlayfs. However, I understand that not everyone finds it convenient to always use a different `CONTAINER_NAME` with every copy of the `scripts` folder.

In my opinion, this patch to disable both AppArmor and fuse-overlayfs should be a command line argument to the `run-docker.sh` script (and `build-package.sh` script) in upstream, but I'm not sure if thunder-coding would approve of that or not (because they wanted to encourage the use of both AppArmor and fuse-overlayfs in the main repository). I will try, and find out.

If this mode were a command line argument to the `run-docker.sh` script (and `build-package.sh` script) in upstream, then do you think that would solve your problem, because we could both use that command line argument while building our TUR and local packages, and then you wouldn't have a problem with building packages using TUR immediately after building packages using termux-packages, or am I not yet understanding the problem correctly?

PR #2409, opened 2026-04-05
add ratpoison in tur-repo

LukeGTH:


tur/ratpoison/build.sh
```diff
@@ -0,0 +1,18 @@
+TERMUX_PKG_HOMEPAGE=https://codeberg.org/LukeGTH/Ratpoison-For-Termux
+TERMUX_PKG_DESCRIPTION="Ratpoison is a minimalistic, keyboard-driven tiling window manager for X11, heavily inspired by GNU Screen"
+TERMUX_PKG_LICENSE="GPL-2.0"
+TERMUX_PKG_MAINTAINER="@termux-user-repository"
+TERMUX_PKG_VERSION=1.4.10
+TERMUX_PKG_SRCURL="https://codeberg.org/LukeGTH/Ratpoison-For-Termux.git"
```
robertkirkman: Could you instead use this URL, and apply the Termux support as `.patch` files in the same folder as `build.sh`?

https://download.savannah.gnu.org/releases/ratpoison/ratpoison-1.4.9.tar.gz

robertkirkman (changes requested):

tur/ratpoison/build.sh
LukeGTH: Ok

tur/ratpoison/build.sh
```diff
@@ -0,0 +1,8 @@
+TERMUX_PKG_HOMEPAGE=https://www.nongnu.org/ratpoison/
+TERMUX_PKG_DESCRIPTION="Ratpoison is a keyboard-driven tiling window manager for X11, inspired by GNU Screen"
+TERMUX_PKG_LICENSE="GPL-2.0"
+TERMUX_PKG_MAINTAINER="@termux-user-repository"
+TERMUX_PKG_VERSION=1.4.9
+TERMUX_PKG_SRCURL="https://download.savannah.gnu.org/releases/ratpoison/ratpoison-1.4.9.tar.gz"
```
robertkirkman: Ok, could you make it like this now so that the version number is not hardcoded?

```bash
TERMUX_PKG_VERSION=1.4.9
TERMUX_PKG_SRCURL="https://download.savannah.gnu.org/releases/ratpoison/ratpoison-$TERMUX_PKG_VERSION.tar.gz"
```

tur/ratpoison/build.sh
```diff
@@ -0,0 +1,9 @@
...
+TERMUX_PKG_SHA256="3a36da22a2b5c330bac2580b5b8a5d03b787edc70ca5b1766ac4840f997000d9"
+TERMUX_PKG_BUILD_DEPENDS="autoconf, automake, libtool, pkg-config, libx11, libxext, libxft, libxrandr, libxtst, libxi, libxrender"
```
robertkirkman: Ok, now could you remove this `TERMUX_PKG_BUILD_DEPENDS` line, the reason is because, to calculate the most typical `TERMUX_PKG_BUILD_DEPENDS` for termux packages, what you do is first add all build dependencies, then from that list, subtract all the dependencies that are runtime dependencies as well as build-time dependencies, (`libx11`, `libxext`, `libxft`, `libxrandr`, `libxtst`, `libxi`, `libxrender`), then also subtract all the dependencies that are already installed by the `scripts/setup-termux.sh` script,

https://github.com/termux/termux-packages/blob/master/scripts/setup-termux.sh

(`autoconf`, `automake`, `libtool`, `pkg-config`),

and in this case, once you do that calculation you can see there are no `TERMUX_PKG_BUILD_DEPENDS` left, so you can remove this line.

LukeGTH: To be honest, I used an LDD to check the dependencies, then I included them there. I was finding it strange that there were so many dependencies; it looked like I was going to compile a Systemd

robertkirkman: Yes , `ldd` is good for checking for `TERMUX_PKG_DEPENDS`, keep that, just remove the `TERMUX_PKG_BUILD_DEPENDS` line

LukeGTH: Removed

tur/ratpoison/build.sh
robertkirkman: @LukeGTH pinging just in case you did not see this message yet, if you did already and you are busy then I am sorry

LukeGTH: I'm never busy, I don't do anything all day, my day consists of eating, going to school, and sleeping

robertkirkman:
Ok now there is only one thing left to do which is, please try to combine all the commits together so that in this place, only one commit shows, but keep all the code the same

image

robertkirkman (approved):

PR #2419, opened 2026-04-11
chore(tur-on-device): recompile using an open source `libc.so` to execute build tools

robertkirkman:


licy183:
I don't understand what this PR is trying to do.

robertkirkman:
> I don't understand what this PR is trying to do.

The reason this is good to do is because the `libc.so` file, and other files, of the original termux-docker could not be confirmed to have come from any specific image that was built by a known organization or built in GitHub Actions publicly.

Therefore, any unknown changes that could have been written into these packages by unknown behaviors in the binaries of indeterminate origin will be removed by this PR.

PR #2421, opened 2026-04-12
addpkg(tur/hnefatafl-copenhagen): 5.12.0-1

robertkirkman:
- Unfortunately, it does not work properly. There are multiple severe bugs. I am uploading it so that it easier to save and so that others can help if they know how.


dcampbell24:
possibly related: https://github.com/iced-rs/iced/issues/2847, https://github.com/pop-os/cosmic-text/issues/446

robertkirkman:
(relaying information from Discord to GitHub to keep track of it) It has been discovered that the issue is partially reproducible on Fedora 43, and that the appearance of the pieces on Arch Linux is distorted compared to how they appear on Debian 13. Because of this, I suspect that the symptoms of the issue can probably be influenced by external dependency files not contained within the `hnefatafl-copenhagen` program, which can vary between different Desktop Linux distros, and therefore tracking down the root cause of the differences between the behavior of the program on Desktop Linux distros can probably also narrow down the root cause of the Termux manifestation of the issue.

PR #2426, opened 2026-04-13
addpkg(tur-on-device/qt5-qmake): 5.15.18

robertkirkman:
- Currently, the binary that comes from `pkg install qt5-qmake` is a precompiled binary of indeterminate origin, which also causes it to be difficult for maintainers to bump, resulting in `qt5-qmake` in Termux being outdated by several versions. https://github.com/termux/termux-packages/blob/59a9d95c5761d97dde55eed218df9f87d0995295/x11-packages/qt5-qmake/prebuilt.tar.xz This PR will resolve this problem by simultaneously:
  - Refactoring the abnormal `qt5-qmake` package structure into a normal `TERMUX_ON_DEVICE_BUILD=true` package that can be built efficiently within tur-on-device
  - Bumping the `qt5-qmake` package to version 5.15.18 and adding a version guard to make sure that whenever the package is rebuilt, the maintainer is prompted to bump `qt5-qmake` to match `qt5-qtbase` if necessary
  - Moving the `qt5-qmake` package to tur-on-device in TUR and rebuilding it there for distributing the update

- Tested the resulting build and confirmed to work at least as well as the current `qt5-qmake` by using these commands:

```
git clone -b 5.15.17 --depth=1 https://code.qt.io/qt/qtbase.git
cp -r qtbase/examples/opengl/hellogl2/ .
cd hellogl2
qmake -project \
    "QMAKE_CXXFLAGS+=-I$PREFIX/include/QtWidgets" \
    "QMAKE_LFLAGS+=-lQt5Widgets"
qmake
make
```

### DIFF
--- a/tur-on-device/qt5-qmake/build.sh
+++ b/tur-on-device/qt5-qmake/build.sh
@@ -1,0 +1,98 @@
+TERMUX_PKG_HOMEPAGE=https://www.qt.io/
+TERMUX_PKG_DESCRIPTION="Qt 5 QMake"
+TERMUX_PKG_LICENSE="LGPL-3.0"
+TERMUX_PKG_MAINTAINER="@termux-user-repository"
+TERMUX_PKG_VERSION="5.15.18"
+TERMUX_PKG_SRCURL="https://download.qt.io/archive/qt/${TERMUX_PKG_VERSION%.*}/${TERMUX_PKG_VERSION}/submodules/qtbase-everywhere-opensource-src-${TERMUX_PKG_VERSION}.tar.xz"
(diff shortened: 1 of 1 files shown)


robertkirkman:
@licy183 when you look at the code of a package like `qt5-qmake`, can you understand now why I wanted to improve the source transparency of termux-docker and tur-on-device, then use tur-on-device to recompile packages that are like that?

https://github.com/termux/termux-packages/tree/59a9d95c5761d97dde55eed218df9f87d0995295/x11-packages/qt5-qmake

it seemed to me like doing this is a large improvement over the current situation. however if you do not think this is a good idea and think that I should stop, then let me know.

PR #2430, opened 2026-04-15
bump(tur/surge-cli): 0.8.1

robertkirkman:
- Also fix error `error acquiring lock: failed to ensure config dirs: mkdir /run: read-only file system`

### DIFF
--- a/tur/surge-cli/build.sh
+++ b/tur/surge-cli/build.sh
@@ -2,18 +2,19 @@ TERMUX_PKG_HOMEPAGE=https://github.com/SurgeDM/Surge
 TERMUX_PKG_DESCRIPTION="Blazing fast TUI download manager built in Go for power users"
 TERMUX_PKG_LICENSE="MIT"
 TERMUX_PKG_MAINTAINER="@LukeGTH"
-TERMUX_PKG_VERSION=0.8.0
-TERMUX_PKG_REVISION=1
+TERMUX_PKG_VERSION="0.8.1"
 TERMUX_PKG_SRCURL="https://github.com/SurgeDM/Surge/archive/refs/tags/v${TERMUX_PKG_VERSION}.tar.gz"
-TERMUX_PKG_SHA256="317183ecc2589a407baae10e3e892be4df21171c1bdf0bbc41053f8be910f771"
+TERMUX_PKG_SHA256=fd1da2688f06ea0cc054e3656e3a0075a550e9971c5fe3ff93ca1d7858c24d38
 TERMUX_PKG_BUILD_DEPENDS="file"
+TERMUX_PKG_BUILD_IN_SRC=true
 
-termux_step_pre_configure() {
+termux_step_post_get_source() {
 	termux_setup_golang
+	go mod vendor
 }
 
 termux_step_make() {
-	cd "$TERMUX_PKG_SRCDIR"
+	termux_setup_golang
 	go build -o surge
 }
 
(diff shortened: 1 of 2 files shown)


tur/surge-cli/build.sh
```diff
@@ -2,18 +2,25 @@ TERMUX_PKG_HOMEPAGE=https://github.com/SurgeDM/Surge
...
+TERMUX_PKG_BUILD_IN_SRC=true
 
 termux_step_pre_configure() {
 	termux_setup_golang
+	export GOPATH="$TERMUX_PKG_SRCDIR/go"
+	go get ./cmd
+	chmod +w "$GOPATH" -R
+
+	local d
+	for d in go/pkg/mod/github.com/adrg/xdg*; do
```
licy183: I'd prefer to use `go mod vendor`.

See https://github.com/termux/termux-packages/pull/24770

robertkirkman: Yes you are right. hold on

LukeGTH:
Yes, that's an error I wasn't aware of. I use the XDG_RUNTIME_DIR exported correctly in the .zshrc file, which is why it worked for me

tur/surge-cli/build.sh
robertkirkman: Is it better now? Also, thank you because originally, it was wrong because `@TERMUX_PREFIX@` was not replaced. by using a `.patch` file, that is now automatically replaced correctly.

tur/surge-cli/build.sh
```diff
@@ -2,18 +2,18 @@ TERMUX_PKG_HOMEPAGE=https://github.com/SurgeDM/Surge
...
-TERMUX_PKG_REVISION=1
+TERMUX_PKG_VERSION="0.8.1"
 TERMUX_PKG_SRCURL="https://github.com/SurgeDM/Surge/archive/refs/tags/v${TERMUX_PKG_VERSION}.tar.gz"
-TERMUX_PKG_SHA256="317183ecc2589a407baae10e3e892be4df21171c1bdf0bbc41053f8be910f771"
+TERMUX_PKG_SHA256=fd1da2688f06ea0cc054e3656e3a0075a550e9971c5fe3ff93ca1d7858c24d38
 TERMUX_PKG_BUILD_DEPENDS="file"
+TERMUX_PKG_BUILD_IN_SRC=true
 
-termux_step_pre_configure() {
+termux_step_post_get_source() {
```
TomJo2000, line 11: This needs to happen in `termux_step_post_get_source` so it's done before patching I assume?
https://github.com/termux/termux-packages/wiki/Building-packages#build-steps

robertkirkman: Yes that's correct

tur/surge-cli/build.sh
```diff
@@ -2,18 +2,18 @@ TERMUX_PKG_HOMEPAGE=https://github.com/SurgeDM/Surge
...
 	termux_setup_golang
+	go mod vendor
 }
 
 termux_step_make() {
-	cd "$TERMUX_PKG_SRCDIR"
```
licy183, line 16: I think `termux_setup_golang` is also necessary here.

robertkirkman: Could you show the series of commands that is necessary to reproduce an error if `termux_setup_golang` is not there?

I have wondered about that, I wanted to see it so I can check it in the future on other packages.

licy183: For example, after the source code is extracted, press Ctrl+C, and then use `./build-package.sh -i surge-cli -c`

robertkirkman: ok, thank you. I will try to support the `-c` argument in the future. There is an extremely large number of packages that the `-c` argument does not work on.

LukeGTH: The golang setup was aided by Twaik; without it, I was getting errors like "go: command not found"

PR #2443, opened 2026-04-19
susdu

inrryoff:
# Description:
Add susdu, a modern and secure alternative to tsu. While inspired by tsu, it features a completely different implementation focused on supporting various root managers and improving the user experience with built-in aliases and an easier setup.
## Changes:
- Added tur/susdu/build.sh
- Initial version 1.0.0
- Platform independent script

---

---

---

Screenshot_20260419-194631_Termux

---


tur/susdu/build.sh
```diff
@@ -0,0 +1,17 @@
...
+TERMUX_PKG_LICENSE="MIT"
+TERMUX_PKG_MAINTAINER="@inrryoff"
+TERMUX_PKG_VERSION=1.0.0
+TERMUX_PKG_SKIP_SRC_EXTRACT=true
+TERMUX_PKG_PLATFORM_INDEPENDENT=true
+
+termux_step_get_source() {
+	termux_download "https://githubusercontent.com" \
+		"$TERMUX_PKG_BUILDER_DIR/susdu" \
+		709b76060b85fc68f211c1c11380a253ac3b95905fcdc5215cef29a4777bf2ee
```
robertkirkman: This is a bit weird way to download the source code and it does not work. You should use `TERMUX_PKG_SRCURL`.

inrryoff: I did it this way because my internal compilation tests kept generating this error.

install: cannot stat '/data/data/com.termux/files/home/.termux-build/susdu/src/susdu': No such file or directory
~/tur $

tur/susdu/build.sh
```diff
@@ -0,0 +1,12 @@
+TERMUX_PKG_HOMEPAGE=https://github.com/inrryoff/susdu
```
inrryoff: "Fixed! Now using the official method with a .tar.gz source from releases and TERMUX_PKG_SRCURL. Tested locally and it's building perfectly. Thanks for the guidance!"

PR #2462, opened 2026-04-28
bump(tur-on-device/nqp): 2026.04

rc-chuah:


robertkirkman:
> `Found /data/data/com.termux/files/usr/bin/moar version 2026.03, which is too old. Wanted at least 2026.04`

now that I merged your other PR, do you think that will fix this?

robertkirkman:
By the way, in the future, it is OK now to put changes for packages in `tur` folder and packages in `tur-on-device` folder in the same PR. It should work and allow bumping all of these packages simultaneously without worrying about this error.

robertkirkman:
Good,

next, though, the 2 commits should be combined into 1 commit within the PR.
image

PR #2467, opened 2026-05-01
new(tur/call-cleaner): sweep old call recordings on /sdcard with trash + restore

trehansalil:
## What

Adds `call-cleaner` to TUR — a small Python CLI that sweeps old call recordings on `/sdcard` into a 30-day trash with restore. Designed for OnePlus + other Android phones running Termux. Includes a curses TUI, native scheduling via `termux-job-scheduler`, and Android notifications via Termux:API (when installed).

## Why TUR (and not termux-packages)

Per termux-packages packaging policy, "not installable through language-specific package managers" — `call-cleaner` is pure-Python and is on PyPI as `pip install call-cleaner`. TUR is the right home.

## Recipe details

- Source: PyPI sdist `call_cleaner-0.2.0.tar.gz` (SHA256 verified).
- Pure-Python, platform-independent, no native build steps.
- Runtime deps: `python`, `termux-api` (used best-effort for notifications, low-storage, low-battery checks; cleanly degrades if absent).
- License: MIT.

## Verifying

```sh
pkg install tur-repo
pkg install call-cleaner
cleaner config init
cleaner install-schedule
cleaner doctor
```

## Project links

- Source: https://github.com/trehansalil/call-cleaner
- PyPI: https://pypi.org/project/call-cleaner/
- License: https://github.com/trehansalil/call-cleaner/blob/main/LICENSE


licy183:
I think it should be named as `python-call-cleaner`.

PR #2470, opened 2026-05-04
new package: wiggle

flosnvjx:


tur/wiggle/build.sh
```diff
@@ -0,0 +1,19 @@
+TERMUX_PKG_HOMEPAGE=https://github.com/neilbrown/wiggle
+TERMUX_PKG_DESCRIPTION="A program for applying patches that patch cannot apply because of conflicting changes"
+TERMUX_PKG_LICENSE="GPL-2.0"
+TERMUX_PKG_MAINTAINER="@flosnvjx"
+TERMUX_PKG_VERSION="1.3"
+TERMUX_PKG_SRCURL="https://github.com/neilbrown/wiggle/archive/refs/tags/v$TERMUX_PKG_VERSION.tar.gz"
+TERMUX_PKG_DEPENDS="ncurses"
+TERMUX_PKG_SHA256=SKIP_CHECKSUM
```
robertkirkman: Why `TERMUX_PKG_SHA256=SKIP_CHECKSUM`?

flosnvjx: I picked a previously prepared package as a build.sh template and forgot to fill this line, it should now fixed

tur/wiggle/build.sh
```diff
@@ -0,0 +1,20 @@
...
+TERMUX_PKG_SRCURL="https://github.com/neilbrown/wiggle/archive/refs/tags/v$TERMUX_PKG_VERSION.tar.gz"
+TERMUX_PKG_DEPENDS="ncurses"
+TERMUX_PKG_SHA256="ff92cf0133c1f4dce33563e263cb30e7ddb6f4abdf86d427b1ec1490bec25afa"
+TERMUX_PKG_BUILD_IN_SRC=true
+TERMUX_PKG_EXTRA_MAKE_ARGS="wiggle"
+
+termux_step_post_get_source() {
+	sed -e '/^PREFIX *= */s/=/?=/;
+	s/^CFLAGS *= *.*/CFLAGS := $(CFLAGS) $(CPPFLAGS) $(LDFLAGS) -I./
+	/^install/,/^version/s/\$(STRIP)//' -i Makefile
```
robertkirkman: Could you apply this as a `.patch` file in this case please, so that if the `Makefile` changes too much during an upstream update, this package does not get messed up, but instead has a forced auto update issue because of patch failing to apply? Also, could auto update be enabled on this package?

flosnvjx: done

robertkirkman:
This program has a high repology score.

Why submit it to TUR instead of the main repository?

flosnvjx:
compared to patchutils (which i choose to sent a pr to main), this package isn’t quite actively maintained imo (no release since 2020, but there are activity as late as in 2024), i also found that for my scenario, recountdiff & rediff is usually enough to use, so i send this to tur instead.

and for repology, i cannot access it due to connectivity issue of my isp, but i checked fedora and arch: arch doesn’t have this package in their official repositories, and fedora has seemingly marked it as orphaned and masked it in the latest rawhide source tree. debian does indeed package this program, as noted in the upstream README.

robertkirkman:
Ok

PR #2502, opened 2026-05-23
Add mouseos package

w3356jchooks-cloud:


tur/mouseos/build.sh
```diff
@@ -0,0 +1,17 @@
...
+TERMUX_PKG_VERSION=1.0.0
+TERMUX_PKG_DEPENDS="x11-repo, tigervnc, xfce4, xfce4-goodies, pulseaudio"
+TERMUX_PKG_PLATFORM_INDEPENDENT=true
+
+termux_step_make_install() {
+mkdir -p "$TERMUX_PREFIX/bin"
+cat << 'INNER_EOF' > "$TERMUX_PREFIX/bin/mouseos"
+#!/data/data/com.termux/files/usr/bin/bash
+echo "Initializing Mouse OS..."
+vncserver :1
```
robertkirkman: Is this just a script that prints "initializing mouse os" and then runs `vncserver :1` and then that's it? this does not comply with the TUR package policy to disallow packages which are only a few lines of scripting code:

[TUR package policy](https://github.com/termux-user-repository/tur/wiki/TUR-package-policy.)

licy183:
It looks like this PR cannot be merged, as it is a very simple script.

PR #2505, opened 2026-05-24
addpkg(tur/gdrive-downloader): 2.0

PChaicot:
https://github.com/termux/termux-packages/pull/29883#issuecomment-4517910711
soon, the package gdrive-downloader to be moved into termux user repository


robertkirkman:
Thanks I agree

Could you open a PR to remove it from the main repository at the same time? so that I can click the buttons to merge them at the exact same time

robertkirkman:
for future reference, now it can have `TERMUX_PKG_MAINTAINER="@termux-user-repository"`, I will change that whenever the next time I change all of those again is, since it happens sometimes and I changed them all to match recently so there aren't many currently

PR #2506, opened 2026-05-24
tree-wide: `TERMUX_PKG_MAINTAINER="@termux"` -> `TERMUX_PKG_MAINTAINER="@termux-user-repository"`

robertkirkman:


robertkirkman:
@licy183 If there is some reason that this is not done that I forgot about, please let me know, since I don't remember why these are still with `@termux`, and I had actually thought I changed these already but I guess I did not yet.

PR #2509, opened 2026-05-26
bump(tur-on-device/blender4): 4.5.10

robertkirkman:


robertkirkman:
@licy183 I don't understand why, but blender 4.5.10 was not uploaded, despite building. I will try to build it using the workflow dispatch.

robertkirkman:
In the workflow dispatch, it says an "account suspended" error, but I don't know if it is my account that is suspended, or the TUR account.

licy183:
I think it may be a bug of GitHub.

PR #2523, opened 2026-05-28
Bump shader slang to 2026.10

otreblan:


licy183:
It should be able to get rebuilt successfully...

PR #2557, opened 2026-06-15
antigravity-termux: new package at 1.0.6

javedahmed82:
Adding antigravity-termux, a bridge installer for the official Antigravity CLI in Termux via Ubuntu proot. This package has been tested and fixes TLS and detection issues.

### DIFF
--- a/tur/antigravity-termux/build.sh
+++ b/tur/antigravity-termux/build.sh
@@ -1,0 +1,16 @@
+TERMUX_PKG_HOMEPAGE=https://github.com/javedahmed82/antigravity-termux
+TERMUX_PKG_DESCRIPTION="Antigravity CLI bridge for Termux (Glibc Bridge via Ubuntu proot)"
+TERMUX_PKG_LICENSE="MIT"
+TERMUX_PKG_MAINTAINER="javedahmed82 <javedjakali82@gmail.com>"
+TERMUX_PKG_VERSION=1.0.6
+TERMUX_PKG_SRCURL=https://github.com/javedahmed82/antigravity-termux/archive/refs/tags/v${TERMUX_PKG_VERSION}.tar.gz
+TERMUX_PKG_SHA256=5f5789d1dee76fa7627f64ffa9c1fe7d1ba392a303964647ec38dc18805738d6
+TERMUX_PKG_DEPENDS="proot-distro, curl, tar, coreutils"
+TERMUX_PKG_BUILD_IN_SRC=true
+TERMUX_PKG_PLATFORM_INDEPENDENT=true
+
+termux_step_make_install() {
+    mkdir -p $TERMUX_PREFIX/bin
+    cp $TERMUX_PKG_SRCDIR/install-agy.sh $TERMUX_PREFIX/bin/antigravity-termux-setup
+    chmod +x $TERMUX_PREFIX/bin/antigravity-termux-setup
+}
--- a/tur/antigravity-termux/build.sh
+++ b/tur/antigravity-termux/build.sh
@@ -9,8 +9,28 @@ TERMUX_PKG_DEPENDS="proot-distro, curl, tar, coreutils"
 TERMUX_PKG_BUILD_IN_SRC=true
 TERMUX_PKG_PLATFORM_INDEPENDENT=true
 
+termux_step_post_get_source() {
+	cat << 'EOF' > LICENSE
+MIT License
+
+Copyright (c) 2026 Javed Ahmed
+
+Permission is hereby granted, free of charge, to any person obtaining a copy
+of this software and associated documentation files (the "Software"), to deal
+in the Software without restriction, including without limitation the rights
+to use, copy, modify, merge, publish, distribute, sublicense, and/or sell
+copies of the Software, and to permit persons to whom the Software is
+furnished to do so, subject to the following conditions:
+
+The above copyright notice and this permission notice shall be included in all
+copies or substantial portions of the Software.
+
+THE SOFTWARE IS PROVIDED "AS IS", WITHOUT WARRANTY OF ANY KIND, EXPRESS OR
+IMPLIED, INCLUDING BUT NOT LIMITED TO REGARDLESS OF THE FAILURE OF THIS
+SOFTWARE.
+EOF
+}
+
 termux_step_make_install() {
-    mkdir -p $TERMUX_PREFIX/bin
-    cp $TERMUX_PKG_SRCDIR/install-agy.sh $TERMUX_PREFIX/bin/antigravity-termux-setup
-    chmod +x $TERMUX_PREFIX/bin/antigravity-termux-setup
+	install -Dm755 "$TERMUX_PKG_SRCDIR/install-agy.sh" "$TERMUX_PREFIX/bin/antigravity-termux-setup"
 }
--- a/packages/antigravity-termux/build.sh
+++ b/packages/antigravity-termux/build.sh
@@ -1,0 +1,16 @@
+TERMUX_PKG_HOMEPAGE=https://github.com/javedahmed82/antigravity-termux
+TERMUX_PKG_DESCRIPTION="Antigravity CLI bridge for Termux (Glibc Bridge via Ubuntu proot)"
+TERMUX_PKG_LICENSE="MIT"
+TERMUX_PKG_MAINTAINER="javedahmed82 <javedjakali82@gmail.com>"
+TERMUX_PKG_VERSION=1.0.6
+TERMUX_PKG_SRCURL=https://github.com/javedahmed82/antigravity-termux/archive/refs/tags/v${TERMUX_PKG_VERSION}.tar.gz
+TERMUX_PKG_SHA256=5f5789d1dee76fa7627f64ffa9c1fe7d1ba392a303964647ec38dc18805738d6
+TERMUX_PKG_DEPENDS="proot-distro, curl, tar, coreutils"
+TERMUX_PKG_BUILD_IN_SRC=true
+TERMUX_PKG_PLATFORM_INDEPENDENT=true
+
+termux_step_make_install() {
+    mkdir -p $TERMUX_PREFIX/bin
+    cp $TERMUX_PKG_SRCDIR/install-agy.sh $TERMUX_PREFIX/bin/antigravity-termux-setup
+    chmod +x $TERMUX_PREFIX/bin/antigravity-termux-setup
+}


tur/antigravity-termux/build.sh
```diff
@@ -0,0 +1,16 @@
+TERMUX_PKG_HOMEPAGE=https://github.com/javedahmed82/antigravity-termux
+TERMUX_PKG_DESCRIPTION="Antigravity CLI bridge for Termux (Glibc Bridge via Ubuntu proot)"
```
robertkirkman, line 2: I think this package is redundant with installing Antigravity CLI in the `proot-distro install ubuntu` command of the `proot-distro` package.

javedahmed82: This package provides a simpler user experience by allowing users to access Antigravity directly from Termux without manually setting up and managing an Ubuntu proot environment. It reduces setup complexity and offers a lightweight bridge for users who only need Antigravity functionality.

javedahmed82: The package is not intended to replace proot-distro. Its purpose is to provide a convenient wrapper for users who only need Antigravity CLI functionality without manually entering and managing an Ubuntu environment. It reduces setup steps, provides a native Termux command interface, and improves accessibility for less experienced users.

While Antigravity can already be installed through Ubuntu in proot-distro, this package focuses on improving usability and reducing friction for users who want quick access to Antigravity functionality directly from Termux.

robertkirkman: I have an idea, could you publish your repo on GitHub Container Registry, then rather than needing to use `pkg install tur-repo` and then `pkg install antigravity-termux`, people could use `pkg install proot-distro` and `proot-distro install ghcr.io/javedahmed82/antigravity-termux` and that would basically be almost the same thing with almost the same amount of commands, but not as much maintenance required in TUR and also all your users can get your updates through proot-distro directly from you, without needing to have a TUR package updated?

@licy183 @sylirre sorry to bother you about this idea I just wondered, do you agree with me that maybe GitHub Container registry is a better place than TUR for projects like this, which require a rootfs of GNU/Linux to run the precompiled closed-source program?

javedahmed82: Thank you for the suggestion. My main goal was to provide a simple and accessible installation path for Termux users who want to use Antigravity CLI with minimal setup.

I had not considered GitHub Container Registry as a distribution method, but it does seem like an interesting alternative. I'll look into the feasibility of distributing and updating the Ubuntu environment through GHCR while keeping the user experience simple.

I appreciate the feedback and would be interested in hearing the opinions of the other maintainers as well.

robertkirkman: > provides a native Termux command interface

isn't it true that proot-distro allows, by default, also using Termux commands in `PATH` within the proot-distro, meaning it can also provide that functionality, or is there something major that this version has regarding that which is better?

javedahmed82: You're correct that proot-distro already provides access to the Termux environment and that similar functionality can be achieved manually.

The main benefit of this package is automation rather than additional runtime capabilities. It performs the Ubuntu installation, Antigravity installation, API key forwarding, wrapper creation, and shell integration automatically, reducing the setup process to a single package installation.

My goal was to make Antigravity CLI accessible to less experienced Termux users who may not be familiar with configuring and maintaining a proot environment manually.

robertkirkman: Yes, I think a method which uses GitHub Container Registry can do that even more directly than TUR can, since all of those steps can probably run fully automated within your GitHub repo, including the package publishing. I think you should look into it and see if it meets all of your needs.

robertkirkman: From the user perspective, after using `proot-distro install ghcr.io/javedahmed82/antigravity-termux` and then `proot-distro run antigravity-termux`, what else would be required? To me the setup seems fairly simple and also, importantly makes it clear that antigravity is a Glibc package and cannot be run with bionic libc, following precedents for distinction between a package built in the termux-package-builder and a package built by a company like Google for Ubuntu and then repacked elsewhere without rebuilding it.

sylirre:
1. Package born 3 days ago and which is just a some kind of automation or installer isn't eligible for inclusion into official Termux repo.

   Respecting the [packaging policy](https://github.com/termux/termux-packages/blob/master/CONTRIBUTING.md#packaging-policy) is important to keep repository in healthy and maintainable. Packagers must be sure that submitted project is trustworthy, has future and won't become abandonware anytime soon, won't become ransomware or something like.

   Moreover, I will say that this is in the first place needed for the end users.

2. If your software needs a Linux distribution, it's indeed better to go OCI image way.

   `proot-distro` has support for OCI images and its usage is very easy and convenient.

3. `https://github.com/javedahmed82/antigravity-termux/blob/master/install-agy.sh#L39` - don't do that. API keys shouldn't be hardcoded into scripts.

Anyway, just for experiment I'm approving this package. We'll see how this go in long term.

javedahmed82:
Thank you for the review and approval.

Regarding the API key concern, the script does not hardcode any API keys. It simply forwards existing environment variables ("ANTIGRAVITY_API_KEY" and "GEMINI_API_KEY") from the user's Termux session into the proot environment, so users do not need to configure them twice.

I also appreciate the feedback regarding OCI images and will explore that approach further.

robertkirkman:
I didn't realize this depends on `proot-distro`, and I thought that it implemented its own proot. I have now read the code that sylirre linked, and I can see that the package content is a simple script. In that case, yes this is too simple a wrapper and it falls under the TUR policy to reject packages that are just a few lines of scripting code: [TUR package policy](https://github.com/termux-user-repository/tur/wiki/TUR-package-policy.)

Your tool already relies heavily on `proot-distro`, so I think what would be best is for you to package your project on GitHub Container Registry for use inside proot-distro in a way that's much easier and more reliable and streamlined for users to manage with `proot-distro` commands.

sylirre:
The correct way of building antigravity proot environment

`mkdir ./Antigravity && cd ./Antigravity`

Create `./Dockerfile` with this content:
```
FROM ubuntu:24.04

RUN apt update && apt install -y curl ca-certificates && update-ca-certificates --fresh
RUN curl -fsSL https://antigravity.google/cli/install.sh | bash

ENTRYPOINT ["/root/.local/bin/agy"]
```
*This can be built by Docker and uploaded to GitHub Container Registry.*

Run:
```
proot-distro build -t antigravity-termux:latest .
```

Now you can install it:
```
proot-distro install -n agy antigravity-termux:latest
```

Usage:
```
proot-distro run --env ANTIGRAVITY_API_KEY="apikey" --env GEMINI_API_KEY="apikey" agy
```

If too long, then make a shell alias for it:
```
alias agy="proot-distro run --env ANTIGRAVITY_API_KEY="$ANTIGRAVITY_API_KEY" --env GEMINI_API_KEY="$GEMINI_API_KEY" agy"
```

sylirre:
Btw it should be executable by `glibc-runner`. In such case proot is not needed.

javedahmed82:
Thank you for the detailed feedback and examples. I'll look into the OCI image and glibc-runner approaches.

IntinteDAO:
Maybe it's better to use udocker instead proot? Ofc. udocker runs proot, however udocker is good for application

sylirre:
@IntinteDAO udocker or proot-distro (in both cases Docker image) are acceptable ways to deal with it, but ultimately all depends what do you expect from coding agents. Proot hides all Android-specific issues, so building/debugging stuff for native Termux going to be hard.

IntinteDAO:
> @IntinteDAO udocker or proot-distro (in both cases Docker image) are acceptable ways to deal with it, but ultimately all depends what do you expect from coding agents. Proot hides all Android-specific issues, so building/debugging stuff for native Termux going to be hard.

@sylirre I agree. Another option is to simply support only open-source software. That's not really a problem, though, because there are alternatives to Antigravity (For example, Pi Agent has add-ons that provide access to an Antigravity account).

javedahmed82:
"Thanks for the discussion @IntinteDAO @sylirre. Antigravity CLI is widely used by developers for AI workflows on Termux, which inspired this bridge to simplify setup. I'm open to whatever distribution method (TUR package or GHCR container) best aligns with repository standards."

IntinteDAO:
> "Thanks for the discussion @IntinteDAO @sylirre. Antigravity CLI is widely used by developers for AI workflows on Termux, which inspired this bridge to simplify setup. I'm open to whatever distribution method (TUR package or GHCR container) best aligns with repository standards."

Yes, but I see a significant problem with delivering Antigravity this way. As an AI Agent, it uses system commands, but due to isolation, it can't do that.

Antigravity works as an application, of course, but not with its full capabilities.

javedahmed82:
Thanks for bringing up this point @IntinteDAO! 

The use of `proot-distro` (Ubuntu) was specifically chosen to address glibc dependency requirements for Node/Python/binary extensions that the Antigravity CLI and its agent tooling rely on, which often fail under native Termux Bionic libc.

Regarding agent execution and isolation:
1. **Full Linux Tooling inside PRoot**: Inside the Ubuntu PRoot container, Antigravity has access to standard Glibc development tools (`apt`, `pip`, `npm`, `git`, `gcc`, build tools, dev servers, etc.), allowing it to execute code, run tests, and manage dependencies just as it would on a standard Linux environment.
2. **Workspace Access**: `proot-distro` mounts the host Termux `$HOME` directory inside the container, so file modifications, code creation, and git workflows directly affect the host workspace seamlessly.
3. **Host Termux Bridging**: If accessing host Termux-native binaries (like `termux-api` or `pkg`) from within PRoot is desired for specific workflows, we can easily bind-mount `/data/data/com.termux/files/usr/bin` or add a wrapper bridge in `install-agy.sh`.

I'd love to hear your thoughts on whether binding specific host paths or adjusting container mounts would resolve your concerns for specific use cases!

robertkirkman:
> I'm open to whatever distribution method (TUR package or GHCR container) best aligns with repository standards.

I think that GHCR container is best

javedahmed82:
Thanks for the feedback @robertkirkman!

I completely agree with following TUR's repository standards and best practices. If distributing via a GHCR container (e.g., `ghcr.io/javedahmed82/antigravity-termux`) is the preferred approach for proot/glibc-dependent CLI tools, I am more than happy to pivot to that approach.

Could you share any recommendations or repository guidelines on how TUR prefers container-based packages to be distributed? For instance:
1. Should we publish the container image on GHCR and provide a lightweight TUR wrapper package that invokes it?
2. Or should we host/distribute the container image independently via GHCR?

I want to make sure it aligns best with TUR's architecture and maintenance goals!

robertkirkman:
You should host and distribute the container image independently via GHCR, so that your users get your updates and access to it immediately, similarly to other containers such as the ones in the DockerHub list here:


https://hub.docker.com/search?badges=official&badges=verified_publisher&operating_system=linux&architecture=arm64

javedahmed82:
Thank you so much for the clear guidance and recommendation, @robertkirkman!

That makes complete sense. Distributing independently via GHCR will allow users to receive direct updates immediately without overhead on the TUR repository. 

I will set up the GHCR container image and release workflows on the main repository (`javedahmed82/antigravity-termux`) and close this PR accordingly. 

Thanks again to the TUR team for your time and review!

javedahmed82:
Quick Update: As recommended, we have officially set up and published the multi-arch GHCR Container image (`ghcr.io/javedahmed82/antigravity-termux:latest`) on the main repository (`javedahmed82/antigravity-termux`), featuring support for both Direct Browser OAuth Login and API keys.

Thanks again to @IntinteDAO and @robertkirkman for the valuable guidance and feedback!

PR #2569, opened 2026-06-17
bada: New package

Natarizki:
Bada-Lang: hybrid C/Rust-style programming language compiler with LLVM backend.

Compiles to native ARM64 binaries via LLVM IR, with full C/C++ interoperability
through an integrated C++ transpiler mode. Supports inline assembly, structs,
enums, pattern matching, closures, module imports, and a custom standard library.

Homepage: https://github.com/Natarizki/bada-lang
License: Apache-2.0


robertkirkman:
I have a question about this type of package which is, why do you only want to add this package to Termux and not to other distros?

Natarizki:
yeah, because bada only supports aarch64 (I still haven't upgraded my LLVM IR to support x86_64, i386, i686, and etc.).

robertkirkman:
i see, but surely it also would work on Debian aarch64, right?

Natarizki:
I think so

PR #2577, opened 2026-06-21
glyph: new package

Natarizki:
## Description

Adds `glyph` — a pure C, zero-dependency ASCII art generator for the terminal.

### Features
- Custom font format (`.nfont`) with full FIGlet (`.flf`) support
- 24-bit true color styles: plain, gradient, rainbow, gay, border
- Image-to-ASCII conversion (PNG/JPG) via bundled stb_image
- UTF-8 safe rendering
- Zero external dependencies (pure C + libc only)
- Auto-resolves fonts from local, user, and system FIGlet font directories

### Links
- Source: https://github.com/Natarizki/glyph
- License: Apache-2.0
- Release: https://github.com/Natarizki/glyph/releases/tag/v1.0.0

### Build tested
- [x] Builds successfully via `make install`
- [x] CI verified on Linux (x86_64, ARM64), macOS, and Windows via GitHub Actions

## Checklist
- [x] I have read the contribution guidelines
- [x] `build.sh` follows the sample template format
- [x] Package builds and installs correctly

### DIFF
--- a/tur/glyph/build.sh
+++ b/tur/glyph/build.sh
@@ -1,0 +1,13 @@
+TERMUX_PKG_HOMEPAGE=https://github.com/Natarizki/glyph
+TERMUX_PKG_DESCRIPTION="Pure C ASCII art generator with custom fonts, true color styles, and image-to-ASCII conversion"
+TERMUX_PKG_LICENSE="Apache-2.0"
+TERMUX_PKG_MAINTAINER="Natarizki <natakeren890@gmail.com>"
+TERMUX_PKG_VERSION="1.0.0"
+TERMUX_PKG_SRCURL=https://github.com/Natarizki/glyph/archive/refs/tags/v${TERMUX_PKG_VERSION}.tar.gz
+TERMUX_PKG_SHA256=85134fce77e16fecc204852070fd8004a066bf04fcab1d38e77ec7327f62f8a0
+TERMUX_PKG_DEPENDS=""
+TERMUX_PKG_BUILD_IN_SRC=true
+
+termux_step_make_install() {
+	make install PREFIX=$TERMUX_PREFIX
+}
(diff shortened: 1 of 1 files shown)


hastagaming:
this is cool

robertkirkman:
I have a question about this type of package which is:
1. why do you only want to add this package to Termux and not to other distros?
2. why did you choose the same name and goal as the other `glyph` that has a substantially larger repology score? https://github.com/seatedro/glyph
3. also, @hastagaming why do you think this is cool?

Natarizki:
1. i want to add to other distros 
2. I use glyph as the name because it's sounded better

hastagaming:
@robertkirkman, because i used it and its like fastfetch but you can make own, and it cool beause you can use gif and it can run animation

robertkirkman:
> i want to add to other distros

this is a great plan, I would like you to also submit it to another distro and this way the other distro and TUR can both review this.

Natarizki:
@robertkirkman actually, I want to change the name

tur/glyph/build.sh
```diff
@@ -0,0 +1,13 @@
+TERMUX_PKG_HOMEPAGE=https://github.com/Natarizki/glyph
+TERMUX_PKG_DESCRIPTION="Pure C ASCII art generator with custom fonts, true color styles, and image-to-ASCII conversion"
+TERMUX_PKG_LICENSE="Apache-2.0"
+TERMUX_PKG_MAINTAINER="Natarizki <natakeren890@gmail.com>"
+TERMUX_PKG_VERSION="1.0.0"
+TERMUX_PKG_SRCURL=https://github.com/Natarizki/glyph/archive/refs/tags/v${TERMUX_PKG_VERSION}.tar.gz
```
GourangaDasSamrat, line 6: Please quote the url.

tur/glyph/build.sh
```diff
@@ -0,0 +1,13 @@
...
+TERMUX_PKG_SHA256=85134fce77e16fecc204852070fd8004a066bf04fcab1d38e77ec7327f62f8a0
+# TERMUX_PKG_DEPENDS=""
```
GourangaDasSamrat, line 8: I think this comment is unnecessary.

PR #2588, opened 2026-06-24
dj64dev: fix serious memory corruption

stsp:


robertkirkman:
please increment `TERMUX_PKG_REVISION`, and also, do you have the capability to place this patch in a tag at this location instead of in a patch in github.com/termux-user-repository/tur?

https://github.com/stsp/dj64dev/releases

stsp:
> please increment TERMUX_PKG_REVISION

Done.

> and also, do you have the capability to place this patch in a tag

Oh, this will have a much bigger
chance of breaking the old versions
of everything else, which is in tur. :)
This patch alone has very small
breaking potential, but updating
from tag will likely mean to update
all the relevant packages, not just
this one.

robertkirkman:
> updating from tag will likely mean to update all the relevant packages, not just this one.

I don't understand what you mean exactly, could you explain which other packages would need to be updated?

stsp:
dj64dev is a dep in a long chain:
smallerc->djstub->dj64dev->comcom64->dosemu2
I simply don't know what will
happen if dj64dev alone is updated.
Termux build was tested by other
people in this github forum, I only
submitted the build.

PR #2597, opened 2026-06-26
memtest_vulkan: new package

wataamee777:
Opensource cross-platform tool written in vulkan compute to stress test video memory for stability during overclocking or repair.

Just start application, wait at least 6 minutes and stop testing by Ctrl+C. Detected errors are displayed immediately during test run.

Requires system-provided vulkan loader and driver supporting Vulkan 1.1 (already installed with graphics drivers on most OS).


licy183:
`memtest_vulkan` is not a valid deb name. It should be `memtest-vulkan`.

robertkirkman:
yes,

it's true that this has been packaged in desktop linux distros, which is good. However, it has not yet been packaged in Debian or other distros that use `.deb` such as Termux, which is why we need to be the first to make a package of it named `memtest-vulkan` instead.

https://repology.org/project/memtest-vulkan/versions

PR #2604, opened 2026-06-27
padrelivio: new package

enrilinux:
Padre Livio Fanzaga ASCII art


licy183:
This package doesn't conform the policy of TUR package. It is a simple script.

PR #2606, opened 2026-06-27
ctags: disable libxml

### LINKED ISSUE #2599: [Bug]: ctags: error while loading shared libraries: libxml2.so.2: cannot open shared object file: No such file or directory
### Problem description

Out of sudden, I've started to
get this on CI:
```
ctags: error while loading shared libraries: libxml2.so.2: cannot open shared object file: No such file or directory
```
https://github.com/dosemu2/dosemu2/actions/runs/28249037251/job/83696499131

I don't think I changed anything
like this... Any idea what could happen?

### What steps will reproduce the bug?

Rebuild dj64dev package perhaps?

### What is the expected behavior?

It used to work.

### System information

termux-info:

```

```

stsp:
I am getting this when building ctags:
ctags: error while loading shared libraries: libxml2.so.2: cannot open shared object file: No such file or directory

With this change everything works.
I suspect the deps are set wrongly, and are using the target libxml2. Host's libxml2 was probably installed during container setup, but not any more.

Fixes #2599

### DIFF
--- a/tur/ctags-cross/build.sh
+++ b/tur/ctags-cross/build.sh
@@ -3,11 +3,11 @@ TERMUX_PKG_DESCRIPTION="Universal ctags: Source code index builder"
 TERMUX_PKG_LICENSE="GPL-2.0"
 TERMUX_PKG_MAINTAINER="@termux-user-repository"
 TERMUX_PKG_VERSION="2:6.1.0"
-TERMUX_PKG_REVISION=1
+TERMUX_PKG_REVISION=2
 TERMUX_PKG_SRCURL=https://github.com/universal-ctags/ctags/archive/refs/tags/v${TERMUX_PKG_VERSION:2}.tar.gz
 TERMUX_PKG_SHA256=1eb6d46d4c4cace62d230e7700033b8db9ad3d654f2d4564e87f517d4b652a53
-TERMUX_PKG_BUILD_DEPENDS="libiconv, libjansson, libxml2, libyaml"
-TERMUX_PKG_EXTRA_CONFIGURE_ARGS="--enable-tmpdir=$TERMUX_PREFIX/tmp --disable-static"
+TERMUX_PKG_BUILD_DEPENDS="libiconv, libjansson, libyaml"
+TERMUX_PKG_EXTRA_CONFIGURE_ARGS="--enable-tmpdir=$TERMUX_PREFIX/tmp --disable-static --disable-xml"
 TERMUX_PKG_BUILD_IN_SRC=true
 TERMUX_PKG_HOSTBUILD=true
 


licy183:
Termux's libxml2 has bumped to `libxml2.so.16`. Host's libxml2 should still be `libxml2.so.2`.

stsp:
But I wonder what was actually
wrong. Maybe you or @robertkirkman 
have any idea? Maybe we had to
build libxml2-host? If so, other deps
of that package are also unsafe...
Confused.

robertkirkman:
@stsp @licy183 so, I believe that I 100% fully understand what is occurring that resulted in

- #2599 

It is because in this PR, the termux cross-compiler container has been globally upgraded to Ubuntu 26.04:

- https://github.com/termux/termux-packages/pull/30225

```
builder@b030037d8efe:~/termux-packages$ ls -lah /usr/lib/x86_64-linux-gnu/libxml2.so.16
lrwxrwxrwx 1 root root 17 Jun 19 17:43 /usr/lib/x86_64-linux-gnu/libxml2.so.16 -> libxml2.so.16.1.2
builder@b030037d8efe:~/termux-packages$ cat /etc/os-release 
PRETTY_NAME="Ubuntu 26.04 LTS"
NAME="Ubuntu"
VERSION_ID="26.04"
VERSION="26.04 LTS (Resolute Raccoon)"
VERSION_CODENAME=resolute
ID=ubuntu
ID_LIKE=debian
HOME_URL="https://www.ubuntu.com/"
SUPPORT_URL="https://help.ubuntu.com/"
BUG_REPORT_URL="https://bugs.launchpad.net/ubuntu/"
PRIVACY_POLICY_URL="https://www.ubuntu.com/legal/terms-and-policies/privacy-policy"
UBUNTU_CODENAME=resolute
LOGO=ubuntu-logo
builder@b030037d8efe:~/termux-packages$ 
```

**Ubuntu 26.04** does **not** have the same `/usr/lib/x86_64-linux-gnu` libraries that **Ubuntu 24.04** did.

As you can see above, it does **not** have a `libxml2.so.2`. It has a `libxml2.so.16`.

>  Host's libxml2 should still be `libxml2.so.2`.

The observation appears to suggest that that statement might be false, unless TUR is supposed to be downgraded back to Ubuntu 24.04 and that is what you mean.

> Maybe we had to build libxml2-host? 

I would have done that, but I don't understand why licy183 says that "Host's libxml2 should still be `libxml2.so.2`", so we might want to find out why before doing anything.

It is possible that my observation is incomplete or incorrect.

robertkirkman:
My hypothesis leads me to assume that @stsp did not test **rebuilding `ctags-cross` with no code changes other than `TERMUX_PKG_REVISION`**.

If they did already, and still saw the same error already and that's why they proceeded to this, then I would be wrong.

tur/ctags-cross/build.sh
```diff
@@ -3,11 +3,11 @@ TERMUX_PKG_DESCRIPTION="Universal ctags: Source code index builder"
 TERMUX_PKG_LICENSE="GPL-2.0"
 TERMUX_PKG_MAINTAINER="@termux-user-repository"
 TERMUX_PKG_VERSION="2:6.1.0"
-TERMUX_PKG_REVISION=1
+TERMUX_PKG_REVISION=2
 TERMUX_PKG_SRCURL=https://github.com/universal-ctags/ctags/archive/refs/tags/v${TERMUX_PKG_VERSION:2}.tar.gz
 TERMUX_PKG_SHA256=1eb6d46d4c4cace62d230e7700033b8db9ad3d654f2d4564e87f517d4b652a53
-TERMUX_PKG_BUILD_DEPENDS="libiconv, libjansson, libxml2, libyaml"
-TERMUX_PKG_EXTRA_CONFIGURE_ARGS="--enable-tmpdir=$TERMUX_PREFIX/tmp --disable-static"
+TERMUX_PKG_BUILD_DEPENDS="libiconv, libjansson, libyaml"
```
robertkirkman, line 9: In my opinion, this package is written somewhat incorrectly because it has bionic libc dependencies in `TERMUX_PKG_BUILD_DEPENDS`, but _only_ builds code in `termux_step_host_build()`, which is a function that does **not** run after `termux_setup_toolchain_29()`, which means that this package produces **only** GNU/Linux binaries, which means that the bionic libc packages marked here in `TERMUX_PKG_BUILD_DEPENDS` are not having an effect on this package, and should probably be marked only on a different package that actually builds bionic-libc code linked to those dependencies instead.

stsp:
> My hypothesis leads me to assume that @stsp did not test rebuilding ctags-cross with
> no code changes other than TERMUX_PKG_REVISION.

Yes, I didn't check that.
Let me know if I have to, and
I'll do later.

robertkirkman:
@stsp Yes I think you should check that, and, I think something that is helpful is to download the `ctags-cross` package and extract it inside the GNU/Linux docker container, and from GNU/Linux use the `ldd` command on it, and that will show the locations in the docker container where the runtime dependencies of the binary are. If they are located anywhere inside the `/data` folder, they are probably Termux package dependencies, but if they are located anywhere in the `/usr` folder, they are Ubuntu package dependencies and they will be affected primarily by changes to the [`Dockerfile`](https://github.com/termux/termux-packages/blob/master/scripts/Dockerfile) and the [`setup-ubuntu.sh`](https://github.com/termux/termux-packages/blob/master/scripts/setup-ubuntu.sh).

licy183:
Ah, my local builder is still on Ubuntu 24; I wasn't aware that the builder for the main repository had already been updated to Ubuntu 26.

robertkirkman:
As far as I can tell, `ctags-cross` is the only TUR `-cross` package, so it is the only package in TUR that stores a binary for the Ubuntu container, and hopefully is the only one in TUR affected by this problem.

robertkirkman:
Oh, there are also `-host-tools` package for chromium building packages, which might also be affected, but they are regularly updated and recompiled, so might not often show this type of error.

stsp:
> use the ldd command on it, and that will show

No, it won't:
```
$ldd ctags 
	linux-vdso.so.1 (0x0000749ad5fc3000)
	libxml2.so.2 => not found
	libjansson.so.4 => /usr/lib/x86_64-linux-gnu/libjansson.so.4 (0x0000749ad5db3000)
	libyaml-0.so.2 => /usr/lib/x86_64-linux-gnu/libyaml-0.so.2 (0x0000749ad5d91000)
	libpcre2-8.so.0 => /usr/lib/x86_64-linux-gnu/libpcre2-8.so.0 (0x0000749ad5ce5000)
	libc.so.6 => /usr/lib/x86_64-linux-gnu/libc.so.6 (0x0000749ad5a00000)
	/lib64/ld-linux-x86-64.so.2 (0x0000749ad5fc5000)
```

There is no rpath, so full path
is only available in /etc/ld.so.cache
I am afraid.

robertkirkman:
> `libxml2.so.2 => not found`

I meant that it will show the location of the library after the `ctags-cross` has been rebuilt and linked to `libxml2.so.16`, or it will show the location of the library of the old one if you run the `ldd ctags` command in Ubuntu 24.04 instead of in Ubuntu 26.04

stsp:
https://github.com/dosemu2/dosemu2/actions/runs/28249037251/job/84044378312

This run is with --disable-xml
removed. It will really really help
if you make termux to print the
exact configure line before running
it (we currently don't see arguments
that are passed), but this:
```
checking for libxml-2.0 >= 2.7.7... yes
```
suggests that libxml is now in.
Things went much further, so
disabling libxml confirmed to
be unneeded. However now it
fails with:
```
ld: error: unknown argument '--require-defined=____crt1_startup'
```
Have to investigate how come this
perfectly valid option unsupported...

robertkirkman:
On my screen, it looks like this build compiled a `thunks_c.o` with `arm-linux-androideabi-clang` and then attempted to link it using `x86_64-linux-gnu-gcc`,

image

but I'm not sure what I am looking at in this build, so I'm not sure if this is meant to be built for GNU/Linux, or for the target Android, at this point of the log.

robertkirkman:
oh, maybe the android build was linked right above that, and then this other part is a build for GNU/Linux, that needs to run right afterward?

robertkirkman:
To work around problems with the `x86_64-linux-gnu-gcc` command of Ubuntu 26.04, I recommend trying to get the build system of the affected package to use the `$TERMUX_HOST_LLVM_BASE_DIR/bin/clang` command for the same purpose instead.

It is reportedly significantly more stable during Ubuntu upgrades in the termux-package-builder Docker image than the `x86_64-linux-gnu-gcc` command, for the same purpose of compiling code for GNU/Linux for running in the builder.

robertkirkman:
`$TERMUX_HOST_LLVM_BASE_DIR/bin/clang` is not a made up variable/location, by the way; it is actually there and is Clang 21 for GNU/Linux not Android, which you can use.

stsp:
> oh, maybe the android build was linked right above that, and then this other part is a build for
> GNU/Linux, that needs to run right afterward?

Yes, because there is also this
a few lines above:
```
[0174](https://github.com/dosemu2/dosemu2/actions/runs/28249037251/job/84044378312#step:6:10180)
x86_64-linux-gnu-gcc -Wall -O2 -Wmissing-declarations -Wwrite-strings -ggdb3 -Wunused -Wmissing-prototypes -DREV_ID=\"Non-git_build\" -fno-pic -fno-pie -ffreestanding -fno-stack-protector -march=i386 -fcf-protection=none -m32 -I. -I/home/builder/.termux-build/comcom64/src/src/32/.. -DCOMCOM_VERSION=\"0.4\" -I/data/data/com.termux/files/usr/i386-pc-dj32/include -DDJ64 -DDJ32 -nostdinc -imacros /data/data/com.termux/files/usr/i386-pc-dj32/include/dj64/a_mac.h -I. -o thunks_c.o -c /home/builder/.termux-build/comcom64/src/src/32/../thunks_c.c
```
Its just that multiple targets
are built in parallel, so its not
quite clear from the output
what relates to what.
Maybe clang could do better,
but in this particular case -Wl
fails, i.e. this is a linker option.
Strange...

stsp:
AI says this is because lld
needs --underfined instead
of --require defined...
Will try.

PR #2615, opened 2026-07-01
revbump(tur): for Python 3.14

robertkirkman:


robertkirkman:
Do not merge until immediately (preferably within seconds) after the main repository receives the python 3.14 from the staging repository.

PR #2616, opened 2026-07-01
revbump(tur-on-device): for Python 3.14

robertkirkman:
Blender packages excluded for a separate, subsequent build


robertkirkman:
Do not merge until immediately (preferably within seconds) after the main repository receives the python 3.14 from the staging repository.

tur/python-opengl/build.sh
```diff
@@ -3,7 +3,7 @@ TERMUX_PKG_DESCRIPTION="A cross-platform open source Python binding to the stand
 TERMUX_PKG_LICENSE="BSD"
 TERMUX_PKG_MAINTAINER="@fervi"
 TERMUX_PKG_VERSION="3.1.6"
-TERMUX_PKG_REVISION=4
+TERMUX_PKG_REVISION=5
```
robertkirkman, line 6: `tur/python-opengl` is a dependency of `tur-on-device/usd`, therefore it needs to also be rebuilt in the same PR that is rebuilding tur-on-device packages, otherwise an error occurs.

PR #2619, opened 2026-07-03
ci: switch back from staging termux-packages to normal after Python 3.14

robertkirkman:
- After https://github.com/termux-user-repository/tur/pull/2615

- After https://github.com/termux-user-repository/tur/pull/2616

### DIFF
--- a/.github/workflows/package_on_device.yml
+++ b/.github/workflows/package_on_device.yml
@@ -273,11 +273,7 @@ jobs:
       env:
         ARCH: ${{ matrix.target_arch }}
       run: |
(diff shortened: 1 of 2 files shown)


robertkirkman:
Since this has never been done or publicly coordinated before in the time I have worked on Termux, I think it is good if I post a link to the matching commit that had to be done at the same time in the main repository, so that future contributors understand how to do this and what it is for.

https://github.com/termux/termux-packages/commit/cc0b56024ddb2b78ce4b1b4cb045abc52a3ffc83

PR #2637, opened 2026-07-12
Added rish package

Nixxy-lv:
Added the rish package, allowing users to run the ADB Shell directly in Termux via Shizuku


tur/rish/rish
```diff
@@ -0,0 +1,26 @@
...
+  fi
+  if [ -w $DEX ]; then
+    echo "Cannot remove the write permission of $DEX."
+    echo "You can copy to file to terminal app's private directory (/data/data/<package>, so that remove write permission is possible"
+    exit 1
+  fi
+fi
+
+[ -z "$RISH_APPLICATION_ID" ] && export RISH_APPLICATION_ID="com.termux"
+/system/bin/app_process -Djava.class.path="$DEX" /system/bin --nice-name=rish rikka.shizuku.shell.ShizukuShellLoader "$@"
```
robertkirkman: Could you please implement it as a patch or package of the release archive of the actually official one?

https://github.com/RikkaApps/Shizuku/blob/2650830c5b099ae0dd34fedf614d4f592ca05d65/manager/src/main/assets/rish

tur/rish/rish
```diff
@@ -0,0 +1,26 @@
+#!/system/bin/sh
+mkdir /data/data/com.termux/files/usr/share/rish/ > /dev/null 2>%1
+
+PREFIX=/data/data/com.termux/files/usr
+DEX="$PREFIX/share/rish/rish_shizuku.dex"
+
+if [ ! -f "$DEX" ]; then
```
robertkirkman: I'd like to suggest that you implement here downstream, if making this a TUR package, the compatibility check message for users of termux-docker.

https://github.com/termux/termux-x11/blob/dfbb20f9b2bfe3ba2d9f796184e778c761174b3d/termux-x11#L2

I believe spreading that where relevant can help solidify the formal difference between termux-docker and standard Android, prevent confusion and disappointment by setting user expectations for what termux-docker is compatible with, and raise awareness for users of normal Android who notice as well, which Termux packages require a separate helper app `.apk` to be installed, and which can operate standalone.

Nixxy-lv: Implemented! Added the compatibility check for termux-docker via sed right at the beginning of the rish script. Thanks for the suggestion!

tur/rish/rish
Nixxy-lv: Done! I have refactored the build.sh to fetch the source code directly from the official archive and implemented the downstream compatibility check for termux-docker as suggested. Thank you for the helpful feedback!

tur/rish/build.sh
```diff
@@ -0,0 +1,26 @@
...
+termux_step_make_install() {
+    mkdir -p $TERMUX_PREFIX/bin
+    mkdir -p $TERMUX_PREFIX/share/rish
+
+    cp manager/src/main/assets/rish $TERMUX_PREFIX/bin/
+    cp manager/src/main/assets/rish_shizuku.dex $TERMUX_PREFIX/share/rish/
+
+    sed -i "s|DEX=\"\$BASEDIR\"/rish_shizuku.dex|DEX=\"$TERMUX_PREFIX/share/rish/rish_shizuku.dex\"|" $TERMUX_PREFIX/bin/rish
+
+    sed -i 's/export RISH_APPLICATION_ID="PKG"/export RISH_APPLICATION_ID="com.termux"/' $TERMUX_PREFIX/bin/rish
```
robertkirkman: Please do not hardcode string `com.termux` and instead, use this variable: `$TERMUX_APP__PACKAGE_NAME`

tur/rish/build.sh
```diff
@@ -0,0 +1,26 @@
...
+
+    sed -i '2i\if [ -f /.dockerenv ] || [ -f /run/.containerenv ]; then echo "rish: Shizuku requires a real Android environment and is incompatible with termux-docker."; exit 1; fi' $TERMUX_PREFIX/bin/rish
```
robertkirkman: There is a reason why I implemented the check by checking for the presence of `/system/bin/getprop`. That is because I am developing a fork of termux-docker that will run termux packages **outside** of docker and all other container runtimes without the presence of SurfaceFlinger, Android Runtime, or Zygote.

I have not released all code yet and I do not plan to release all code until I am finished, but I am releasing pieces of it when asked and when necessary.

A minimal standalone chroot version of it is publicly released in the description of this issue:

- https://github.com/tiann/KernelSU/issues/3520

Here is one of my publicly released demo videos which runs additional, unreleased code:

https://www.youtube.com/watch?v=yDxvbkGJdbI

robertkirkman: also the reason I requested the check be added originally is because the software here uses `getprop`. For that reason, there will be an error if `/system/bin/getprop` does not exist, exactly the same situation that `termux-x11` exemplifies. Therefore it is clearly only logical for the check to be for `getprop`, not for docker environment files.

tur/rish/build.sh
Nixxy-lv: Done! Replaced the hardcoded string with the $TERMUX_APP__PACKAGE_NAME environment variable in the sed patch. Thanks for catching that!

tur/rish/build.sh
Nixxy-lv: I have updated the check to look for the presence/executability of /system/bin/getprop instead of relying on Docker-specific files. This will properly safeguard rish across standard containers, chroots, and your upcoming custom environment.

tur/rish/build.sh
```diff
@@ -0,0 +1,49 @@
...
+	:
+}
+
+termux_step_make_install() {
+	mkdir -p "$TERMUX_PREFIX/bin"
+	mkdir -p "$TERMUX_PREFIX/share/rish"
+
+	local APK_NAME="shizuku-v${TERMUX_PKG_VERSION}.r1049.0e53409-release.apk"
+	termux_download \
+		"https://github.com/RikkaApps/Shizuku/releases/download/v${TERMUX_PKG_VERSION}/${APK_NAME}" \
```
robertkirkman: Is it possible for you to compile the `rish_shizuku.dex` in this `build.sh`? I know the `termux-x11-nightly` package doesn't, but that causes problems for me such as not being able to patch that package during recompiling normally, so I would prefer to avoid adding more packages that download precompiled java dexes if possible

PR #2638, opened 2026-07-13
open-vcdiff: new package

flosnvjx:


tur/open-vcdiff/build.sh
```diff
@@ -0,0 +1,36 @@
...
+"
+
+termux_step_post_get_source() {
+	git fetch --unshallow
+
+	local testCommit="$(git log --format=%H -1)"
+	local testCommitDateRev="$(TZ=UTC git log --date=format:%Y%m%d --format=%cd | uniq -c | awk '{if ($1>1) {print ($2 ".r" ($1-1));} else print $2; exit;}')"
+
+	if [ "$_SUBTREE_COMMIT_DATEREV/$_SUBTREE_COMMIT" != "$testCommitDateRev/$testCommit" ]; then
+		if [ "$_SUBTREE_COMMIT_STICK" -gt 0 ]; then
```
robertkirkman: Why are all these `if` statements about checking version necessary for a project that is archived and will never receive updates?

tur/open-vcdiff/build.sh
```diff
@@ -0,0 +1,36 @@
+TERMUX_PKG_HOMEPAGE="https://github.com/google/open-vcdiff"
+TERMUX_PKG_DESCRIPTION="An encoder/decoder for the VCDIFF (RFC3284) format"
+TERMUX_PKG_LICENSE="Apache-2.0"
+TERMUX_PKG_MAINTAINER="@termux-user-repository"
+_SUBTREE_COMMIT_DATEREV=20190311
+_SUBTREE_COMMIT=868f459a8d815125c2457f8c74b12493853100f9
+_SUBTREE_COMMIT_STICK=1
+TERMUX_PKG_VERSION="0.8.4.git$_SUBTREE_COMMIT_DATEREV"
+TERMUX_PKG_SRCURL="git+https://github.com/google/open-vcdiff"
```
robertkirkman, line 8: I have a security concern about adding very old software that is abandoned and will never receive updates upstream, due to the likelihood of vulnerabilities and responsibility for the security of Termux packages.

For example, I have never added my package of archaic `squashfs-tools` at the last commit before the upstream migration to `pthread_cancel()` even to TUR, because I want to make sure that users see a security warning and have to explicitly bypass the warning manually _before_ they install the package, rather than after.

- https://github.com/termux/termux-packages/issues/22959#issuecomment-2621339981

@licy183 is it ok to add `open-vcdiff` despite being abandoned and, if so then should I add `squashfs-tools` version 4.2 also?

licy183: I don't actually think that deprecated packages are a problem. Many packages in TUR are already in that situation...

robertkirkman: Ok, understood. I use `unsquashfs` so I might add it to TUR in the future after `open-vcdiff`.

tur/open-vcdiff/build.sh
flosnvjx: Ah, the OctoDroid (which I use to access GitHub sometimes) doesn't support to display archival status of a repo.

flosnvjx: Just removed some unnecessary logic in the build.sh.

Since there is a small chance that upstream may unarchive the repo someday (which I doesn't think it will happen, though), I think it is no hurt to keep the outdate commit anchor detection logic.

robertkirkman (approved):

PR #2642, opened 2026-07-15
bump(tur/codex): 0.144.4

ancientcatz:


robertkirkman:
You know about how this version of `codex` is currently non-functional, right?

I think it might be ok to package this version,

- https://github.com/wallentx/codex-termux/pull/5

if it works, given that upstream does not accept PRs, so it cannot be upstreamed.

PR #2660, opened 2026-07-18
addpkg(tur/coded): 0.1.1

HashShin:
Add Coded, a mobile-first browser code editor designed to work well in Termux.

Upstream: https://github.com/HashShin/coded
License: MIT

### DIFF
--- a/tur/coded/build.sh
+++ b/tur/coded/build.sh
@@ -1,0 +1,26 @@
+TERMUX_PKG_HOMEPAGE=https://github.com/HashShin/coded
+TERMUX_PKG_DESCRIPTION="A mobile-first code editor that runs in your browser"
+TERMUX_PKG_LICENSE="MIT"
+TERMUX_PKG_MAINTAINER="@HashShin"
+TERMUX_PKG_VERSION="0.1.1"
+TERMUX_PKG_SRCURL="https://github.com/HashShin/coded/archive/refs/tags/v${TERMUX_PKG_VERSION}.tar.gz"
+TERMUX_PKG_SHA256=0123375865e6ae49d21a75ba535ccfe32bf767df3314ff60f71cc68dec02c5bc
+TERMUX_PKG_BUILD_IN_SRC=true
+TERMUX_PKG_AUTO_UPDATE=true
+
+termux_step_pre_configure() {
+	termux_setup_golang
+}
+
+termux_step_make() {
+	go build \
+		-trimpath \
+		-ldflags="-s -w" \
+		-o coded \
+		.
+}
+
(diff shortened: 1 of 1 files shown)


robertkirkman:
Ok,

@HashShin I tested this and it is good.

I will approve it on two conditions:

- Do you promise to keep using this and making updates for this to fix issues in the future for a long time?
- Do you promise to not turn it into a virus?

This passes the TUR policy, but your project is created 5 days ago. For many recent projects like that, there is a concern that they will upload it once and never make updates again, until eventually the account is abandoned, becomes hacked and turns into a virus. It is good to allow a good project like this on TUR, but for a project that is this new, only if you agree to the above conditions. Do you agree?

robertkirkman:
also, Just make the title of the PR and the commit this please:

addpkg(tur/coded): 0.1.1

combining the 3 commits into 1 commit:

image

robertkirkman:
I was also wondering, what is this big thing, do you promise it is not a virus?

https://github.com/HashShin/coded/blob/6c1b8c64cbc12a21033edc8868bf9b467c76c306/static/vendor/prism/prism-bundle.min.js

can you explain what that does and do you have an unminified version of it?

HashShin:
> Ok,
> 
> @HashShin I tested this and it is good.
> 
> I will approve it on two conditions:
> 
> * Do you promise to keep using this and making updates for this to fix issues in the future for a long time?
> * Do you promise to not turn it into a virus?
> 
> This passes the TUR policy, but your project is created 5 days ago. For many recent projects like that, there is a concern that they will upload it once and never make updates again, until eventually the account is abandoned, becomes hacked and turns into a virus. It is good to allow a good project like this on TUR, but for a project that is this new, only if you agree to the above conditions. Do you agree?

Yes, I agree to both I'll keep maintaining this and fixing issues over time, and add some features if ideas come up. I won't ever turn it into a virus. If I ever can't continue, I'll notify you instead of leaving it abandoned. Thanks

> I was also wondering, what is this big thing, do you promise it is not a virus?
> 
> https://github.com/HashShin/coded/blob/6c1b8c64cbc12a21033edc8868bf9b467c76c306/static/vendor/prism/prism-bundle.min.js
> 
> can you explain what that does and do you have an unminified version of it?

This is PrismJS, an open-source library I'm using to improve the syntax highlighting in the editor. It looks big and unreadable because it's minified (compressed to load faster) and bundles about 24 language grammars into one file. It's not a virus; it just colors the code on screen.l

robertkirkman:
Ok good, then next just combine the 3 commits into 1 commit in this PR

robertkirkman (approved):

PR #2669, opened 2026-07-19
revbump(tur/bazel8): move to standalone package

xJonathanLEI:
Bazel 9 was [released in Jan 2026](https://blog.bazel.build/2026/01/20/bazel-9.html) and here we are half a year later still stuck on Bazel 8 in TUR.

We should bump the `bazel` package to the 9.x LTS but before that we should archive the current major version as `bazle8`, following the convention established by previous major version bumps.

This PR was created simply by:

- deleting the subpackage
- duplicating the `bazel` package into `bazel8`
- adding the version description suffix, and bumping the revision
- removing the two unversioned install commands in `termux_step_make_install`

Will submit follow-up PR for rolling the `bazel` package to 9.x after this is merged.

### DIFF
--- a/tur/bazel/bazel8.subpackage.sh
+++ b/tur/bazel/bazel8.subpackage.sh
@@ -1,5 +1,0 @@
-TERMUX_SUBPKG_DESCRIPTION="Correct, reproducible, and fast builds for everyone (Version 8)"
-TERMUX_SUBPKG_DEPEND_ON_PARENT=deps
-TERMUX_SUBPKG_INCLUDE="
-bin/bazel-$TERMUX_PKG_VERSION
-"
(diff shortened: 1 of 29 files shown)


licy183:
Deleting a `bazel8` subpackage also requires revbumping `bazel`; otherwise, `bazel8` cannot be truly compiled successfully.

xJonathanLEI:
@licy183 I did bump the new `bazel8`. Just for clarification, you mean I need to _also_ bump the main `bazel`?

licy183:
Yes, `bazel` should be revbumped, because its file `tur/bazel/bazel8.subpackage.sh` is deleted.

xJonathanLEI:
Done. Force pushed with an additional bump to `bazel` itself. Thanks!

licy183:
Thanks!

PR #2680, opened 2026-07-21
[NOT WORKING] addpkg(tur/pcsx2): 2.2.0

IntinteDAO:
The PCSX2 build completes, but no working. When I tried to run BIOS, I received an error that something wrong is with GPU Drivers. Maybe it's true, maybe it's false. I put it, because maybe one day I (or someone else) will fix a issues

### DIFF
--- a/tur/pcsx2/build.sh
+++ b/tur/pcsx2/build.sh
@@ -1,0 +1,30 @@
+TERMUX_PKG_HOMEPAGE=https://pcsx2.net
+TERMUX_PKG_DESCRIPTION="A PlayStation 2 emulator"
+TERMUX_PKG_LICENSE="GPL-3.0-or-later"
+TERMUX_PKG_MAINTAINER="@IntinteDAO"
+TERMUX_PKG_VERSION="2.2.0"
+TERMUX_PKG_SRCURL=git+https://github.com/PCSX2/pcsx2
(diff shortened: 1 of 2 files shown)


robertkirkman:
the last that I had heard all ARM versions of pcsx2 are closed source and the official one does not support ARM, only x86?

robertkirkman:
(on a much deeper level than whether the ELF is compiled for x86 or not, because in emulators, the dynamic recompiler has to be able to read the opcodes of the console ROM, then emit opcodes for the target CPU)

robertkirkman:
I recommend testing x86_64 build only first. if you progress to the point of being able to get the x86_64 build to work and then the aarch64 build fails due to above explained problem, I suggest `TERMUX_PKG_EXCLUDED_ARCHES="arm, aarch64"`.

IntinteDAO:
> I recommend testing x86_64 build only first. if you progress to the point of being able to get the x86_64 build to work and then the aarch64 build fails due to above explained problem, I suggest `TERMUX_PKG_EXCLUDED_ARCHES="arm, aarch64"`.

I think you have right. PCSX2 on Debian is also x86_64 only

However it's a fork for ARM: https://github.com/ARMSX2/ARMSX2
Maybe we should test it and create other packages for other architectures

PR #2702, opened 2026-08-01
addpkg(tur/openarena): 0.8.8

IntinteDAO:
@robertkirkman 

* The problem is definitely with that archive on archive.org. I'm not sure if I should re-upload it - maybe you have an idea.
* The audio seems to be more delayed than usual. I have no idea why.

If you have any suggestions for changes to the package, I'd love to hear them.

### DIFF
--- a/tur/openarena/build.sh
+++ b/tur/openarena/build.sh
@@ -1,0 +1,56 @@
+TERMUX_PKG_HOMEPAGE=https://openarena.ws
+TERMUX_PKG_DESCRIPTION="OpenArena game engine"
+TERMUX_PKG_LICENSE="GPL-2.0"
+TERMUX_PKG_MAINTAINER="@termux"
+TERMUX_PKG_VERSION=0.8.8.20260412
+TERMUX_PKG_SRCURL=https://github.com/OpenArena/engine/archive/95c63426c896b2f6277ee940f053a649dd97364a.tar.gz
+TERMUX_PKG_SHA256=3bbd517fbf1e49e701a786adab56032108bf082ccf9915a003ca9ed9c8639db0
+TERMUX_PKG_DEPENDS="libcurl, libjpeg-turbo, libxmp, openal-soft, openarena-data, opusfile, sdl2, zlib, libx11"
+TERMUX_PKG_GROUPS="games"
+TERMUX_PKG_BUILD_IN_SRC=true
+
+termux_step_pre_configure() {
+	if [ "$TERMUX_ARCH" = "i686" ]; then
+		CFLAGS+=" -DC_ONLY -DNO_VM_COMPILED"
+	fi
+}
+
+termux_step_make() {
+	local oa_arch="$TERMUX_ARCH"
+	local extra_flags=""
+	if [ "$oa_arch" = "i686" ]; then
+		oa_arch="x86"
+		extra_flags="HAVE_VM_COMPILED=false"
+	fi
+
+	make -j $TERMUX_PKG_MAKE_PROCESSES \
+		PLATFORM=linux ARCH="$oa_arch" $extra_flags \
+		BUILD_CLIENT=1 BUILD_SERVER=1 USE_CURL=1 USE_MUMBLE=0 \
+		USE_CODEC_VORBIS=1 USE_CODEC_XMP=1 USE_CODEC_OPUS=1 USE_LOCAL_HEADERS=0
+}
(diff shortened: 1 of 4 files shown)


tur/openarena/build.sh
```diff
@@ -0,0 +1,39 @@
...
+	install -Dm755 "$build_dir"/openarena* "$build_dir"/oa_ded* -t "$TERMUX_PREFIX/libexec/openarena/"
+	mkdir -p "$TERMUX_PREFIX/bin"
+	ln -sf "$TERMUX_PREFIX/libexec/openarena/openarena"* "$TERMUX_PREFIX/bin/openarena"
+	ln -sf "$TERMUX_PREFIX/libexec/openarena/oa_ded"* "$TERMUX_PREFIX/bin/openarena-ded"
+
+	install -Dm644 -t "$TERMUX_PREFIX/share/applications" "$TERMUX_PKG_BUILDER_DIR/openarena.desktop"
+	install -Dm644 misc/quake3-tango.png "$TERMUX_PREFIX/share/icons/hicolor/256x256/apps/openarena.png"
+
+	termux_download \
+		"https://web.archive.org/web/20240716135128if_/http://download.tuxfamily.org/openarena/rel/088/openarena-0.8.8.zip" \
```
robertkirkman: Why does Arch Linux package not need this thing?

https://aur.archlinux.org/cgit/aur.git/tree/PKGBUILD?h=openarena

IntinteDAO: > Why does Arch Linux package not need this thing?
> 
> https://aur.archlinux.org/cgit/aur.git/tree/PKGBUILD?h=openarena

My bad :( Too many files ;)

IntinteDAO:
@robertkirkman So Termux-packages or stay in TUR? :P 
But maybe you had an idea with (pulseaudio?) delay.

robertkirkman:
but why are there two source URLs?

IntinteDAO:
> but why are there two source URLs?

Generally speaking, the first package contains the OpenArena game engine, which is a modified version of ioquake3.

The second package contains the game files.

robertkirkman:
However, why does this have only one source URL?

image

I asked you this earlier less explicitly, saying "Why does Arch Linux package not need this thing?", but you did not explain why, so I assume you did not understand my question.

IntinteDAO:
> However, why does this have only one source URL?
> image
> 
> I asked you this earlier less explicitly, saying "Why does Arch Linux package not need this thing?", but you did not explain why, so I assume you did not understand my question.

Ach, ok - sorry.

Generally speaking, the second OpenArena package provides pre-compiled binaries for Windows, Linux, and macOS -we have to compile them (the entire engine) ourselves. That’s why we have two packages: we take the engine from one (and compile it), and the files from the other.

This Arch Linux package doesn’t build the engine; it uses a pre-built binary instead.

obraz

robertkirkman:
Ok That makes sense, thanks!

tur/openarena/build.sh
```diff
@@ -0,0 +1,56 @@
...
+
+	install -Dm644 -t "$TERMUX_PREFIX/share/applications" "$TERMUX_PKG_BUILDER_DIR/openarena.desktop"
+	install -Dm644 misc/quake3-tango.png "$TERMUX_PREFIX/share/icons/hicolor/256x256/apps/openarena.png"
+
+	termux_download \
+		"https://downloads.sourceforge.net/project/oarena/openarena-${TERMUX_PKG_VERSION%.*}.zip" \
+		"$TERMUX_PKG_TMPDIR/openarena-${TERMUX_PKG_VERSION%.*}.zip" \
+		"5a8faf7f5b51f351b0a1618c06b6b98a5f1a6758f1d39818de2c87df2a0bac4a"
+
+	mkdir -p "$TERMUX_PREFIX/share/openarena"
```
robertkirkman, line 52: If you build the package 2 times in a row without deleting the docker container in between building it the 1st and 2nd times, does this part work correctly and still install the files in the same place as the first time?

IntinteDAO: Sure, I don't have any problems with that - building the app works correctly every time.

robertkirkman: Yes you're right

robertkirkman:
Now could you combine all the commits into 1 commit?

image

robertkirkman (approved):

PR #2703, opened 2026-08-01
addpkg(tur/redeclipse): 2.0.0

IntinteDAO:


robertkirkman (approved):
It's completely good and working, and has a high repology score, but it is very intensive and requires a lot of system resources so it can be in TUR since it won't work well on most devices

PR #2704, opened 2026-08-03
addpkg(tur/lincity): 1.13.1

IntinteDAO:


robertkirkman:
There's already `pkg install lincity-ng`. Isn't that kind of the same thing?

IntinteDAO:
> There's already `pkg install lincity-ng`. Isn't that kind of the same thing?

No :) Lincity is heavily inspired by the first version of SimCity, while Lincity-NG is inspired by later versions.

They're two different games, even though they're both about the same thing.

robertkirkman:
Ok good

tur/lincity/0001-translations.patch
```diff
@@ -0,0 +1,3514 @@
...
++msgid "It's at a solar power station."
++msgstr "Esta e uma usina de energia solar."
++
++#: engine.c:674
++msgid "It's at a substation."
++msgstr "Isto e uma sub-estacao."
++
++#: engine.c:676
++msgid "It's at a residential area."
++msgstr "Isto e uma area residencial."
```
robertkirkman: `.po` files are classic glibc translation files.. are you sure this works in bionic termux?

robertkirkman:
Why does Wikipedia's page of "Lincity" say that it has "Successor lincity-ng?" https://en.wikipedia.org/wiki/Lincity is there something in Lincity that is not in Lincity-ng?

IntinteDAO:
> Why does Wikipedia's page of "Lincity" say that it has "Successor lincity-ng?" https://en.wikipedia.org/wiki/Lincity is there something in Lincity that is not in Lincity-ng?

Lincity is more classic. I’m not saying that many people will play Lincity (especially since it’s no longer being developed), but Debian treats it as a separate game, and that makes sense. It’s a bit like comparing Super Mario Bros. and Super Mario Bros. 3. The games are similar in some ways, but completely different.

The core concepts of these games are the same, but Lincity NG features "3D gameplay" a new interface, music, effects, and so on.

robertkirkman:
Ok

tur/lincity/0001-translations.patch
IntinteDAO: I think I'll delete it, but not because it doesn't work on Termux - it's because I can't get it to work on Debian, lol

IntinteDAO: Fixed

tur/lincity/lincity.xpm
```diff
@@ -0,0 +1,47 @@
...
+".++.@+++@@++++++++$$@=@#@$$$$$+.",
+".++@.++@@@@@++++++++$$$$$$$$$$+.",
+".++.@+@@@@@..+++++++++$$$$$$$$+.",
+".++@.@@@@.....++++++++++$$$$+++.",
+".++.@.@.@.@.@.+++++++++++++++++.",
+".++@.@.@.@.@.@.++++++++++++++++.",
+".+++++%%%%%%%+++++++$$$$$++++++.",
+".++++%-%%%%%+$+++++$#####$+++++.",
+".+++%%%%%;%%%$$+++$#######$$+++.",
+".++++%-%;%;%+$$$+$#####..##$$++.",
```
robertkirkman, line 37: Where did this come from, is it from the source code?

robertkirkman: Yes I know it's a picture but I wonder if it can be copied during the build from the place where you found it

tur/lincity/build.sh
```diff
@@ -0,0 +1,26 @@
+TERMUX_PKG_HOMEPAGE=https://sourceforge.net/projects/lincity/
+TERMUX_PKG_DESCRIPTION="A city/country simulation game for X11"
+TERMUX_PKG_LICENSE="GPL-2.0"
+TERMUX_PKG_MAINTAINER="@IntinteDAO"
+TERMUX_PKG_VERSION=1.13.1
+TERMUX_PKG_SRCURL=https://downloads.sourceforge.net/project/lincity/Lincity%20Development%20Source/${TERMUX_PKG_VERSION}/lincity-${TERMUX_PKG_VERSION}.tar.gz
```
GourangaDasSamrat, line 6: url should be quoted.

```suggestion
TERMUX_PKG_SRCURL="https://downloads.sourceforge.net/project/lincity/Lincity%20Development%20Source/${TERMUX_PKG_VERSION}/lincity-${TERMUX_PKG_VERSION}.tar.gz"
```

PR #2708, opened 2026-08-05
new package: docker-qemu

GourangaDasSamrat:
what: adds a `docker-qemu` package that installs a small CLI (`docker-qemu`)
for running Docker inside a QEMU/Alpine virtual machine on Termux, without
root. Commands: setup, init, start, stop, ssh, status. Depends only on
already-packaged qemu-utils / qemu-system-x86-64-headless / openssh / wget,
nothing new to compile — TERMUX_PKG_SKIP_SRC_EXTRACT is used since the two
scripts install straight from $TERMUX_PKG_BUILDER_DIR, same pattern as
tur-repo's own build.sh.

why: Termux users on unrooted devices have no way to run Docker directly
(no cgroups/namespaces access without root). This package gets them a
working Docker environment by running a lightweight Alpine VM under
software-emulated QEMU (no KVM available without root, so it's TCG —
expect it to be noticeably slower than native Docker, which is called out
in the package description so nobody's surprised).

The Alpine ISO (~66MB) is not bundled — `docker-qemu setup` downloads it
from dl-cdn.alpinelinux.org at first run, along with Alpine's own published
.sha256 file, and verifies against that before creating the disk image.
This means the checksum never needs to be manually updated when Alpine
ships new point releases; only ALPINE_VERSION/ALPINE_BRANCH at the top of
docker-qemu.sh need bumping.

### DIFF
--- a/tur/docker-qemu/alpine-provision.sh
+++ b/tur/docker-qemu/alpine-provision.sh
@@ -1,0 +1,8 @@
+#!/bin/sh
+set -e
+setup-alpine
+apk update
+apk add docker docker-cli-compose
+rc-update add docker boot
+service docker start
+echo "Docker is ready. Test with: docker run hello-world"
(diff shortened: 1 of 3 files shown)


robertkirkman:
Is this similar to this?

https://github.com/luisdavim/termux-qemu-docker

GourangaDasSamrat:
@robertkirkman yes, it's similar to luisdavim/termux-qemu-docker,
but mine's one don't need go or building process.

PR #2709, opened 2026-08-06
qemu-system-i386-10: fix build

licy183:


robertkirkman:
Thank you I remembered to change it for `qemu-system-x86-64-9` but not `qemu-system-i386-10`

PR #2713, opened 2026-08-07
[NEW PACKAGE] tnode v0.1.3

Irithell:
tnode is a POSIX compatibility shim for Android's SDCARD filesystem,
distributed as a CLI wrapper (tnode) and a shared library (libthook.so)
injected via LD_PRELOAD.

It transparently patches libc calls that fail on FUSE-backed SDCARD
partitions, enabling Node.js, Python, C, C++, Ruby and more runtimes to
run natively from /storage/emulated/0 in Termux without root or proot.

The build script downloads pre-compiled aarch64 and x86_64 binaries
from the GitHub release and verifies them against checksums.sha256
before installing.

Related upstream contributions during tnode development:
- termux/termux-packages#30831 / #30833 — Go FcntlFlock patch (merged)
- oven-sh/bun#36852 / #36853 — Bun linkat EACCES fallback (merged)

### DIFF
--- a/tur/tnode/build.sh
+++ b/tur/tnode/build.sh
@@ -1,0 +1,28 @@
+TERMUX_PKG_HOMEPAGE=https://github.com/Irithell/tnode
+TERMUX_PKG_DESCRIPTION="POSIX compatibility shim for Android SDCARD filesystem — LD_PRELOAD bridge for Node.js, Python, Ruby, C and C++ runtimes in Termux"
+TERMUX_PKG_LICENSE="MIT"
+TERMUX_PKG_MAINTAINER="@Irithell"
+TERMUX_PKG_VERSION="0.1.4"
+TERMUX_PKG_SRCURL=https://github.com/Irithell/tnode/archive/refs/tags/v${TERMUX_PKG_VERSION}.tar.gz
+TERMUX_PKG_SHA256=720509789e5eb17b79e624582666d97faf835d40927947060808e88f4a6b5b60
+TERMUX_PKG_AUTO_UPDATE=true
+TERMUX_PKG_BUILD_IN_SRC=true
+
+termux_step_pre_configure() {
+	termux_setup_rust
+}
+
+termux_step_make() {
+	cargo build \
+		--target "$CARGO_TARGET_NAME" \
+		--release \
+		--locked
+}
+
+termux_step_make_install() {
+	install -Dm755 "target/${CARGO_TARGET_NAME}/release/tnode" \
+		"${TERMUX_PREFIX}/bin/tnode"
+
+	install -Dm644 "target/${CARGO_TARGET_NAME}/release/libthook.so" \
+		"${TERMUX_PREFIX}/lib/libthook.so"
+}


licy183:
Generally, packages in TUR should download their source code and then execute the build process with GitHub Action, rather than directly downloading external binaries.

tur/tnode/build.sh
```diff
@@ -0,0 +1,34 @@
...
+TERMUX_PKG_AUTO_UPDATE=true
+TERMUX_PKG_BUILD_IN_SRC=true
+
+termux_step_pre_configure() {
+	termux_setup_rust
+
+	case "$TERMUX_ARCH" in
+		aarch64) CARGO_TARGET="aarch64-linux-android" ;;
+		x86_64)  CARGO_TARGET="x86_64-linux-android"  ;;
+		*)       termux_error_exit "unsupported architecture: $TERMUX_ARCH" ;;
```
robertkirkman: I have a question which is, if this entire thing is software designed specifically for Termux, then why can't you implement support for 32-bit architectures, which are widely used by Termux users? With software designed primarily for desktop PC, the situation is different because almost no desktop PCs are still 32-bit, so it can be excused if the upstream does not have the resources to maintain 32-bit support, but I don't really think it's excusable for a Termux-only project.

Irithell: Thank you for bringing this up. You make a valid point — 32-bit Android
(armv7 and i686) is still widely used in the Termux ecosystem.

The target-specific assembly (global_asm!) is used only on aarch64 for
the dynamic linker trampoline. All other architectures automatically fall
back to a generic pure-Rust atomic pointer implementation (AtomicPtr),
which adjusts pointer size natively for 32-bit targets.

I just verified cross-compilation for armv7-linux-androideabi and
i686-linux-android — both build cleanly with zero code changes required.
Full 32-bit support (armv7 and i686) will be added to the release pipeline
in the next version :)

tur/tnode/build.sh
```diff
@@ -8,17 +8,30 @@ TERMUX_PKG_SHA256=f81b959bab01c818c8830994df1bdfaee1f4dd96ed5e198a46e9b543141877
 TERMUX_PKG_AUTO_UPDATE=true
 TERMUX_PKG_BUILD_IN_SRC=true
 
+case "$TERMUX_ARCH" in
+	aarch64) CARGO_TARGET="aarch64-linux-android"    ;;
+	x86_64)  CARGO_TARGET="x86_64-linux-android"     ;;
+	arm)     CARGO_TARGET="armv7-linux-androideabi"  ;;
+	i686)    CARGO_TARGET="i686-linux-android"       ;;
```
robertkirkman: Are you able to just use this variable instead `$CARGO_TARGET_NAME`?

Irithell: algo assim ficaria melhor?

```
TERMUX_PKG_HOMEPAGE=https://github.com/Irithell/tnode
TERMUX_PKG_DESCRIPTION="POSIX compatibility shim for Android SDCARD filesystem — LD_PRELOAD bridge for Node.js, Python, Ruby, C and C++ runtimes in Termux"
TERMUX_PKG_LICENSE="MIT"
TERMUX_PKG_MAINTAINER="@Irithell"
TERMUX_PKG_VERSION="0.1.4"
TERMUX_PKG_SRCURL=https://github.com/Irithell/tnode/archive/refs/tags/v${TERMUX_PKG_VERSION}.tar.gz
TERMUX_PKG_SHA256=f81b959bab01c818c8830994df1bdfaee1f4dd96ed5e198a46e9b5431418777d
TERMUX_PKG_AUTO_UPDATE=true
TERMUX_PKG_BUILD_IN_SRC=true

termux_step_pre_configure() {
	termux_setup_rust
}

termux_step_make() {
	cargo build \
		--target "$CARGO_TARGET_NAME" \
		--release \
		--locked
}

termux_step_make_install() {
	install -Dm755 "target/${CARGO_TARGET_NAME}/release/tnode" \
		"${TERMUX_PREFIX}/bin/tnode"

	install -Dm644 "target/${CARGO_TARGET_NAME}/release/libthook.so" \
		"${TERMUX_PREFIX}/lib/libthook.so"
}

Comment (robertkirkman) [reply tur/tnode/build.sh]: Yes like that

**Comment by Irithell** [comment]
@robertkirkman is there anything else pending for the review to be approved?

**Comment by robertkirkman** [comment]
I will benchmark its performance with code projects in `/storage/emulated/0` vs performance of working with projects only in `/data/data/com.termux/files/home` soon

File: tur/tnode/build.sh [hunk 3]
```diff
@@ -0,0 +1,28 @@
+TERMUX_PKG_HOMEPAGE=https://github.com/Irithell/tnode
+TERMUX_PKG_DESCRIPTION="POSIX compatibility shim for Android SDCARD filesystem — LD_PRELOAD bridge for Node.js, Python, Ruby, C and C++ runtimes in Termux"
+TERMUX_PKG_LICENSE="MIT"
+TERMUX_PKG_MAINTAINER="@Irithell"
+TERMUX_PKG_VERSION="0.1.4"
+TERMUX_PKG_SRCURL=https://github.com/Irithell/tnode/archive/refs/tags/v${TERMUX_PKG_VERSION}.tar.gz
```
GourangaDasSamrat, line 6: url should be quoted 

```suggestion
TERMUX_PKG_SRCURL="https://github.com/Irithell/tnode/archive/refs/tags/v${TERMUX_PKG_VERSION}.tar.gz"
```

tur/tnode/build.sh
```diff
@@ -0,0 +1,28 @@
+TERMUX_PKG_HOMEPAGE=https://github.com/Irithell/tnode
+TERMUX_PKG_DESCRIPTION="POSIX compatibility shim for Android SDCARD filesystem — LD_PRELOAD bridge for Node.js, Python, Ruby, C and C++ runtimes in Termux"
+TERMUX_PKG_LICENSE="MIT"
+TERMUX_PKG_MAINTAINER="@Irithell"
```
GourangaDasSamrat, line 4: TERMUX_PKG_MAINTAINER format is : `Name <email>`

PR #2729, opened 2026-08-17
New package: plan9port

Krush206:
This is the continuation of https://github.com/termux/termux-packages/pull/30124.


Krush206:
The build is failing, @robertkirkman. `which` is missing.

robertkirkman:
> The build is failing, @robertkirkman. `which` is missing.

@Krush206 

Honestly let's put `which` in this file because I meant to but I keep forgetting.

https://github.com/termux/termux-packages/blob/master/scripts/setup-termux.sh

It's supposed to be because `which` is preinstalled in the termux-packages ubuntu package builder and I am trying to make them synchronized until these problems are all gone.

this (tur-on-device) is not in Ubuntu anymore which is why this error occurred

robertkirkman:
do you want to make the PR (to put `which` in `PACKAGES+=` variable in that file in termux-packages repository) or should I? yes editing that file will affect tur-on-device, and immediately after it gets merged yes.

robertkirkman:
So that you learn what caused the `which` problem and how I'm trying to fix it in the big picture:

`scripts/setup-ubuntu.sh` and `scripts/setup-termux.sh` are desynchronized meaning that the commands and libraries available in the Ubuntu termux-package-builder-image that gets built end up being different and much more than the ones in the pure-Termux  termux-package-builder-image that gets built,

I have started synchronizing them but I did not get everything synchronized yet and they can probably never be 100% synchronized because Ubuntu and Termux are just different, but the closer they can be the more stable building is.

So when an error occurs because of missing build dependency only in on-device building or tur-on-device building, instead of adding it to `TERMUX_PKG_BUILD_DEPENDS` of the package, I check if that dependency is preinstalled in the Ubuntu package builder image already, either from Ubuntu core packages or from added packages from `setup-ubuntu.sh`, and if so then I add it to `setup-termux.sh` instead so the error never occurs in other packages that only fail when building on-device but not in the Ubuntu package builder image.

licy183:
Alternatively, I would prefer to include `which` in `BUILD_DEPENDS`.

Krush206:
By synchronized, do you mean the `PACKAGE` entries, @robertkirkman? I've no problem issuing a pull request. However, I'm concerned this would be _too much_ for a single package. Wouldn't other packages break?

robertkirkman:
> Alternatively, I would prefer to include `which` in `BUILD_DEPENDS`.

But that does not solve the problem of `which` being unavailable in on-device building of absolutely every single package in the repository, while being available to every single package being built in the ubuntu termux-package-builder container.

robertkirkman:
> By synchronized, do you mean the `PACKAGE` entries, @robertkirkman? I've no problem issuing a pull request. However, I'm concerned this would be _too much_ for a single package. Wouldn't other packages break?

No it will fix all the packages not break them

Krush206:
Damn. Now I can't get past this linking error, @robertkirkman. https://github.com/termux/termux-packages/issues/14205 may give a hint. However, adding `ndk-multilib` as dependency haven't made the fix.

robertkirkman:
@Krush206 `ndk-multilib` does not help with this,

what you need is called `libandroid-utimes`,

and you can activate it using `LDFLAGS+=" -landroid-utimes"`

Krush206:
Is the symbol `mblen` part of `libandroid-support`, @robertkirkman?

robertkirkman:
Yes

```
~ $ nm -gD $PREFIX/lib/libandroid-support.so
                 U __cxa_atexit@LIBC
                 U __cxa_finalize@LIBC
                 U __errno@LIBC
                 U __register_atfork@LIBC
                 U __stack_chk_fail@LIBC
0000000000006c54 T iswalnum
0000000000006c90 T iswalpha
0000000000006cdc T iswblank
0000000000006cec T iswcntrl
0000000000006d1c T iswctype
0000000000006de4 T iswdigit
0000000000006df4 T iswgraph
0000000000006e30 T iswlower
0000000000006e5c T iswprint
0000000000006ed0 T iswpunct
0000000000006f10 T iswspace
0000000000006f3c T iswupper
0000000000006f64 T iswxdigit
0000000000006f80 T mblen
                 U mbrtowc@LIBC
0000000000006f90 T mbsnrtowcs
0000000000007104 T mbsrtowcs
                 U mbtowc@LIBC
                 U strcmp@LIBC
00000000000075ec T towctrans
0000000000007534 T towlower
00000000000073a8 T towupper
                 U wcschr@LIBC
000000000000753c T wcswidth
000000000000759c T wctrans
0000000000006d78 T wctype
0000000000007608 T wcwidth
~ $ 
```

Krush206:
Too bad. The build is crashing here too. =(

I don't know what's exit code 143. I suspect the build crashed.

Krush206:
Shall we proceed with `inferno-emu`, @robertkirkman?

robertkirkman:
> Too bad. The build is crashing here too. =(
> 
> I don't know what's exit code 143. I suspect the build crashed.

yes, it means that unfortunately plan9port is having the same problem here that it does in cross-compilation.

That just doesn't happen on your phone, right?

eventually, I can check whether it happens in local termux-docker, or only in GitHub Actions

Krush206:
I haven't tried from another device. However, on my main device, I had to toggle a developer option. The package shall build fine with this developer option enabled.

robertkirkman:
Is it the disable child process restrictions setting?

Krush206:
Yes, @robertkirkman.

PR #2732, opened 2026-08-18
addpkg(tur/ffdec): 26.2.1

TheMightyArie:
This PR will add FFdec (also known as JPEXS Free Flash Decompiler) package to TUR packages list.

~~Build for all architectures through `termux-docker` are successful ([results](https://github.com/TheMightyArie/tur/actions/runs/32141210076)).~~

~~**NOTE**: Since this is a Java application, I could not compile it for TUR. So I skipped that and built it on-device instead.~~

### DIFF
--- a/tur/ffdec/build.sh
+++ b/tur/ffdec/build.sh
@@ -1,0 +1,62 @@
+TERMUX_PKG_HOMEPAGE=https://github.com/jindrapetrik/jpexs-decompiler
+TERMUX_PKG_DESCRIPTION="JPEXS Free Flash Decompiler"
+TERMUX_PKG_LICENSE="GPL-3.0"
+TERMUX_PKG_MAINTAINER="@termux-user-repository"
+TERMUX_PKG_VERSION="26.2.1"
+TERMUX_PKG_GIT_BRANCH="version${TERMUX_PKG_VERSION}"
+TERMUX_PKG_SRCURL="git+https://github.com/jindrapetrik/jpexs-decompiler"
+TERMUX_PKG_UPDATE_TAG_TYPE="latest-release-tag"
+TERMUX_PKG_DEPENDS="openjdk-21, openjdk-21-x"
+TERMUX_PKG_AUTO_UPDATE=true
+TERMUX_PKG_BUILD_IN_SRC=true
+TERMUX_PKG_HOSTBUILD=true
+
+termux_step_host_build() {
+	if [[ ! -v ANT_HOME ]]; then
+		termux_download_ubuntu_packages ant
+	fi
+}
+
+termux_step_pre_configure() {
+	if [[ ${TERMUX_ON_DEVICE_BUILD} = true ]]; then
+		export JAVA_HOME=${TERMUX__PREFIX__LIB_DIR}/jvm/java-21-openjdk
+	fi
+}
+
+termux_step_make() {
+	export PATH="$TERMUX_PKG_HOSTBUILD_DIR/ubuntu_packages/usr/bin:$PATH"
+	ant build
+}
+
+termux_step_make_install() {
+	mkdir -p $TERMUX__PREFIX/share/java
+	cp -r ./dist/ $TERMUX__PREFIX/share/java/ffdec/
+	cat << EOF > $TERMUX__PREFIX/bin/ffdec
+#!$TERMUX_PREFIX/bin/sh
+
+java -jar $TERMUX__PREFIX/share/java/ffdec/ffdec.jar "\$@"
+EOF
+	cat << EOF > $TERMUX__PREFIX/bin/ffdec-cli
+#!$TERMUX_PREFIX/bin/sh
+
+java -jar $TERMUX__PREFIX/share/java/ffdec/ffdec-cli.jar "\$@"
+EOF
+	chmod 755 $TERMUX__PREFIX/bin/ffdec
+	chmod 755 $TERMUX__PREFIX/bin/ffdec-cli
+	# Removes all Windows executables
+	rm -f $TERMUX__PREFIX/share/java/ffdec/*.exe
+	rm -f $TERMUX__PREFIX/share/java/ffdec/*.bat
+	mkdir -p $TERMUX__PREFIX/share/applications
+	cat << EOF > $TERMUX__PREFIX/share/applications/com.jpexs.decompiler.desktop
+[Desktop Entry]
+Type=Application
+Version=26.2.1
+Name=FFdec
+Comment=JPEXS Free Flash Decompiler
+Exec=ffdec
+Icon=$TERMUX__PREFIX/share/java/ffdec/icon.png
+Categories=Development;
+MimeType=application/x-shockwave-flash;
+EOF
+	chmod +x $TERMUX__PREFIX/share/applications/com.jpexs.decompiler.desktop
+}
--- a/tur/ffdec/fix-paths.patch
+++ b/tur/ffdec/fix-paths.patch
@@ -1,0 +1,39 @@
+diff --git a/build.xml b/build.xml
+index e02160dfb..6f2bd332f 100644
+--- a/build.xml
++++ b/build.xml
+@@ -218,8 +218,8 @@
+             toFile="${app.deb.desktop.temp}"
+             name="${product.name}"
+             comment="${app.description.short}"
+-            exec="/usr/bin/${app.package}"
+-            icon="/usr/share/java/${app.package}/icon.png"
++            exec="@TERMUX_PREFIX@/bin/${app.package}"
++            icon="@TERMUX_PREFIX@/share/java/${app.package}/icon.png"
+             categories="Development;Network;Utility;WebDevelopment;Java"
+         />
+         <deb 
+@@ -231,11 +231,11 @@
+             <version upstream="${version}" />
+             <maintainer name="${vendor}" email="${app.vendor.mail}"/>
+             <description synopsis="${app.description.short}">${app.description}</description>
+-            <tarfileset dir="${dist.dir}" prefix="usr/share/java/${app.package}">
++            <tarfileset dir="${dist.dir}" prefix="@TERMUX_PREFIX@/share/java/${app.package}">
+                 <include name="**"/>
+             </tarfileset>
+-            <tarfileset file="${dist.dir}/${app.script}" fullpath="usr/bin/${app.package}" filemode="755" />
+-            <tarfileset file="${app.deb.desktop.temp}" fullpath="usr/share/applications/${app.deb.desktop.file}.desktop"/>
++            <tarfileset file="${dist.dir}/${app.script}" fullpath="@TERMUX_PREFIX@/bin/${app.package}" filemode="755" />
++            <tarfileset file="${app.deb.desktop.temp}" fullpath="@TERMUX_PREFIX@/share/applications/${app.deb.desktop.file}.desktop"/>
+         </deb>
+         <property name="deb.file" value="${releases.dir}/${prefix.filename}_${version}${version.suffix}.deb" />        
+         <move file="${deb.file.orig}" tofile="${deb.file}" />
+@@ -340,7 +340,7 @@
+     </target>
+     <target name="osx-app-bundle-no-dist" depends="-loadversion">
+         <property name="app.script.temp" value="${app.bundle.dir}/${app.osx.dir}.app/Contents/MacOS/${app.osx.dir}"/>
+-        <echo file="${app.script.temp}">#!/usr/bin/env bash&#10;$(dirname "$${BASH_SOURCE[0]}")/../Resources/${app.script} "$$&#64;"</echo>
++        <echo file="${app.script.temp}">#!@TERMUX_PREFIX@/bin/env bash&#10;$(dirname "$${BASH_SOURCE[0]}")/../Resources/${app.script} "$$&#64;"</echo>
+         <property name="app.info.temp" value="${app.bundle.dir}/${app.osx.dir}.app/Contents/Info.plist"/>
+         <echo file="${app.info.temp}"><![CDATA[<?xml version="1.0" encoding="UTF-8"?>
+ <!DOCTYPE plist PUBLIC "-//Apple//DTD PLIST 1.0//EN" "http://www.apple.com/DTDs/PropertyList-1.0.dtd">


tur-on-device/ffdec/build.sh
```diff
@@ -0,0 +1,55 @@
...
+
+termux_step_make() {
+	ant build
+}
+
+termux_step_make_install() {
+	mkdir -p $TERMUX__PREFIX/share/java
+	cp -r ./dist/ $TERMUX__PREFIX/share/java/ffdec/
+	cat << EOF > $TERMUX__PREFIX/bin/ffdec
+#!/data/data/com.termux/files/usr/bin/sh
```
robertkirkman: Don't forget to replace `/data/data/com.termux/files/usr` with `$TERMUX_PREFIX` in these

TheMightyArie: I did not know that works on shebangs.

robertkirkman:
> NOTE: Since this is a Java application, I could not compile it for TUR. So I skipped that and built it on-device instead.

what error occurred, this was because of the Openjdk versions in the Ubuntu termux-package-builder, right?

TheMightyArie:
No, the reason is it can't use Termux's Apache ANT builder. And also, none of the workarounds for installing regular Apache ANT worked for the build in `tur`.

TheMightyArie:
Update: There is an another workaround to fix the cross-compilation, which will make the package move from TUR-On-Device to TUR. Until I push the changes, do not run any checks.

robertkirkman:
wait, should I run it yet? what?

TheMightyArie:
You can try now. NoName provided me another fix you mentioned.

TheMightyArie:
Run it in TUR action, not TUR-On-Device.

robertkirkman:
Ok it is running in TUR action not TUR-on-device

TheMightyArie:
I had to remove `ant` from `TERMUX_PKG_BUILD_DEPENDS` to avoid possible conflict when installing by `termux_download_ubuntu_packages` function.

robertkirkman:
> I had to remove `ant` from `TERMUX_PKG_BUILD_DEPENDS` to avoid possible conflict when installing by `termux_download_ubuntu_packages` function.

does that actually create a conflict? what error occurred then?

TheMightyArie:
> does that actually create a conflict? what error occurred then?

I'm not sure but build system says `ant` command not found, and pretty sure package is not installed from Ubuntu as it's already installed from Termux, and as I said TUR build system can't use Termux's `ant` when it's not being compiled on-device. But looks like it's still failing.

TheMightyArie:
> But looks like it's still failing.

I think I have to perform a host build instead of `termux_step_make()`.

tur/ffdec/build.sh
```diff
@@ -0,0 +1,56 @@
...
+TERMUX_PKG_SHA256=bc74ba4610df1da263f012011e1f45e3a8189427ee087f69d95f40c105a07d98
+TERMUX_PKG_DEPENDS="openjdk-21, openjdk-21-x"
+TERMUX_PKG_AUTO_UPDATE=false
+TERMUX_PKG_BUILD_IN_SRC=true
+
+termux_step_pre_configure() {
+	if [[ ${TERMUX_ON_DEVICE_BUILD} = true ]]; then
+		export JAVA_HOME=${TERMUX__PREFIX__LIB_DIR}/jvm/java-21-openjdk
+	elif [[ ! -v ANT_HOME ]]; then
+		termux_download_ubuntu_packages ant
```
robertkirkman: This should either be in `termux_step_host_build()` (to fully efficiently cache it) or the prefix argument (see the comments in the `termux_step_host_build()` function implementation source file) to move the download folder. The problem is that during `termux_step_make()`, you need to set this:

```bash
export PATH="$TERMUX_PKG_HOSTBUILD_DIR/ubuntu_packages/usr/bin:$PATH"
```

robertkirkman (changes requested):

tur/ffdec/build.sh
robertkirkman: the reason why `termux_download_ubuntu_packages()` does not set `PATH` or `LD_LIBRARY_PATH` automatically is because in some packages the cross-compilation setup gets very complicated so the installation, rather than just download, of the Ubuntu packages couldn't be unified into a single DRY implementation.

TheMightyArie: Screenshot_20260818-173520_Termux
If the code block you mentioned still needs to go to host build override, I will put it to there before pushing the changes. Otherwise, did I do it right?

robertkirkman: Yes

TheMightyArie: Or this?
Screenshot_20260818-173828_Termux

TheMightyArie:
I hope I changed it properly.

tur/ffdec/build.sh
robertkirkman: no um, don't export variables during `termux_step_host_build` unless you have to, that's not really reliable. instead, only use `termux_download_ubuntu_packages` in `termux_step_host_build`.

TheMightyArie: This is right then?
Screenshot_20260818-175545_Termux

TheMightyArie: Or the one with ANT_HOME check.
Screenshot_20260818-175742_Termux

TheMightyArie: Wait I did it wrong.

TheMightyArie: Screenshot_20260818-175849_Termux

robertkirkman: `termux_step_host_build()` should be defined above `termux_step_pre_configure()`, not after, otherwise yes

TheMightyArie:
Should I add `ant` to `TERMUX_PKG_BUILD_DEPENDS` back?

robertkirkman:
Yes but it failed, so you might need me to help you make this package

tur/ffdec/build.sh
```diff
@@ -0,0 +1,61 @@
...
+TERMUX_PKG_DESCRIPTION="JPEXS Free Flash Decompiler"
+TERMUX_PKG_LICENSE="GPL-3.0"
+TERMUX_PKG_MAINTAINER="@termux-user-repository"
+TERMUX_PKG_VERSION="26.2.1"
+TERMUX_PKG_GIT_BRANCH="version$TERMUX_PKG_VERSION"
+TERMUX_PKG_SRCURL="git+https://github.com/jindrapetrik/jpexs-decompiler"
+TERMUX_PKG_SHA256=bc74ba4610df1da263f012011e1f45e3a8189427ee087f69d95f40c105a07d98
+TERMUX_PKG_DEPENDS="openjdk-21, openjdk-21-x"
+TERMUX_PKG_AUTO_UPDATE=false
+TERMUX_PKG_BUILD_IN_SRC=true
```
robertkirkman, line 11: One of the problems is that you didn't set `TERMUX_PKG_HOSTBUILD=true` here yet

TheMightyArie: Oh!

TheMightyArie:
I also could remove the SHA256 option since we are cloning the repository for source code.

tur/ffdec/build.sh
```diff
@@ -0,0 +1,61 @@
+TERMUX_PKG_HOMEPAGE=https://github.com/jindrapetrik/jpexs-decompiler
+TERMUX_PKG_DESCRIPTION="JPEXS Free Flash Decompiler"
+TERMUX_PKG_LICENSE="GPL-3.0"
+TERMUX_PKG_MAINTAINER="@termux-user-repository"
+TERMUX_PKG_VERSION="26.2.1"
+TERMUX_PKG_GIT_BRANCH="version$TERMUX_PKG_VERSION"
+TERMUX_PKG_SRCURL="git+https://github.com/jindrapetrik/jpexs-decompiler"
```
robertkirkman, line 7: Does this work instead?

```
TERMUX_PKG_SRCURL="https://github.com/jindrapetrik/jpexs-decompiler/archive/refs/tags/version$TERMUX_PKG_VERSION.tar.gz"
```

TheMightyArie: NoName suggested me to get the source as cloned repository but sure!

robertkirkman: does it need the git repository for something?

TheMightyArie: Uh, no? I don't even think there are submodules.

robertkirkman: it built successfully without git clone, so this is preferable.

robertkirkman:
Ok so it's platform independent, right, because it's Java? or are there any native components inside Jar file?

tur/ffdec/build.sh
```diff
@@ -0,0 +1,61 @@
...
+	ant build
+}
+
+termux_step_make_install() {
+	mkdir -p $TERMUX__PREFIX/share/java
+	cp -r ./dist/ $TERMUX__PREFIX/share/java/ffdec/
+	cat << EOF > $TERMUX__PREFIX/bin/ffdec
+#!$TERMUX_PREFIX/bin/sh
+
+java -jar $TERMUX__PREFIX/share/java/ffdec/ffdec.jar "\$@"
```
robertkirkman, line 37: please put the `ffdec.jar` under the root of `$TERMUX_PREFIX/share/java` like all other Jar packages in Termux

```
~ $ apt-file search $PREFIX/share/java
apksigner: /data/data/com.termux/files/usr/share/java/apksigner.jar
apktool: /data/data/com.termux/files/usr/share/java/apktool.jar
bdsup2sub: /data/data/com.termux/files/usr/share/java/bdsup2sub.jar
d8: /data/data/com.termux/files/usr/share/java/d8.jar
ecj: /data/data/com.termux/files/usr/share/java/android-24.jar
ecj: /data/data/com.termux/files/usr/share/java/android.jar
jadx: /data/data/com.termux/files/usr/share/java/jadx-1.5.5-all.jar
libbcprov-java: /data/data/com.termux/files/usr/share/java/bcprov.jar
libcommons-lang3-java: /data/data/com.termux/files/usr/share/java/commons-lang3.jar
pdftk: /data/data/com.termux/files/usr/share/java/pdftk.jar
plantuml: /data/data/com.termux/files/usr/share/java/plantuml.jar
procyon-decompiler: /data/data/com.termux/files/usr/share/java/procyon-decompiler.jar
```

TheMightyArie:
There are Shell scripts to launch FFdec but it has commands Termux wouldn't like. So, I have created my own executable that directly calls Java JAR launch command.

tur/ffdec/build.sh
TheMightyArie: Yeah but I have suspects of the JAR won't be able to find the libraries when it's not in the same folder with them.

tur/ffdec/build.sh
robertkirkman: oh never mind I thought you already had built it with that `TERMUX_PKG_SRCURL` i sent, but you didn't change that yet

tur/ffdec/build.sh
robertkirkman: Ok I checked you are actually right it does require that stuff

robertkirkman:
All of the files are `.swc`, `.jar` or `.sh` files so it's platform independent, right?

TheMightyArie:
> All of the files are `.swc`, `.jar` or `.sh` files so it's platform independent, right?

What do you mean exactly?

TheMightyArie:
Oh, I think I get it.
It's Java, so it should be platform independent.

robertkirkman:
What I mean is  if this package is built once will it work on all architectures like 32-bit ARM, 64-bit ARM without building _and_ hosting a separate package for each architecture?

If so, then `TERMUX_PKG_PLATFORM_INDEPENDENT=true` can be set in it to save space... maybe, if TUR supports `TERMUX_PKG_PLATFORM_INDEPENDENT=true`.

robertkirkman:
Ok it requires building from git, so you can change it back to the `TERMUX_PKG_SRCURL` that has git.

tur/ffdec/build.sh
TheMightyArie: I remember why I used Git cloning for FFdec! Compilation will fail for missing Git tags if you download it as archive.

tur/ffdec/build.sh
TheMightyArie: So I revert the change I just made for that?

robertkirkman (approved):

tur/ffdec/build.sh
```diff
@@ -0,0 +1,61 @@
+TERMUX_PKG_HOMEPAGE=https://github.com/jindrapetrik/jpexs-decompiler
+TERMUX_PKG_DESCRIPTION="JPEXS Free Flash Decompiler"
+TERMUX_PKG_LICENSE="GPL-3.0"
+TERMUX_PKG_MAINTAINER="@termux-user-repository"
+TERMUX_PKG_VERSION="26.2.1"
+TERMUX_PKG_GIT_BRANCH="version${TERMUX_PKG_VERSION}"
+TERMUX_PKG_SRCURL="git+https://github.com/jindrapetrik/jpexs-decompiler"
+TERMUX_PKG_DEPENDS="openjdk-21, openjdk-21-x"
+TERMUX_PKG_AUTO_UPDATE=false
```
robertkirkman: Can auto update work?

(this is asking you to try to enable the auto update and research if it will work on this package or not by starting your own study of the termux auto update code)

TheMightyArie: Not sure, we can try.

robertkirkman: Ok, so currently, it's showing this,

image

this means that it would attempt to auto update the package to a version "3544" immediately after merging,

it seems to happen because the project has nightly versions,

image

so we will need to filter out the nightly versions and only detect the stable versions as auto updates, there are a lot of examples of how to do this in several ways in various packages in the repository and you can research some of them to find enough until you can think of a way to adapt one of them to this package.

robertkirkman: TUR and termux-packages use almost exactly the same auto updating system source code, with only minor differences in the TUR fork, and running on two different GitHub accounts, so two instances running in parallel, one for TUR and one for the termux-packages repository, and each one only auto updates packages in the repository it is for. All syntax and available variables and functions for auto updating are the same between the TUR and termux-packages repositories.

TheMightyArie: Can build scripts run `git` commands, especially in `termux_step_get_source`?

robertkirkman: I think so but there is probably a way to do this that is shorter than using a `git` command

TheMightyArie: Is this even a valid variable? I could actually use this...?
Screenshot_20260818-215819_Chrome

TheMightyArie: Pretty sure using `TERMUX_PKG_UPDATE_TAG_TYPE="latest-release-tag"` will do the trick.

PR #2733, opened 2026-08-18
Update

stsp:


tur/thunk-gen/build.sh
```diff
@@ -3,54 +3,30 @@ TERMUX_PKG_DESCRIPTION="thunk generator for C and assembler"
...
-	termux_step_configure_meson
-}
-
-termux_step_make() {
-	$TERMUX_MESON compile --verbose -C $TERMUX_PKG_BUILDDIR
-}
-
-termux_step_make_install() {
-	$TERMUX_MESON install -C $TERMUX_PKG_BUILDDIR
+	pip install --no-deps --no-build-isolation --prefix "$TERMUX_PREFIX" ply
```
robertkirkman: If you write this instead of this, does it work?

```bash
TERMUX_PKG_PYTHON_RUNTIME_DEPS="ply"
```

stsp: Noted, thanks.
Why it now complains on missing
CCTERMUX_HOST_PLATFORM?
Is it needed for python?

robertkirkman: `CCTERMUX_HOST_PLATFORM` is not available during `termux_step_host_build()`; it normally is available starting in `termux_step_pre_configure()`.

If you have to, you can check the package `virglrenderer-android` to see an example of how `CCTERMUX_HOST_PLATFORM` can be used from `termux_step_host_build()`.

robertkirkman: but it seems to me that things which need `CCTERMUX_HOST_PLATFORM` probably should be built during `termux_step_pre_configure()`, `termux_step_configure()`, and `termux_step_make()`, not `termux_step_host_build()`; but I'm just guessing and I'm not familiar with the `thunk-gen` package currently.

stsp: CCTERMUX_HOST_PLATFORM actually
doesn't matter, confusion on my side.
What matters is this:
https://github.com/termux-user-repository/tur/actions/runs/32165110358/job/95802768057?pr=2733
```
/home/builder/termux-packages/scripts/build/setup/termux_setup_meson.sh: line 53: AR: unbound variable
```

As the result, I returned the unneeded
block that happens to contain CCTERMUX_HOST_PLATFORM
for no good.

stsp: So why does it need AR and the
alike variables?

robertkirkman: I need to go soon but I can study the `thunk-gen` package when I come back; I don't understand why you removed so many lines from it in this PR.

stsp: It needed C compiler, but now its
in python. So I wanted to remove
also
```
        AR=$(command -v llvm-ar)
        CC=$(command -v $CCTERMUX_HOST_PLATFORM-clang)
        CXX=$(command -v $CCTERMUX_HOST_PLATFORM-clang++)
        LD=$(command -v ld.lld)
        STRIP=$(command -v llvm-strip)
```
as its entirely unneeded.

robertkirkman: Oh, I see your problem. Thanks for telling it me is now python only, I didn't understand that.

This is a bug/limitation in Termux `termux_setup_meson`

here is the official way to use `termux_setup_meson` in `termux_step_host_build`

https://github.com/termux/termux-packages/blob/02ea22213472bcc54c069cb4e6d39c5e6b104841/packages/libwayland/build.sh#L43-L47

stsp: Got further, but not much.
How can I disable the target build?
This is a build-time only thing, so I
removed various steps, but it still
wants to perform them.
I thought host-only build can be done,
or should I do something else?

robertkirkman: > I thought host-only build can be done,
or should I do something else?

In my opinion, the package organization structure of your packages in Termux might not be ideal.

after some time I will try to study your packages and try to restructure them.

For example, I suspect that it might be more appropriate and stable to combine several of your packages together into a single package that has multiple URLs in a `TERMUX_PKG_SRCURL` array (do you know what that is already?)

stsp: Ok, but this will mean adding
thunk-gen to every package.
Should I do that?

stsp: > multiple URLs in a TERMUX_PKG_SRCURL array (do you know what that is already?)

No, I haven't yet seen an example
of combining them into an array.

stsp: Hmm, it might be possible to just
merge all packages into 1... Not sure
how difficult would that be. Will try
tomorrow.

stsp: I think termux doesn't want to
mix build systems. So if I merge
thunk-gen with other packages
that use non-meson, then termux
gets confused.
Or can it handle multiple build
systems within the package?

stsp:
So the problem with merging
packages seems to be that termux
doesn't handle similar archive names.
TERMUX_PKG_SRCURL=(url1/1.tar.gz url2/1.tar.gz)
It downloads url1/1.tar.gz, then
url2/1.tar.gz and overwrites previous
download, and then things go wrong.

stsp:
Yes, I confirmed the above.
I removed thunk-gen from the
package that needs it, but it was
previously installed by another
package. That avoided the download
clash, and things built.
So yes, we have a genuine termux
bug here.

robertkirkman:
> So the problem with merging packages seems to be that termux doesn't handle similar archive names. TERMUX_PKG_SRCURL=(url1/1.tar.gz url2/1.tar.gz) It downloads url1/1.tar.gz, then url2/1.tar.gz and overwrites previous download, and then things go wrong.

TUR package uploads have been disabled for maintenance unrelated to this, so there is plenty of time to work on this and no rush at the moment. After some time I will study the packages and check what seems to be wrong

stsp:
Ok so to reproduce, you need to
remove the last "debug" patch, and
then in fdpp package you'll get the
clashing download.

robertkirkman:
I'm sorry for delay, yes you are correct, the packages `fdpp` and `thunk-gen` can't be combined together without overriding `termux_step_get_source` because they are incompatible with `termux_download_src_archive`.

I am studying the structure of your packages to see if there are any improvements I could suggest after doing that.

robertkirkman:
@stsp I have analyzed your packages, then:

- I removed everything that looks unnecessary to me,
- and to replace it, instead I restructured the dependency tree for your packages,
- particularly heavily modifying the dependency tree surrounding `thunk-gen`
  - according to these premises that seem not to be factored into the current package dependency structure:
    - **a single `thunk-gen` package can be used for both cross-building and on-device building inside of the Termux App inside of Android** because the `thunk-gen` is now written in Python, which runs anywhere
    - **Meson in termux-packages is supposed to be fully automatic**, i.e. if the proper build setup steps are followed in all cases, "`meson`" and "`meson-python`" should not need to be specified explicitly as build dependencies of anything.
- then I added everything that fixes all errors that I encountered in every build configuration that I have tested,
  - including particularly patches to **avoid the use of absolute path `/tmp`**,
    - because that does not exist on Android, so cannot be used during the building of packages inside of the Termux App
      - (which usually does not occur within GitHub Actions, but can occur if someone using Termux App wants to find and modify the source code of the packages on their own)

these are the changes I suggest, and this patch applies onto the current state of this PR.

[stsp-2.patch](https://github.com/user-attachments/files/31350153/stsp-2.patch)


what do you think of these changes?

stsp:
Ok, thanks.
So:
```
+TERMUX_PKG_DEPENDS="python"
```
in thunk-gen - this likely needs also ply.
Basically you return to the previous (current)
scheme before my combining attempt, and
with an intention to build on device. I very
much doubt someone will build dosemu on
device, but if you like - why not. :)

```
+TERMUX_PKG_PYTHON_COMMON_BUILD_DEPS="ply"
```
(in fdpp)
Since you are back to split scheme, I just
suggest to leave ply for thunk-gen. Moving
it to descendants was only needed for combining.

```
+-XLD_IMB = -Ttext-segment
++XLD_IMB = --image-base
```
(in dj64dev)
If this patch is needed, then there is a bug.
On line 74 makefile already have this:
```
XLD = ld.lld
XLD_IMB = --image-base
```
which is used when llvm is detected.

> I removed everything that looks unnecessary to me,

Please note that it was added after your
suggestion to merge some packages. :)

> including particularly patches to avoid the use of absolute path /tmp,

This is only because you want an on-target
build. No problems, I just never considered
that as a goal.

robertkirkman:
Ok, let's consider each problem in order, starting with the first:

just now after reading your message, when I remove `TERMUX_PKG_PYTHON_COMMON_BUILD_DEPS="ply"` from `dj64dev`, and add it to `thunk-gen`, this error occurs (during the cross build)

```
aarch64-linux-android-clang -fstack-protector-strong -Oz  -fno-common -nostdinc -ggdb3 -O2 -ffreestanding -fno-stack-protector -Wall -Wcast-qual -Werror -Wno-parentheses -Wmissing-declarations -Wmissing-prototypes -Wpointer-arith -Wshadow -Wwrite-strings -Wundef -Wno-array-bounds -Wsign-compare -Wold-style-definition -fpic -flto=auto -ffat-lto-objects     -Wno-unneeded-internal-declaration     -Wno-unknown-warning-option -xc thunks_a.C -c  -o thunks_a.o
/usr/bin/x86_64-linux-gnu-strip --strip-debug ../../../lib/crt0.elf
chmod -x ../../../lib/crt0.elf
Traceback (most recent call last):
  File "/data/data/com.termux/files/usr/share/thunk_gen/thunk_gen.py", line 10, in <module>
    from ply import lex, yacc
ModuleNotFoundError: No module named 'ply'
Traceback (most recent call last):
  File "/data/data/com.termux/files/usr/share/thunk_gen/thunk_gen.py", line 10, in <module>
    from ply import lex, yacc
ModuleNotFoundError: No module named 'ply'
make[3]: *** [/data/data/com.termux/files/usr/share/thunk_gen/thunk_gen.mk:34: thunk_calls.h] Error 1
make[3]: *** Waiting for unfinished jobs....
make[3]: *** [/data/data/com.termux/files/usr/share/thunk_gen/thunk_gen.mk:39: thunk_asms.h] Error 1
make[3]: Leaving directory '/home/builder/.termux-build/dj64dev/build/src/libc/dj64'
make[2]: *** [makefile:19: all] Error 2
make[2]: Leaving directory '/home/builder/.termux-build/dj64dev/build/src/libc'
make[1]: *** [makefile:9: subs] Error 2
make[1]: Leaving directory '/home/builder/.termux-build/dj64dev/build/src'
make: *** [makefile:61: /home/builder/.termux-build/dj64dev/build/lib/libdj64.so] Error 2
```

however, re-adding `TERMUX_PKG_PYTHON_COMMON_BUILD_DEPS="ply"` to `dj64dev` and recompiling it using `scripts/run-docker.sh ./build-package.sh -I -f dj64dev` builds it successfully.

I think I know why this occurs, but what is the reason that it seemed to you that it needs to be moved?

robertkirkman:
Unfortunately, in termux-packages cross-compilation, there is no way to create a build-time nested dependency relationship between a Python build dependency and a PyPi runtime dependency for that build dependency, other than to make a `build.sh` package of the PyPi package, which in this case for example would be `python-ply`, that could be added to `TERMUX_PKG_DEPENDS` of `thunk-gen` to indicate runtime dependency for running `thunk-gen` during builds, but you can let me know what you would prefer to do; I think I would suggest just use `TERMUX_PKG_PYTHON_COMMON_BUILD_DEPS="ply"` in reverse build dependencies of `thunk-gen`.

stsp:
> prefer to do; I think I would suggest just use TERMUX_PKG_PYTHON_COMMON_BUILD_DEPS="ply" in
> reverse build dependencies of thunk-gen.

This is not a matter of preference.
My logic is this: thunk-gen changed
its deps after a python rewrite, but
it is undesirable to break all users at
the same time. So for portability and
modularity we put deps to the package
that brings them.
If this is technically or ideologically
not feasible for termux, then let's do
it the termux way - there is no preference,
just logic. :)

robertkirkman:
> So for portability and
modularity we put deps to the package
that brings them.
If this is technically or ideologically
not feasible for termux, then let's do
it the termux way 

there is a technical barrier in Termux here, but it is not impossible to overcome. I will write the necessary boilerplate to do it, wait a while

robertkirkman:
Ok, here is the other way:

[stsp-2.patch](https://github.com/user-attachments/files/31351937/stsp-2.patch)

This also works, and you can see here that `ply` has been removed from the explicit Python build dependencies of `dj64dev`, `comcom64` and `fdpp` and moved to the runtime dependencies of `thunk-gen`.

however, what actually is still necessary in this configuration is to put `TERMUX_PKG_SETUP_PYTHON=true` in reverse dependencies of `thunk-gen` instead; this variable set to `true` roughly means: "this package has a cross-build and build dependency on a Python dependency that has been pulled into `/data/data/com.termux/files/usr/lib/python3.14/site-packages` from somewhere" - given only that information using `TERMUX_PKG_SETUP_PYTHON=true`, the cross-builder is then able to detect and automatically use `ply` from that location.

let me know what you think of this version.

stsp:
> +TERMUX_PKG_BUILD_DEPENDS="djstub, djstub-cross, python, thunk-gen"

Why such a disparity, djstub has
both versions, but thunk-gen - not?

> +-XLD_IMB = -Ttext-segment
> ++XLD_IMB = --image-base

Just a reminder. :)

> +TERMUX_PKG_BUILD_DEPENDS="binutils, nasm, python, thunk-gen"

(fdpp)
So I guess python is redundant here?
Also note that binutils are special here:
the dependency (here and in some
other packages) on host-binutils, namely
on x86 ones. So if you aim for on-target
build, I don't think you need to add binutils -
it will then find llvm. And they were added
by you, not by me - what was the reason?

The idea to provide ply, is interesting.

robertkirkman:
>> +-XLD_IMB = -Ttext-segment
>> ++XLD_IMB = --image-base

> Just a reminder. :)

yes, let's focus on fixing one part at a time so that the changes don't get too confusing; in the last post, I only changed things aimed towards resolving the `ply` dependency concern.

>  +TERMUX_PKG_BUILD_DEPENDS="djstub, djstub-cross, python, thunk-gen"
> Why such a disparity, djstub has
> both versions, but thunk-gen - not?

could you help me learn about `djstub`,

to start with, is there any active native code in the software, or is it also now purely Python, Shell and unused code?

For example, what is `stub.exe`? is that used or usable in either Termux, or the Ubuntu cross-builder? When I try to run it in the Ubuntu cross-builder like this, an error occurs, is that supposed to happen?

```
builder@04e5e76905d1:~/termux-packages$ /data/data/com.termux/files/usr/opt/djstub/cross/share/djstub2/stub.exe 
bash: /data/data/com.termux/files/usr/opt/djstub/cross/share/djstub2/stub.exe: cannot execute binary file: Exec format error
builder@04e5e76905d1:~/termux-packages$ 
```

stsp:
> to start with, is there any active native code in the software, or is it also now purely Python, Shell and
> unused code?

Its fully host-independent.
You can consider it as written in python.

> For example, what is stub.exe? is that used or usable in either Termux, or the Ubuntu cross-builder?

It can be used everywhere: the python
script uses it during linking of DOS executables.
DOS executables may be considered not
cross-platform, but in our case they actually
are cross-platform: they are executed only
inside an emulator.
So in essence, stub.exe is a part of cross-compiler
and not a runnable entity.

robertkirkman:
Ok, thank you for explaining; in this case, you are correct that there is a discrepancy between how `thunk-gen` is treated in my changes and how `djstub` is treated; if possible, `djstub-cross` should be removed and a single copy of `djstub` run for both on-device builds and cross-compilation builds. I will try to implement that now.

robertkirkman:
It might sound a bit unexpected, but there actually is a reason why removing this:

```bash
TERMUX_PKG_NO_SHEBANG_FIX_FILES="
opt/${TERMUX_PKG_NAME}/cross/bin/*
"
```

has a chance to actually work; it is because of this line:

https://github.com/termux/termux-packages/blob/54008619a4e4c71a16d7697c3eb12bcfb8f38f03/scripts/build/termux_step_override_config_scripts.sh#L9

making `/data/data/com.termux/files/usr/bin/sh` executable during cross-compilations. Far from everything in `/data/data/com.termux/files/usr/bin` can be executed that way and for many programs it's not a good solution, but for Shell programs running with `sh`, it may work. I will be checking it.

robertkirkman:
Ah I see, so, @stsp `djstub` provides `djlink`, and the build of `djlink` for running inside Termux has shebang `#!/data/data/com.termux/files/usr/bin/env bash`,

```
output $ cat ./data/data/com.termux/files/usr/bin/djlink 
#!/data/data/com.termux/files/usr/bin/env bash
```

unfortunately that would create a bit too much of a mess trying to run directly from the Ubuntu cross-builder, so it's better to have `djstub-cross` for that providing scripts with shebang `#!/usr/bin/env bash`. Does this sufficiently explain why `djstub-cross` is necessary to have, but not `thunk-gen-cross`?

stsp:
Wait a minute, also python has shebang! :)
What's the difference?

robertkirkman:
> Wait a minute, also python has shebang! :) What's the difference?

I see, you're right about that. I assumed that because `thunk-gen` did not install anything into `/data/data/com.termux/files/usr/bin`, that it was being run using this command: `python -m thunk_gen` (an invocation method of Python that does not require shebangs) however, I can see that is a wrong assumption because actually it is installed here: `/data/data/com.termux/files/usr/share/thunk_gen/thunk_gen.py` and it is probably run from there directly, not using `python -m`. This means that actually `thunk-gen-cross` is necessary, and I will add it back.

stsp:
> This means that actually thunk-gen-cross is necessary

Well, to me it means that you may
need to seriously reconsider the
architecture. :) Turning cross-platform
solutions into platform-dependent
just because of shebang...

robertkirkman:
> > This means that actually thunk-gen-cross is necessary
> 
> Well, to me it means that you may need to seriously reconsider the architecture. :) Turning cross-platform solutions into platform-dependent just because of shebang...

Somewhat, but there is still something I don't understand about `thunk-gen`; could you explain why in the current version in the repository, the pkg-confg file for `thunk-gen-cross` needs to be installed into a folder that is named after the `$TERMUX_HOST_PLATFORM`, instead of the default folder for that prefix, which then requires that variable to also be used to construct the `PKG_CONFIG_PATH` of all reverse dependencies during cross-building?

https://github.com/termux-user-repository/tur/blob/2f7b0334df70c382c5b19850740ed37b613b22ed/tur/thunk-gen/build.sh#L39

I would want to also remove that, but I should ask you to find out if there is a reason you know changing that folder would break it, because it doesn't look to me like it would. I would like to set `TERMUX_PKG_PLATFORM_INDEPENDENT=true` on `thunk-gen` and `thunk-gen-cross`, but in order to do so without creating nondeterministic behavior, I need to remove the use of `$TERMUX_HOST_PLATFORM` during setting the installation folders of `thunk-gen-cross`.

stsp:
No, this was before the python rewrite.
My current patch does this:
```
export PKG_CONFIG_PATH="${_PREFIX_FOR_THUNK_GEN}/share/pkgconfig"
```

robertkirkman:
Oh I see, thank you. Now I have made sure to put back the use of `thunk-gen-cross` as it is in this PR's current state, but with the `thunk-gen` and `thunk-gen-cross` packages restored, while also retaining the changes I made for `ply`:

[stsp-2.patch](https://github.com/user-attachments/files/31354014/stsp-2.patch)

If you check this, the remaining concerns you had are these, right?

- `fdpp`: marking a build dependency on `python` might be redundant and unnecessary, if reducing redundant dependencies is desired

- `fdpp`: I need to explain why it depends on `binutils` at build-time, and discuss potential alternative solutions to the problem

- `dj64dev`: the patch applying `--image-base` is messy and the root cause of the problem still needs to be found

do I understand the remaining concerns correctly?

stsp:
> fdpp: marking a build dependency on python might be redundant and unnecessary

Yes, as you have already found the
way to pull in thunk-gen deps into fdpp.
python comes only from that.

> fdpp: I need to explain why it depends on binutils at build-time, and discuss potential alternative
> solutions to the problem

Since my patch doesn't have it, I don't
think there are a big problems without.
On termux it is preferable to use llvm
instead of binutils, because llvm supports
all arches at once, while binutils must
be specially configured for cross-build.

> dj64dev: the patch applying --image-base is messy and the root cause of the problem still needs to
> be found

Indeed, although would be good to know
what was the problem. If it actually was, it
is likely then an upstream problem.

robertkirkman:
Ok, I agree it's fine to remove `python` from the build dependencies of `fdpp` since `thunk-gen` is there. (Just now I have remembered to also add `python-ply` to `TERMUX_SUBPKG_DEPENDS` of `thunk-gen-cross` as well, since that one does also depend on it since it has `TERMUX_SUBPKG_DEPEND_ON_PARENT=false`).

Let's discuss `binutils` in the build dependencies of `fdpp`. The reason why I added it there is because there are uses of the command `ld.bfd` explicitly within the build system of `fdpp`. 

https://github.com/dosemu2/fdpp/blob/c28781b601534cceeb4de887c02cc9ed57e746e5/fdpp/clang.mak#L8

During cross-compilation, I assume that `/usr/bin/ld.bfd` provided by Ubuntu is used, since it is present, but during on-device build, there is no guarantee that `ld.bfd` will be present without the package `binutils`, which provides it, being specified as a build dependency. This causes this error to occur during on-device build of `fdpp` if `binutils` is not added to `TERMUX_PKG_BUILD_DEPENDS`:

```
ld.bfd -melf_i386 -static -Map=../fdppkrnl.35.10.map -o ../fdppkrnl.35.10.elf \
	--emit-relocs  -T/data/data/com.termux/files/home/.termux-build/fdpp/src/fdpp/kernel/kernel.ld kernel.o entry.o io.o console.o serial.o printer.o execrh.o nlssupt.o procsupt.o dosidle.o int2f.o nls_hc.o intr.o irqstack.o cpu.o plt.o cdata.o floppy.o rdpcclk.o wrpcclk.o wratclk.o
bash: line 1: ld.bfd: command not found
make[2]: *** [makefile:61: ../fdppkrnl.35.10.elf] Error 127
make[2]: Leaving directory '/data/data/com.termux/files/home/.termux-build/fdpp/build/fdpp/kernel'
make[1]: *** [makefile:69: fdppkrnl.35.10.elf] Error 2
make[1]: *** Waiting for unfinished jobs....
```

Another option, that I just now tested which does also work around that error without having to have `binutils` installed, is to patch it to use `ld.lld` instead, which is the linker preferred during on-device building of packages (and also preferred during cross-compilation for Termux):

```diff
--- a/fdpp/clang.mak
+++ b/fdpp/clang.mak
@@ -5,7 +5,7 @@
 CROSS_LD ?= x86_64-linux-gnu-ld
 ifeq ($(shell $(CROSS_LD) --version 2>/dev/null),)
 ifneq ($(filter x86_64 amd64 i686,$(shell uname -m)),)
-CROSS_LD = ld.bfd
+CROSS_LD = ld.lld
 else
 $(error binutils-x86-64-linux-gnu not installed)
 endif
```

however, I wonder whether this is something that should be patched some other way, if you think that this codepath with `ld.bfd` should not be entered during on-device build.

The Android device that I am running Termux App in to test these builds has a x86_64 processor, not an aarch64 processor. Could that be affecting the behavior of the makefiles of `fdpp` and potentially other packages among these?

(when developing on-device build support for packages, I do like to ensure that all architectures of Android are calculated equally in the code of the packages, without any failures occurring only on architecture combinations with Android that are less common)

stsp:
> but during on-device build, there is no guarantee that ld.bfd will be present

Right, so for that scenario you need:
`make CROSS_LD=ld.lld` instead of the dep.

> (when developing on-device build support for packages, I do like to ensure that all architectures of
> Android are calculated equally in the code of the packages, without any failures occurring only on
> architecture combinations with Android that are less common)

Which is why I limited myself to host-only.
But the build infrastructure is able to take
the challenge.

robertkirkman:
Does this look acceptable to you? This is only necessary during on-device build, but it also works during cross-compilation, so it's fine to apply it unconditionally, right?

```bash
TERMUX_PKG_HOMEPAGE=https://github.com/dosemu2/fdpp
TERMUX_PKG_DESCRIPTION="64-bit DOS core"
TERMUX_PKG_LICENSE="GPL-3.0"
TERMUX_PKG_MAINTAINER="@stsp"
TERMUX_PKG_VERSION="1.11"
TERMUX_PKG_SRCURL="https://github.com/dosemu2/fdpp/archive/refs/tags/${TERMUX_PKG_VERSION[0]}.tar.gz"
TERMUX_PKG_SHA256="66a66103847ed33167c9d869c9272d6a96613489b381060e61f84e385e8a3db5"
TERMUX_PKG_SETUP_PYTHON=true
TERMUX_PKG_DEPENDS="libelf"
TERMUX_PKG_BUILD_DEPENDS="nasm, thunk-gen, thunk-gen-cross"

termux_step_pre_configure() {
	if [[ "$TERMUX_ON_DEVICE_BUILD" == "false" ]]; then
		local _PREFIX_FOR_THUNK_GEN="${TERMUX_PREFIX}/opt/thunk-gen/cross"
		export PKG_CONFIG_PATH="${_PREFIX_FOR_THUNK_GEN}/share/pkgconfig"
	fi

	TERMUX_PKG_EXTRA_MAKE_ARGS="CROSS_LD=$LD"
}
```

(`$LD` is the appropriate variable referencing `ld.lld` in termux-packages, but it is not set in global scope, only after `termux_step_setup_toolchain()` has run, so this is the typical way a custom variable for the linker is set in termux-packages)

stsp:
Yes, that should work.
Makefiles should be able to figure out
what ld is that, and act accordingly.
I am not sure if you need libelf in build
deps too, but you know what you do.

robertkirkman:
> I am not sure if you need libelf in build
deps too, but you know what you do.

in termux-packages, if a package is in `TERMUX_PKG_DEPENDS`, it is calculated as a build dependency as well as a runtime dependency, so it is not necessary to also specify it in `TERMUX_PKG_BUILD_DEPENDS` unless there is a desire to specifically show for clarity which dependencies are runtime-only dependencies and which dependencies are both runtime and build-time dependencies.


Finally, regarding `--image-base`, if I remove the patch from `dj64dev`, during the on-device build of `comcom64` later on, this error occurs:

```
nl -v0 <thunk_c.h | sed -E 's/^ *//' >thunk_calls.tmp
(/data/data/com.termux/files/usr/share/thunk_gen/thunk_gen.py -a 4 -p 4 <thunk_calls.tmp >thunk_calls.h) || (rm -f thunk_calls.h ; false)
/data/data/com.termux/files/usr/share/thunk_gen/parse_decls.sh 3 thunk_calls.tmp >plt.inc || (rm -f plt.inc ; false)
set -o pipefail; echo "#include <dj64/plt.S.inc>" | \
    x86_64-linux-android-cpp -x assembler-with-cpp -I.  -I. -I/data/data/com.termux/files/home/.termux-build/comcom64/src/src -I/data/data/com.termux/files/usr/i386-pc-dj64/include -DDJ64 -nostdinc - | \
    x86_64-linux-android-clang -x assembler -m32 -Wa,-defsym,_DJ64=1 -c  -I. -I/data/data/com.termux/files/home/.termux-build/comcom64/src/src -o plt.o -
x86_64-linux-android-clang  -fstack-protector-strong -Oz -Wall -O2 -Wmissing-declarations -Wwrite-strings -ggdb3 -Wunused -Wmissing-prototypes -DREV_ID=\"Non-git_build\" -fpic -ffreestanding -fno-stack-protector  -isystem/data/data/com.termux/files/usr/include/c++/v1 -isystem/data/data/com.termux/files/usr/include -I. -I/data/data/com.termux/files/home/.termux-build/comcom64/src/src -DCOMCOM_VERSION=\"0.5\" -I/data/data/com.termux/files/usr/i386-pc-dj64/include -DDJ64 -nostdinc -imacros /data/data/com.termux/files/usr/i386-pc-dj64/include/dj64/a_mac.h -I. -o thunks_c.o -c /data/data/com.termux/files/home/.termux-build/comcom64/src/src/thunks_c.c
ld int21.o int21_next.o int23.o int0.o int75.o asm.o ms.o term.o plt.o -melf_i386 -static /data/data/com.termux/files/usr/i386-pc-dj64/lib/uplt.o -Ttext-segment=0x08068000 -o tmp.o.elf
ld: error: -Ttext-segment is not supported. Use --image-base if you intend to set the base address
make[1]: *** [/data/data/com.termux/files/usr/i386-pc-dj64/share/dj64.mk:149: tmp.o.elf] Error 1
make[1]: *** Waiting for unfinished jobs....
rm tmp.s
make[1]: Leaving directory '/data/data/com.termux/files/home/.termux-build/comcom64/build/src'
make: *** [makefile:49: 64] Error 2
```

during debugging, I was able to use temporary `$(shell echo "test... " >> ~/test.log)` lines to determine that this point of the code in the `dj64.mk` is reached:

image

with this information, does anything in particular stand out to you that should be done, or what else should be checked to determine how to find the root cause? Should this point in the code be reached during on-device building?

stsp:
Ah, I see!
We don't set CROSS_PREFIX, but even
if we did, it assumes CROSS_PREFIX is
only for GNU tool-chain. I guess this is
why we already have a `Darwin` work-around,
so I suppose something like this may help:
```
--- a/dj64.mk
+++ b/dj64.mk
@@ -7,9 +7,11 @@ OS = $(shell uname -s)
 NATIVE=0
 ifneq ($(filter x86_64 amd64 i686 i586 i386,$(MACH)),)
 ifneq ($(OS),Darwin)
+ifeq ($(filter android%,$(MACH)),)
 NATIVE = 1
 endif
 endif
+endif
 ifeq ($(OS),SunOS)
 NATIVE_PREFIX := g
 endif
```
It will do the similar "exception" we already
had for Darwin.

robertkirkman:
Yes, I have just tested that, and that is working. Thank you!

here are all the discussed changes as a patch on this PR: 
[stsp-2.patch](https://github.com/user-attachments/files/31355556/stsp-2.patch)

robertkirkman:
I called the patch `dj64nk-on-device.patch`, but that is a typo, I meant to type `dj64mk-on-device.patch`, and you should change the name to what you think is appropriate if you apply these changes.

robertkirkman:
Additionally, I had not yet added `TERMUX_PKG_REVISION=1` onto all modified packages where needed, so that is a last step, if all the changes will be applied.

stsp:
> +TERMUX_PKG_DEPENDS="libelf, pkg-config"
> +TERMUX_PKG_BUILD_DEPENDS="ctags, ctags-cross, libelf, python, thunk-gen, thunk-gen-cross"

Although in this case you have
double libelf.

So if I am to apply this (tomorrow),
then I also need to know what is this
`command -v` patch. If you, then I
don't need full details.

robertkirkman:
> Although in this case you have
double libelf.

ah yes, `libelf` does not technically need to be in the `TERMUX_PKG_BUILD_DEPENDS` for `dj64dev` also while it is in `TERMUX_PKG_DEPENDS`.

> So if I am to apply this (tomorrow),
then I also need to know what is this
`command -v patch`. If you, then I
don't need full details.

yes, the `which` command is not guaranteed to be available during on-device building, only `command -v`,

and recently, I requested permission to add `which` to the guaranteed preinstalled packages for on-device building, however, my request was rejected,

because of that, I am planning to instead _remove_ `which` from the Ubuntu cross-builder, in order to synchronize the build dependency assumptions of on-device builds with cross-compilations more closely, so while the error doesn't occur during cross-compilations now, it will soon if this is not patched. Permission has been granted for me to remove `which` from the Ubuntu cross-builder.

- https://github.com/termux/termux-packages/pull/31123

stsp:
If this means carrying a lot of
patches (not just 1 here), then I'd
rather add `which` to tur as a start.

stsp:
Ok, looking into the above link -
seems like they are not many. :)

robertkirkman:
Yes, among these packages, only `comcom64` needed a patch for replacing `which` with `command -v` in my test.

stsp:
Done and builds.
Please review.

robertkirkman (approved):
Thank you!

Issue #2739, opened 2026-08-19
[Bug]: FFdec has a hidden installation problem


TheMightyArie:
### Problem description

FFdec is already added to TUR, but somehow it can't be installed with 404 HTTP code. 
Output:
<img width="1080" height="464" alt="Image" src="https://github.com/user-attachments/assets/dccb263b-c891-438e-a2c5-300bb865971f" />

### What steps will reproduce the bug?

`pkg i ffdec` is enough. I already tried updating the packages and running `termux-change-repo`.

### What is the expected behavior?

The package is supposed to be installed without any issues.

### System information

termux-info:

```
Termux Variables:
TERMUX_API_VERSION=0.53.0
TERMUX_APK_RELEASE=GITHUB
TERMUX_APP_PACKAGE_MANAGER=apt
TERMUX_APP_PID=26737
TERMUX_APP__DATA_DIR=/data/user/0/com.termux
TERMUX_APP__LEGACY_DATA_DIR=/data/data/com.termux
TERMUX_APP__SE_FILE_CONTEXT=u:object_r:app_data_file:s0:c0,c257,c512,c768
TERMUX_APP__SE_INFO=default:targetSdkVersion=28:complete
TERMUX_IS_DEBUGGABLE_BUILD=1
TERMUX_MAIN_PACKAGE_FORMAT=debian
TERMUX_VERSION=0.118.3
TERMUX__HOME=/data/data/com.termux/files/home
TERMUX__PREFIX=/data/data/com.termux/files/usr
TERMUX__ROOTFS_DIR=/data/data/com.termux/files
TERMUX__SE_PROCESS_CONTEXT=u:r:untrusted_app_27:s0:c0,c257,c512,c768
TERMUX__USER_ID=0
Packages CPU architecture:
aarch64
Subscribed repositories:
# sources.list
deb https://packages-cf.termux.dev/apt/termux-main stable main
# root-repo (sources.list.d/root.list)
deb https://packages-cf.termux.dev/apt/termux-root root stable
# tur-repo (sources.list.d/tur.list)
deb https://tur.kcubeterm.com tur-packages tur tur-on-device tur-continuous
# x11-repo (sources.list.d/x11.list)
deb https://packages-cf.termux.dev/apt/termux-x11 x11 main
Updatable packages:
All packages up to date
termux-tools version:
1.45.0
Android version:
16
Kernel build information:
[...]

PR #2742, opened 2026-08-20
revbump(tur/ffdec): Re-build attempt

TheMightyArie:
- Possible fix for #2739


robertkirkman:
Ok hold on if I merge this as is it's going to cause GitHub to close that issue but preferably that issue should stay open until the race condition is prevented permanently for all packages so could you change the description and commit to something else that isn't "Fixes"

TheMightyArie:
I actually did that before but then changed it to "Fixes". Reverted it back anyways.

licy183:
Emmm, it will close #2739, but not termux-user-repository/dists#9, so I don't think it matters.

robertkirkman (approved):

PR #2745, opened 2026-08-21
[WIP] addpkg(tur/tiemu): 3.03, addpkg(tur/tilp): 1.19

3ls-it:
##### TI Utils
  
This PR adds the TI calculator utilities TiEmu and TiLP2 and their dependencies.

###### TiEmu v3.03  
  
TiEmu is an emulator for TI89/TI89-Titanium/TI92/TI92+/V200PLT calculators.  
  
- The upstream TiEmu 3.03 distribution includes legacy PedroM 0.81 `.tib` images, which are not compatible with the current TiEmu version. The current PedroM 0.83 distribution provides valid `.89u`/`.9xu` FLASH images, which are installed in TiEmu’s existing PedRom search directory. The interface of the PedroM OS is a Unix-like shell with familiar commands like: `ls`, `mkdir`, `cd`, etc.  
  
###### TiLP2 v1.18 
  
TiLP is a linking program for Texas Instruments' graphing calculators, which allows for moving files to and from the device, making backups, OS update/upgrades, screenshots, etc.
  
----

TI-89 Titanium emulation with TI AMS OS v3.10:  
  
tiemu_ti89t_TiAMS  
   
  
TI-Voyage200 emulation with PedroM OS v0.83:  
  
tiemu_v200_PedroM83  
  
TiLP2 interface:  
tilp2

### DIFF
--- a/tur/libtifiles2/build.sh
+++ b/tur/libtifiles2/build.sh
@@ -1,0 +1,13 @@
+TERMUX_PKG_HOMEPAGE="http://lpg.ticalc.org/prj_tilp/"
+TERMUX_PKG_DESCRIPTION="Provides libtifiles2 library and headers for TiEmu and TiLP"
+TERMUX_PKG_LICENSE="GPL-2.0"
+TERMUX_PKG_MAINTAINER="3ls-it <3ls-it@pm.me>"
+TERMUX_PKG_VERSION=1.1.7
+TERMUX_PKG_SRCURL="https://sourceforge.net/projects/tilp/files/tilp2-linux/tilp2-1.18/libtifiles2-${TERMUX_PKG_VERSION}.tar.bz2"
+TERMUX_PKG_SHA256=9ac63b49e97b09b30b37bbc84aeb15fa7967bceb944e56141c5cd5a528acc982
+TERMUX_PKG_AUTO_UPDATE=false
+TERMUX_PKG_DEPENDS="glib, libarchive, libticonv"
+
+termux_step_pre_configure() {
+		autoreconf -fi
+}
--- a/tur/tilp2/tilp
+++ b/tur/tilp2/tilp
@@ -1,0 +1,46 @@
+#!/data/data/com.termux/files/usr/bin/sh
+# This script was installed as part of the `tilp2` package
+
+set -eu
+
+real="$PREFIX/bin/tilp.real"
+
+if [ ! -x "$real" ]; then
+    printf '%s\n' "tilp: real executable not found: $real" >&2
+    exit 1
+fi
+
+# Permit direct execution when already inside termux-usb.
+if [ -n "${TERMUX_USB_FD:-}" ]; then
+    exec "$real" "$@"
+fi
+
+# Optional explicit device override.
+if [ -n "${TILP_USB_DEVICE:-}" ]; then
+    device=$TILP_USB_DEVICE
+else
+    devices=$(
+        termux-usb -l |
+        grep -o '/dev/bus/usb/[0-9][0-9]*/[0-9][0-9]*' || :
+    )
+
+    count=$(printf '%s\n' "$devices" | sed '/^$/d' | wc -l)
+
+    case "$count" in
+        0)
+            printf '%s\n' "tilp: no USB devices found" >&2
+            exit 1
+            ;;
+        1)
+            device=$(printf '%s\n' "$devices" | sed -n '1p')
+            ;;
+        *)
+            printf '%s\n' \
+                "tilp: multiple USB devices found; set TILP_USB_DEVICE" >&2
+            printf '%s\n' "$devices" >&2
+            exit 1
+            ;;
+    esac
+fi
+
+exec termux-usb -r -E -e "$real" "$device"
--- a/tur/libglade2/build.sh
+++ b/tur/libglade2/build.sh
@@ -1,0 +1,16 @@
+TERMUX_PKG_HOMEPAGE="https://download.gnome.org/sources/libglade/"
+TERMUX_PKG_DESCRIPTION="Library to load .glade files at runtime"
+TERMUX_PKG_LICENSE="LGPL-2.1"
+TERMUX_PKG_MAINTAINER="3ls-it <3ls-it@pm.me>"
+TERMUX_PKG_VERSION="2.6.4"
+TERMUX_PKG_SRCURL="https://download.gnome.org/sources/libglade/${TERMUX_PKG_VERSION%.*}/libglade-${TERMUX_PKG_VERSION}.tar.gz"
+TERMUX_PKG_SHA256=c41d189b68457976069073e48d6c14c183075d8b1d8077cb6dfb8b7c5097add3
+TERMUX_PKG_AUTO_UPDATE=false
+TERMUX_PKG_DEPENDS="atk, glib, gtk2, libxml2"
+TERMUX_PKG_EXTRA_CONFIGURE_ARGS="
+--disable-static
+"
+
+termux_step_pre_configure() {
+		LDFLAGS+=" -lgmodule-2.0"
+}
--- a/tur/libticables2/0001-disable-build-tests.patch
+++ b/tur/libticables2/0001-disable-build-tests.patch
@@ -1,0 +1,21 @@
+--- a/Makefile.am	2013-04-13 09:26:51.000000000 -0400
++++ b/Makefile.am	2026-07-29 23:07:46.599235725 -0400
+@@ -3,7 +3,7 @@
+ ACLOCAL_AMFLAGS=-I m4
+ 
+ # Subdirectories to scan
+-SUBDIRS = build/mingw po src tests
++SUBDIRS = build/mingw po src
+ 
+ if USE_DOCGEN
+   SUBDIRS += docs
+--- a/configure.ac	2016-03-23 07:38:40.000000000 -0400
++++ b/configure.ac	2026-07-29 23:08:40.951235725 -0400
+@@ -38,7 +38,6 @@
+   src/Makefile
+   src/win32/dha/Makefile
+   src/win64/rwp/Makefile
+-  tests/Makefile
+   ticables2.pc
+ ])
+ 
--- a/tur/libticables2/0002-instrument-slv_bulk_read2.patch
+++ b/tur/libticables2/0002-instrument-slv_bulk_read2.patch
@@ -1,0 +1,878 @@
+--- a/src/linux/link_usb.c	2016-09-10 15:45:34.000000000 -0400
++++ b/src/linux/link_usb.c	2026-08-12 13:16:39.697000000 -0400
+@@ -96,6 +96,20 @@
+ # include <usb.h>
+ #endif
+ 
++/* Termux porting instrumentation */
++#define SLV_TRACE(level, fmt, ...)                                      \
++       do {                                                               \
++               if (usb_debug >= (level)) {                                    \
++                       struct timeval trace_tv;                                  \
++                       gettimeofday(&trace_tv, NULL);                            \
++                       fprintf(stderr, "[%ld.%03ld] slv_bulk_read2: " fmt "\n",  \
++                               (long)trace_tv.tv_sec,                               \
++                               trace_tv.tv_usec / 1000,                              \
++                               ##__VA_ARGS__);                                       \
++               }                                                              \
++       } while (0)
++/* End instrumentation */
++
+ /* --- */
+ 
+ #ifdef __WIN32__	// found in src/error.h of libusb-win32
+@@ -202,6 +216,9 @@
+ /* variables for slv_check and slv_bulk_read2 */
+ static int io_pending = 0;
+ static struct usb_urb urb;
++
++static int trace_after_halt_recovery = 0;
++static int recovery_reads = 0;
+ #endif
+ 
+ /* --- */
+@@ -285,6 +302,7 @@
+ #define uInEnd              (((usb_struct *)(h->priv2))->in_endpoint)
+ #define uOutEnd             (((usb_struct *)(h->priv2))->out_endpoint)
+ 
++
+ /* Helpers (=driver API) */
+ 
+ static void tigl_get_product(char * string, size_t maxlen, struct usb_device *dev)
+@@ -438,20 +456,55 @@
+ {
+ 	int ret;
+ 
++	fprintf(stderr,
++		"tigl_reset: enter io_pending=%d nBytesRead=%d "
++		"in_ep=0x%02x out_ep=0x%02x fd=%d\n",
++		io_pending,
++		nBytesRead,
++		uInEnd,
++		uOutEnd,
++		uHdl != NULL ? uHdl->fd : -1);
++
+ 	// Reset out pipe
+ 	ret = usb_clear_halt(uHdl, uOutEnd);
++
++	fprintf(stderr,
++		"tigl_reset: clear OUT ep=0x%02x ret=%d",
++		uOutEnd, ret);
++
++	if (ret < 0)
++		fprintf(stderr, " errno=%d (%s)", errno, strerror(errno));
++
++	fprintf(stderr, "\n");
++
+ 	if (ret < 0) 
+ 	{
+ 		ticables_warning("usb_clear_halt of out pipe (%s).\n", usb_strerror());
+ 	}
++
+ 	
+ 	// Reset in pipe
+ 	ret = usb_clear_halt(uHdl, uInEnd);
++
++	fprintf(stderr,
++		"tigl_reset: clear IN ep=0x%02x ret=%d",
++		uInEnd, ret);
++
++	if (ret < 0)
++		fprintf(stderr, " errno=%d (%s)", errno, strerror(errno));
++
++	fprintf(stderr, "\n");
++
+ 	if (ret < 0) 
+ 	{
+ 		ticables_warning("usb_clear_halt of in pipe (%s).\n", usb_strerror());
+ 	}
+ 
++	fprintf(stderr,
++		"tigl_reset: exit io_pending=%d nBytesRead=%d\n",
++		io_pending,
++		nBytesRead);
++
+ 	return 0;
+ }
+ 
+@@ -488,6 +541,13 @@
+ 
+ static int slv_open(CableHandle *h)
+ {
++	fprintf(stderr,
++		"slv_open: enter io_pending=%d nBytesRead=%d "
++		"priv2=%p\n",
++		io_pending,
++		nBytesRead,
++		(void *)h->priv2);
++
+ 	int ret;
+ 	int i;
+ 	struct usb_config_descriptor *config;
+@@ -497,6 +557,17 @@
+ 
+ 	// open device
+ 	ret = tigl_open(h->address, &uHdl);
++
++	fprintf(stderr,
++		"slv_open: tigl_open returned %d handle=%p",
++		ret,
++		(void *)uHdl);
++
++	if (ret == 0 && uHdl != NULL)
++		fprintf(stderr, " fd=%d", uHdl->fd);
++
++	fprintf(stderr, "\n");
++
+ 	if (ret)
+ 	{
+ 		return ret;
+@@ -540,19 +611,101 @@
+ 	nBytesRead = 0;
+ 	was_max_size_packet = 0;
+ 
++	fprintf(stderr,
++		"slv_open: exit io_pending=%d nBytesRead=%d "
++		"in_ep=0x%02x out_ep=0x%02x max_ps=%d "
++		"handle=%p fd=%d\n",
++		io_pending,
++		nBytesRead,
++		uInEnd,
++		uOutEnd,
++		max_ps,
++		(void *)uHdl,
++		uHdl != NULL ? uHdl->fd : -1);
++
+ 	return 0;
+ }
+ 
++#if 0
+ static int slv_close(CableHandle *h)
+ {
++
++	fprintf(stderr,
++		"slv_close: enter io_pending=%d nBytesRead=%d "
++		"priv2=%p handle=%p fd=%d\n",
++		io_pending,
++		nBytesRead,
++		(void *)h->priv2,
++		(void *)uHdl,
++		uHdl != NULL ? uHdl->fd : -1);
++
+ 	if (uHdl != NULL) 
+ 	{
++		fprintf(stderr,
++			"slv_close: calling tigl_close handle=%p fd=%d\n",
++			(void *)uHdl,
++			uHdl->fd);
++
+ 		tigl_close(&uHdl);
++
++		fprintf(stderr,
++			"slv_close: tigl_close returned handle=%p\n",
++			(void *)uHdl);
+ 	}
+ 	uDev = NULL; 
++	free(h->priv2);
++	h->priv2 = NULL;
++	io_pending = FALSE;
++
++	fprintf(stderr,
++		"slv_close: exit io_pending=%d priv2=%p "
++		"handle=%p\n",
++		io_pending,
++		(void *)h->priv2,
++		(void *)uHdl);
++
++	return 0;
++}
++#endif
++
++static int slv_close(CableHandle *h)
++{
++	fprintf(stderr,
++		"slv_close: enter h=%p priv2=%p handle=%p\n",
++		(void *)h,
++		(void *)h->priv2,
++		(void *)uHdl);
++
++	if (uHdl != NULL) {
++		fprintf(stderr,
++			"slv_close: before tigl_close handle=%p fd=%d\n",
++			(void *)uHdl,
++			uHdl->fd);
++
++		tigl_close(&uHdl);
++
++		fprintf(stderr,
++			"slv_close: after tigl_close handle=%p\n",
++			(void *)uHdl);
++	}
++
++	fprintf(stderr, "slv_close: before uDev=NULL\n");
++	uDev = NULL;
++	fprintf(stderr, "slv_close: after uDev=NULL\n");
++
++	fprintf(stderr, "slv_close: before free priv2=%p\n",
++		(void *)h->priv2);
+ 
+ 	free(h->priv2);
++
++	fprintf(stderr, "slv_close: after free\n");
++
+ 	h->priv2 = NULL;
++	fprintf(stderr, "slv_close: after priv2=NULL\n");
++
++	io_pending = FALSE;
++	fprintf(stderr, "slv_close: exit io_pending=%d\n",
++		io_pending);
+ 
+ 	return 0;
+ }
+@@ -567,16 +720,39 @@
+ {
+ 	int ret = 0;
+ 
++	fprintf(stderr,
++		"slv_reset: enter io_pending=%d nBytesRead=%d "
++		"fd=%d\n",
++		io_pending,
++		nBytesRead,
++		uHdl != NULL ? uHdl->fd : -1);
++
+ #if !defined(__BSD__)
+ 	/* Reset both endpoints (send an URB_FUNCTION_RESET_PIPE) */
+ 	ret = tigl_reset(h);
++
++	fprintf(stderr,
++		"slv_reset: tigl_reset returned %d\n",
++		ret);
++
+ 	if (!ret)
+ 	{
+ #ifdef __WIN32__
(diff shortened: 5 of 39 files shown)


tur/tigcc/build.sh
```diff
@@ -0,0 +1,280 @@
...
+PATH="$PATH:$TIGCC/bin"
+
+## Custom build functions
+_tigcc_build_binutils() {
+		cd $TERMUX_PKG_SRCDIR/build/binutils
+
+		CFLAGS='-std=gnu89' \
+				../../download/binutils.ti/configure \
+				--build=aarch64-unknown-linux-android24 \
+				--host=aarch64-unknown-linux-android24 \
```
robertkirkman: Is it properly building for all target architectures?

3ls-it: Argh! That's what I get for trying to juggle too many projects at once! :man_shrugging:   
  
When converting my on-device build scripts to a working `build.sh` I had _meant_ to change those to
```sh
--build="$TERMUX_BUILD_TUPLE" \
--host="$TERMUX_HOST_PLATFORM" \
```  
I believe those are the correct variables, neh? I was then going to test all architecture builds in `termux-docker` on my build host... and forgot.

3ls-it: Well, hmm. So builds inside `termux-docker` are considered "on-device"? I get:  
```
~/tur $ ./build-package.sh -I -f -a arm tigcc
ERROR: ./build-package.sh: option '-a' is not available for on-device builds
```  
My only other experience with `termux-docker` was when we were trying to build LibreOffice from your TUR repo fork.  
  
And I see the CI builds just failed again....

robertkirkman: Yes, builds inside `termux-docker` are considered "on-device",

though ideally packages should build both on-device and with cross-compilation, so if you want to test whether the PR will build with cross-compilation locally, you should use these commands instead on a PC that has Docker:

```bash
git clone -b ti-utils https://github.com/3ls-it/tur.git
cd tur
./setup-environment.sh
# the container name in these commands is not a placeholder example;
# if you have never edited this file before: scripts/run-docker.sh
# or used this variable before: $CONTAINER_NAME
# then you should copy and paste the commands exactly to remove the
# older container, if it exists.
docker container kill termux-package-builder
docker container rm termux-package-builder
scripts/run-docker.sh ./build-package.sh -I -f ti-utils
```

3ls-it:
```
2026-08-21T16:34:04.5557149Z m68k-coff-cc  -dumpspecs > tmp-specs
2026-08-21T16:34:04.5563930Z /bin/sh: 1: m68k-coff-cc: not found
```  
Looks like a missing symlink, but I haven't needed that. The completed install `tigcc/bin/` looks like this:  
```
a68k*
ar-tigcc*
as*
cc1*
gcc*
ld-tigcc*
m68k-coff-ar -> /data/data/com.termux/files/usr/tigcc/bin/ar-tigcc*
m68k-coff-as -> /data/data/com.termux/files/usr/tigcc/bin/as*
m68k-coff-ld -> /data/data/com.termux/files/usr/tigcc/bin/ld-tigcc*
patcher*
tigcc*
tigccdoc*
tprbuilder*
ttbin2oth*
ttpack*
```  
It might mean that I need
```
m68k-coff-cc ->  /data/data/com.termux/files/usr/tigcc/bin/cc1
```  
but I'm not going to guess; I need to grep though the original author's build scripts and see what I may have missed. Odd, I just did a build in `termux-docker` with this current `tigcc/build.sh` and it completed just fine.

robertkirkman: > Odd, I just did a build in termux-docker with this current tigcc/build.sh and it completed just fine.

yes, the requirements for building using `scripts/run-docker.sh ./build-package.sh -I -f` are stricter, and if doing that is producing errors, usually they take a significant amount of work to resolve.

3ls-it: Yes, thank you for this:

> docker container kill termux-package-builder
> docker container rm termux-package-builder

I have a Linux build host where I use the `termux-package-bulder`, but didn't know how to use it for TUR packages. Now I do. :smile:   
  
Now I see the same error as in the CI and will work on the fix. Also, I wanted to thank you for your tip about using `termux-usb -r -E -e`, that's actually what got this project going, and it works!

robertkirkman: Awesome!

It look me a very long time to learn exactly how and when I should be using `docker container kill termux-package-builder` and `docker container rm termux-package-builder`, but after I did, I have since been able to reliably build Termux packages from as many different folders on the same computer that I need to, up to an unlimited amount of folders.

modifying the `scripts/run-docker.sh` script to have a different name instead of `termux-package-builder`, different for every different folder, has also really allowed me to build multiple packages on the same PC simultaneously, but that is a bit more of an advanced setup to keep track of.

3ls-it: Status update: I've gotten well  past the `m68k-coff-cc` error, and in to a new error territory. I'll get this sorted.

3ls-it: Well, there was a "missing" symlink, which is not needed in  an on-device/termux-docker build, but that was just the opening to the rabbit hole. For this PR, I've removed TGICC.

robertkirkman: > that was just the opening to the rabbit hole

yes, there can sometimes be a very deep rabbit hole, this is "cross-compilation".

the termux-docker image, which does not do cross-compilation, is used in Termux packages only when the package is in this folder: 

https://github.com/termux-user-repository/tur/tree/master/tur-on-device

this folder can be used as a last resort when the package cannot be cross-compiled, but in my opinion it should be preferred to at least try as hard as you can to cross-compile it as well as being able to compile it on-device or inside of termux-docker.

if the termux-package-builder in most GitHub Actions is 0% similar to Android, and Termux on-device is 100% similar to Android, then termux-docker is approximately 90% similar to Android, in the big picture, for the purpose of things you can do in Termux.

3ls-it: OK, thanks, that tracks. And you answered some questions I've had. So, I'll work on it some more, and as a sanity check I just rebuild the package both on-device and in termux-docker, just fine. It occurred to me that in termux-package-builder we're trying to cross-compile a cross-compiler making things just that much more interesting! The fact that this is 20-year old code adds to  the "interesting".   
   
TIGCC definitely works, so I'm happy with that. I've embarked on a huge project: from this same group is the IDE KTIGCC, which also dates back to the early 2000s. It is deeply dependent on KDE/Qt 3. I wanted to "update" it to QT5 or 6. I'm actually looking at having to do a significant re-write. So I'm forking the IDE to "TIGCC-Qt". It will be a huge effort even thought the KTIGCC code-base is not that large. Once it's working and compiling/running on Linux and FreeBSD, I'll do a Termux package.
  
If you're satisfied with the shape of this PR, I'm good too. If you want anything changed, please let me know.

tur/libtifiles2/build.sh
```diff
@@ -0,0 +1,25 @@
+TERMUX_PKG_HOMEPAGE="http://lpg.ticalc.org/prj_tilp/"
+TERMUX_PKG_DESCRIPTION="Provides libtifiles2 library and headers for TiEmu and TiLP"
+TERMUX_PKG_LICENSE="GPL-2.0"
+TERMUX_PKG_MAINTAINER="3ls-it <3ls-it@pm.me>"
+TERMUX_PKG_VERSION=1.1.7
+# No version number in the master tarball; this is the cannonical source
+# and contains the source tarballs for all libti*
```
robertkirkman: What about these ones?

are these usable, or are they a bit off in some way?

https://sourceforge.net/projects/tilp/files/tilp2-linux/tilp2-1.18/

3ls-it: Ah, yes. the 4 components of libti* all have their own version numbers. These are the latest (last!) versions available.

robertkirkman: So the versions on Sourceforge are the same as the ones here, but with more normal archives, right?

3ls-it: Yes, same version. And, when I first set up the build recipes for the libti* components, I didn't know about the  Sourceforge files, I only knew of the tarball from ticalc.org which has all four bundled. Now that I know about the individual tarballs on Sourceforge, I'm inclined to redo `build.sh` for those four using the Sourceforge URLs instead. Fewer moving parts.  Would you agree that we should prefer Sourceforge?

robertkirkman: Yes, I think that makes sense.

tur/ti-utils/build.sh
```diff
@@ -0,0 +1,10 @@
+TERMUX_PKG_HOMEPAGE=https://github.com/termux-user-repository/tur
+TERMUX_PKG_DESCRIPTION="Meta package for TI calculator applications"
+TERMUX_PKG_LICENSE="Public Domain"
+TERMUX_PKG_MAINTAINER="3ls-it <3ls-it@pm.me>"
+TERMUX_PKG_VERSION=1.0
+TERMUX_PKG_AUTO_UPDATE=false
+TERMUX_PKG_SKIP_SRC_EXTRACT=true
+TERMUX_PKG_DEPENDS="tiemu, tilp2"
```
robertkirkman, line 8: Is this package necessary? Does one of them depend on the other, or not at all?

tur/libusb0/build.sh
```diff
@@ -0,0 +1,12 @@
+TERMUX_PKG_HOMEPAGE="https://libusb.info/"
+TERMUX_PKG_DESCRIPTION="Library to enable user space application programs to communicate with USB devices. 0.x branch."
+TERMUX_PKG_LICENSE="LGPL-2.1"
+TERMUX_PKG_MAINTAINER="3ls-it <3ls-it@pm.me>"
+TERMUX_PKG_VERSION=0.1.12
+TERMUX_PKG_SRCURL="http://deb.debian.org/debian/pool/main/libu/libusb/libusb_${TERMUX_PKG_VERSION}.orig.tar.gz"
+TERMUX_PKG_SHA256=37f6f7d9de74196eb5fc0bbe0aea9b5c939de7f500acba3af6fd643f3b538b44
+TERMUX_PKG_AUTO_UPDATE=false
+TERMUX_PKG_DEPENDS="glib"
+TERMUX_SUBPKG_CONFLICTS="libusb-compat"
```
robertkirkman: This package references "SUBPKG" while it is not a subpackage, so this variable wouldn't work, also where is "`libusb-compat`", I don't see any package with that name?

tur/ti-utils/build.sh
3ls-it: I guess it's not strictly necessary. `tiemu` and `tilp2` are not interdependent and can be installed separately. 

What I was trying to do is start a "TI Utilities meta package" to install the whole bundle, which was also going to include `tigcc` and a couple of others in the future. Is this not the right way?

robertkirkman: I am not sure what types of meta packages are permitted in TUR; I don't think TUR currently has any meta packages (other than ones designed to resolve real dependency needs). We should wait for other reviews.

tur/libusb0/build.sh
3ls-it: That is my mistake again. That is supposed to be `TERMUX_PKG_CONFLICTS` not "SUBPKG", plus it needs to be removed as I'm not submitting the `libusb-compat` package.   
  
I do have a working `libusb-compat`. The USB v1 support in TiLP is/was "experimental" and it want's to default to USB v0.1. So I wanted to go with a package that is still maintained. It was trivial to get `libusb-compat` working and the `tilp2` configure stage recognises it as USB v0.1.12. Unfortunately, it won't build against that library as it does not export the full set of symbols that `libusb0` does, so I modelled a `libusb0` package after the termux `libusb`, which works beautifully.
  
Long story short, I was going to submit both, and they conflict, but decided that it was useless to have `libusb-compat` since I have a working `libusb0`. I need to fix this.

3ls-it:
Hi @robertkirkman . I have turned this into a draft to hit "Pause" for a moment. Not for the reason that these packages don't work (they do), but because I was just contacted by Lionel Debroux, one of the original authors and current maintainer of libti*/tilp/tiemu. He contacted me to let me know that they are working on new versions of the libti* group, as well as other new software, and that I might want to work with the newer source instead. I'll look into this before deciding about this PR.

tur/ti-utils/build.sh
3ls-it: I've removed the meta package, now it's just TiEmu and TiLP2 as the two main packages, plus the dependency packages.

debrouxl:
Hi, upstream here :)

* libticables' libusb 0.x backend wasn't removed yet, but it has been untested for a good while, with most distros no longer packaging the ancient libusb 0.x. Building libticables' libusb 1.x backend would be better, all the more libusb 1.x has better support for async queries: no more direct ioctls to work around deficiencies of the libusb 0.x API.
* the libti* releases, especially, are very old by now, and the Git repositories contain many fixes and improvements. I'm slowly working towards a release train. Therefore, packaging snapshots from Git, like most Linux distros which package libti*, gfm and tilp2 do, would yield a more usable result :)
* any plans for packaging TilEm, which is TIEmu's libti*-based TI-Z80 emulator sibling, and whose main author contributed significantly to libti* back in the day ?

3ls-it:
> Hi, upstream here :)

Hello!

>     * libticables' libusb 0.x backend wasn't removed yet, but it has been untested for a good while, with most distros no longer packaging the ancient libusb 0.x. Building libticables' libusb 1.x backend would be better

Sounds good, I'll rebuild against `libusb` instead.

>     * the libti* releases, especially, are very old by now, and the Git repositories contain many fixes and improvements. 

Again, good to know. I'll work on getting those to build from the repository instead of the old tarballs.

>     * any plans for packaging TilEm, which is TIEmu's libti*-based TI-Z80 emulator sibling,  
  
Actually, yes. I meant to mention that to you. Once this PR is merged, that will be the next emulator for this group.

- Thanks!

3ls-it:
Working on building `libticables2` and `tilp2` against `libusb` instead of `libusb0`, but have run into errors that I did not previously encounter.

3ls-it:
> Working on building `libticables2` and `tilp2` against `libusb` instead of `libusb0`, but have run into errors that I did not previously encounter.

OK,`libticables2` is successfully building with `libusb` so the `libusb0` package is removed. I mistakenly thought that `tilp2` was dependent on `libusb`, but it is not, so that dep is removed and we're pulling the tarball from Sourcforge as we are with the `libti*` packges.  This is building in the Docker package-builder and on-device. Everything works as before, however I am seeing faster transfer speeds calc<->host.
 
@debrouxl , I have not yet switch over to the Git repository sources; that will be in a later revision/version bump. Thanks again for your input and guidance.

tur/tilp2/tilp
```diff
@@ -0,0 +1,46 @@
...
+        *)
+            printf '%s\n' \
+                "tilp: multiple USB devices found; set TILP_USB_DEVICE" >&2
+            printf '%s\n' "$devices" >&2
+            exit 1
+            ;;
+    esac
+fi
+
+exec termux-usb -r -E -e "$real" "$device"
```
robertkirkman, line 46: Could you avoid renaming the original binary and name the wrapper something like `termux-usb-tilp`?

This is because people who have a rooted device and are using `sudo tilp` to access USB could find it inconvenient that the original command always attempts to launch `termux-usb` if the script is not edited.

robertkirkman:
@3ls-it did you want to merge it with the stable versions soon, or wait for upstream to finish the next releases that they mentioned?

Also, I have a device for the USB connection, but I'm not experienced with BASIC or generally programming these, only the regular math use, so I wondered what test do you think would be appropriate for checking the USB connection of a phone to this TI-84 plus?

image

tur/tilp2/tilp
robertkirkman: another option might be detecting `id -u` being `0` or `whoami` being `root` and making the script launch `tilp.real` in that case.

3ls-it: Hi there. I've been away, and am just now catching up. I actually was originally thinking of keeping the binary name and calling the script something like `tilp2`, or what ever. So, no, I'm not opposed to keeping the binary name. Not at all.  
  
 Interesting, I have no experience with a rooted device. Didn't know about `sudo tilp` being a way to get USB access without `termux-usb`.  I think your script modification idea is a good one. That way, root or no-root, the user just launches `tilp`. Do you want to do it that way? I like it.  
  
BTW, here is my latest TI related project:
https://github.com/3ls-it/tigcc-qt
It's not a Termux package, but the `CMakeLists.txt` is Termux-aware. I'm working on  FreeBSD, but doing regular check-point builds on Ubuntu and Termux (as a package build). My buddy is working/building on Arch.

robertkirkman: > Interesting, I have no experience with a rooted device. Didn't know about sudo tilp being a way to get USB access without termux-usb. I think your script modification idea is a good one. That way, root or no-root, the user just launches tilp. Do you want to do it that way? I like it.

yes I think you should try that and then eventually I could test it

> https://github.com/3ls-it/tigcc-qt

I see, though you should know that it currently has hardcoded instances of string `/data/data/com.termux/files/usr` in it, so it wouldn't work without modifications in a duplicated Termux that has its package name changed from `com.termux` to anything else.

https://github.com/3ls-it/tigcc-qt/blob/229ca20cec3fb77f6e26efd97d0ae1add66ab145/cmake/FindQScintilla.cmake#L7

3ls-it:
> @3ls-it did you want to merge it with the stable versions soon, or wait for upstream to finish the next releases that they mentioned?

I'd like to merge sooner than later. I like the shape it's in right now with `libusb`. Even though there are a few known issue in TiLP, transferring file back and forth, and doing backup works, as do other basic functions.  
  
The newer `libti*` on git hub has a longer, deeper directory structure, and that alone means new patches. Plus I have no idea what has changed and what has not in the code base. As far as I can tell, switching to the GitHub sources seems like (almost) starting from scratch. What's your take on that aspect? Also, Lionel pointed out that building from the newer source will not fix the broken ROM dumper. That will require more work since the build of `libticalcs2` needs to call the `tigcc` toolchain to properly cross-build the ROM dumper. Since my `tigcc` package only builds on-device, there is more work to do for that as well.  
  
I'm feeling like this point is a good first release, with a later release based on uptream's next release _and_ having the external toolchain working to build the ROM dumper. Sound reasonable to you?  
  
> Also, I have a device for the USB connection, but I'm not experienced with BASIC or generally programming these, only the regular math use, so I wondered what test do you think would be appropriate for checking the USB connection of a phone to this TI-84 plus?  
  
Awesome!  I have two TI-89s and an NSpire CX II CAS (not supported in TiLP; too new) I'd love to know that this build is working with an 83/84 series. If you have the appropriate cable for your 84, it _should_ work. You should test every option especially file transfers/backups. No ROM dump of course. Sometimes you may need to scan for the device a couple of times at first. Another thing that I fixed|worked around is recovery from certain errors caused by trying to delete a locked file on the calculator. If it behaves as my 89s do, you'll get two back-to-back errors, then the connection recovers. Though, the move to USB 1 may have had a positive effect, and I'm hoping that a new upstream release will have fixed that. The bottom line is that I've not tested that series so I really have no idea! :laughing:

tur/tilp2/tilp
3ls-it: Ah, didn't realise that the path could be anything other that `/data/data/com.termux/files/usr`. Plus, yeah, I shouldn't hardcode paths! So we do `@TERMUX_PREFIX@`, right?

robertkirkman: well, `@TERMUX_PREFIX@` will only work if you replace it using a `sed` command before building. I would recommend reading the `$PREFIX` variable, though there is a chance that variable might go away in a future Termux version, but the similar `$TERMUX__PREFIX` variable is not currently set in the Termux App versions 0.118.0 or 0.118.1, only `$PREFIX`, so `$PREFIX` is the only universal variable for finding that path right now.

robertkirkman: I maintain an unofficial example of how to change the path from `/data/data/com.termux/files/usr` to anything else under `/data/data` here: https://github.com/robertkirkman/termux-generator

however, it is not the only fork which changes the package name; there are many others also

3ls-it: Ah, regarding `$PREFIX`: got it , thanks.  
Regarding the termux fully qualified path: I had no idea. Every PR and I learn something new about Termux.

debrouxl:
CX II support is one of the many improvements and bugfixes brought by switching to a Git version, which are listed in the Git history anyway. For now, only on a non-default branch, libticalcs uses the NavNet protocol spoken by the rest of the Nspire series to communicate with the CX II, but eventually the CX II's native NNSE protocol, which is a wrapper for a modified version of NavNet, will be supported, after reintegrating the modifications used for [WebTILP](https://web.tilp/info).

With several additional patches on top of tilibs Git master, since early 2019, libticalcs has been free of the severe maintenance liability represented by both the heavily patched fork of an extremely outdated version of GCC used in the community toolchain for the TI-68k series (preferably provided by GCC4TI), and by standard m68k GCC, which typically either generates improper code or crashes (ICE) with `-mshort`.
Stock m68k binutils, as packaged by e.g. Debian, are enough to rebuild the TI-68k ROM dumpers; stock z80+eZ80 binutils, also packaged by Debian, have recently been made enough to rebuild the TI-Z80 + TI-eZ80 ROM dumpers. IOW, `tigcc` and its underlying toolchain, or other uncommon tools, no longer are prerequisites for building libticalcs.
The integration onto the master branch of more easily buildable versions of the ROM dumpers had been blocked for years by lack of forward progress on additional features for a new tool which the ROM dumper changes depend on. However, I made that forward progress happen over the past few weeks, culminating in a commit which currently has a net diffstat > +8000 (> 2000 for build files, manpages, translation, etc. - but still). It shouldn't be long until the ROM dumper build improvements, and several more commits on top of those, reach the master branch.

3ls-it:
> CX II support is one of the many improvements and bugfixes brought by switching to a Git version

Wow, that's great!

>  Stock m68k binutils, as packaged by e.g. Debian, are enough to rebuild the TI-68k ROM dumpers; stock z80+eZ80 binutils, also packaged by Debian, have recently been made enough to rebuild the TI-Z80 + TI-eZ80 ROM dumpers.

> IOW, `tigcc` and its underlying toolchain, or other uncommon tools, no longer are prerequisites for building libticalcs.

OK, that's good info and I'm really rethinking how to approach this.

>  The integration onto the master branch of more easily buildable versions of the ROM dumpers had been blocked for years by lack of forward progress on additional features for a new tool which the ROM dumper changes depend on. However, I made that forward progress happen over the past few weeks, 

> It shouldn't be long until the ROM dumper build improvements, and several more commits on top of those, reach the master branch.

Thank you very much Lionel. Well, I'm convinced now that it's worth starting over with the newer Git versions. You've mentioned several times that there have been improvements. I believed you, of course, but not knowing the extent, I couldn't get too motivated to start over after putting so much time into the project. Now I see that it will be worth the effort. 
  
Regarding WebTILP, I'm on a non-rooted Android device and it doesn't work for me, even launching the browser with the `termux-usb` wrapper. Perhaps on a rooted device it would work as @robertkirkman explained above. Not something I can test though.  
  
Thank you again. :smile:

debrouxl:
The install_tilp.sh (user script) and recompile_tilp.sh (maintainer scripts) from https://github.com/debrouxl/tilp_and_gfm/tree/master/tilp/trunk/build/scripts show that the build changes stemming from the paths different from those in the release tarballs are expected to be minor.

The tilp_and_gfm Git repository removed compatibility with the outdated KDE support in early 2018, so `tur/tilp2/0001-configure-remove-kde.patch` will go away.

adriweb:
> Regarding WebTILP, I'm on a non-rooted Android device and it doesn't work for me, even launching the browser with the `termux-usb` wrapper. Perhaps on a rooted device it would work as @robertkirkman explained above. Not something I can test though.

Hi, WebTILP maintainer here. My friend @mrwompwomp just tried again with an 84+CE and it did work fine on his Android (using the standard Chrome browser as far as I know), and I have word it's been working for other people as well. There might be a WebSerial-related issue for 84+Evo compatibility but that's a different topic. 

here are some pictures:
- https://cdn.ti-planet.org/forum/gallery/image.php?image_id=24034
- https://cdn.ti-planet.org/forum/gallery/image.php?image_id=24036&short=1

3ls-it:
> Hi, WebTILP maintainer here. My friend @mrwompwomp just tried again with an 84+CE and it did work fine on his Android (using the standard Chrome browser as far as I know), and I have word it's been working for other people as well. There might be a WebSerial-related issue for 84+Evo compatibility but that's a different topic.

Hi Adrien. It's nice to have you and Lionel take some interest in this project. Thanks! I had previously tried WebTILP from inside of Termux:X11, but no joy due to it's unique USB access situation. I did just try on my phone with the default Android Chrome, but it tells me no compatible devices found. :man_shrugging: I then tried with my CX II and at first it was not seen, then I realised that the phone's case may be preventing the A>C USB adapter from seating all the way in the port. I removed the case and tried again, and it works really well for screen shots.  
screenshot-1
  
  
screenshot-2

Trying to dload the OS, though, it times out on the USB connection. I then tried the TI-89 again and the connection is flaky. I can get it to recognise, but get the same timeouts when trying to list files or screenshot. Maybe the little A>C adapter is flaky, and I sometimes wonder if this used GraphLink cable is flaky.   
  
But, back to TiLP2 running in Termux:X11. Now that I have it working, there's not really any flakiness in the connection. Once TiLP is communicating with the calculator, the connection state is stable, so perhaps the adapter and GraphLink are fine.

3ls-it:
Quoth @robertkirkman :
> Also, I have a device for the USB connection, but I'm not experienced with BASIC or generally programming these, only the regular math use, so I wondered what test do you think would be appropriate for checking the USB connection of a phone to this TI-84 plus?  
  
Hi Robert. I'm curious if you've tried connecting your 84+ to this current build of TiLP2, and what the result is.  
Soon I will start working with the Git sources for all of this, starting with the tilib* group. I've been focused on TIGCC-Qt Vim integration and am almost to a point where I feel good about pivoting back to this PR.  
  
-Cheers!

3ls-it:
@robertkirkman  is it possible to re-open? I didn't mean to close, and I've force pushed a new `ti-utils` branch. I don't see an option to re-open.

robertkirkman:
I don't think it is possible to re-open the same PR, it is necessary to make another PR.

robertkirkman:
I see this

image

3ls-it:
> I see this

:roll_eyes:  If only I could see as well as you!! :rofl: 

Thanks.

3ls-it:
Reopening with new push.

3ls-it:
> Reopening with new push.

Hmm, still closed?

robertkirkman:
it won't open again, you have to make another PR

robertkirkman:
wait hold on

robertkirkman:
nope i clicked the button to reopen it and nothing happened.

3ls-it:
I had that same experience; tried it once, nothing, then the option went grey.  
  
That's OK. I'm almost ready to open a new PR. I have everything except the emulator being built from the upstream Git. Everything is working at least as well as it did with the old sources, or better, and we _do_  have usable ROM dumpers now. I've loaded a dump of my TI-89 into the emulator with success.  
  
I've modified the TiLP launcher script as you asked to accommodate rooted device users. There are a few little buggy behaviours, which were also present in the former builds, that I need some guidance from upstream to get control of. (if possible).

Also, I have a packaging related issue/question: The ticalc.org server from where I'm pulling PedroM OS zip file seems to be hard to reach often; not fully reliable. And I worry a bit that they might go away. The zip file is a little under 6MB, would you be opposed to me adding it as payload in the TiEmu package? If that's too big, I understand.
  
Over-all, the big re-do of these packages has been very worth while.

robertkirkman:
> The ticalc.org server from where I'm pulling PedroM OS zip file seems to be hard to reach often; not fully reliable. And I worry a bit that they might go away.

I will check it but I don't think that it is necessary to add a new mirror of it; if there is an existing 3rd party mirror of it created by someone else then that might be acceptable.

3ls-it:
I've found the author's website and download URL: 
https://t3.yaronet.com/  

The PedroM download is the same contents as the Zip file from ticalc.org, however, the archive is 7zip.

robertkirkman:
That's good, then that one can be used to get PedroM.

PR #2755, opened 2026-08-22
refactor(tur/nerd-fonts): consolidate 8 ttf-*-nerd packages into nerd-fonts subpackages (v3.5.1)

GourangaDasSamrat:
## Changes

Per @licy183's suggestion on the original PR, this replaces the 8 independent
`ttf-*-nerd` packages with a single `nerd-fonts` package that builds all of
them and splits each font out into its own subpackage:

- `tur/nerd-fonts/build.sh` — downloads all 8 upstream Nerd Fonts release
  zips (v3.5.1), unpacks each one, and installs every font's `.ttf` files
  plus its own upstream `LICENSE`.
- `tur/nerd-fonts/ttf-*-nerd.subpackage.sh` (×8) — one per font, each keeping
  the same package name and description as before, claiming that font's
  installed files (and its own license) via `TERMUX_SUBPKG_INCLUDE`.

Every installed file is claimed by a subpackage, so the `nerd-fonts` parent
itself ends up empty — it only depends on all 8 subpackages via
`TERMUX_PKG_DEPENDS`, making it a convenience "install everything" package.
Each subpackage sets `TERMUX_SUBPKG_DEPEND_ON_PARENT=""`, so e.g.
`pkg install ttf-hack-nerd` still works standalone, exactly as before.

Also carries over the `TERMUX_PKG_AUTO_UPDATE=true` bump to upstream v3.5.1
from the original PR.

Upstream Release: https://github.com/ryanoasis/nerd-fonts/releases/tag/v3.5.1


licy183:
It would seem more reasonable to create a `nerd-fonts` package and treat the other `ttf-*` as subpackages.

tur/nerd-fonts/ttf-cascadia-code-nerd.subpackage.sh
```diff
@@ -0,0 +1,10 @@
+TERMUX_SUBPKG_DESCRIPTION="Cascadia Code patched with Nerd Fonts icons/glyphs (Font Awesome, Devicons, Octicons, Powerline, etc)"
+TERMUX_SUBPKG_PLATFORM_INDEPENDENT=true
+TERMUX_SUBPKG_DEPEND_ON_PARENT=""
```
licy183: IIRC this should be `TERMUX_SUBPKG_DEPEND_ON_PARENT=false`

tur/nerd-fonts/build.sh
```diff
@@ -0,0 +1,84 @@
...
+# All Nerd Fonts font families ship as separate flat *.ttf archives from the
+# same upstream release, so every archive is downloaded here and unpacked
+# per-font in termux_step_make_install(). Every font's own files then move
+# into its own ttf-*-nerd subpackage (see the *.subpackage.sh files next to
+# this build.sh), so this "nerd-fonts" package itself ends up with no files
+# of its own - it only depends on all of them, acting as a convenience
+# "install everything" meta package. Each ttf-*-nerd subpackage can still be
+# installed on its own, without pulling this meta package in.
+TERMUX_PKG_SRCURL=(
+	https://github.com/ryanoasis/nerd-fonts/releases/download/v${TERMUX_PKG_VERSION}/CascadiaCode.zip
```
licy183: The first one of `TERMUX_PKG_SRCURL` will be directly extracted and will not retain the archive name as a folder name.

I think it should be set to `https://github.com/ryanoasis/nerd-fonts/archive/refs/tags/v${TERMUX_PKG_VERSION}.tar.gz`. In this way, auto-updates can also work properly.

licy183:
Ahh... That's not what I meant.

I meant that `TERMUX_PKG_SRCURL` should still be an array, containing the font zip archive files, and the first item should be the tagged `.tar.gz` file. This way, the `.tar.gz` file will be extracted in `TERMUX_PKG_SRCDIR`, and the font zip files will also be extracted to the corresponding folder, making script writing much simpler.

GourangaDasSamrat:
@licy183 Updated as suggested! Could you please take another look when you get a chance? Thanks!

licy183:
It still doesn't look right, let me fix it.

robertkirkman:
thank you for fixing it licy183 i was going to request this in the original PR but you merged that one before i did

tur/nerd-fonts/build.sh
```diff
@@ -0,0 +1,93 @@
...
+	de7bf85382dad8c239696b7ccddd3e69c5a79b77c6d18c6abe9d6676d28a8764
+	fab782a66f7d3019da64f6572db9fc5d3a4bcb19f9fa13e2d8a62e3693d6396e
+	fb104893ecd8f57e8afacbc0a7086b42657120448d056d6093c728a9afb8e237
+	b5db570b0b2bf5a3a62911aeefd2c8df91a12dcb66261169e7353f984002d5b7
+	e0000bc77cd77279840d8613e98727548f81f291a571be0be8699c8e3be36fd3
+	81cf501f80581bc2e73f70264441cf96980736a4bbb07c090996af6ab8f19013
+)
+TERMUX_PKG_PLATFORM_INDEPENDENT=true
+TERMUX_PKG_AUTO_UPDATE=true
+TERMUX_PKG_DEPENDS="ttf-cascadia-code-nerd, ttf-hack-nerd, ttf-inconsolata-nerd, ttf-jetbrains-mono-nerd, ttf-meslo-nerd, ttf-roboto-mono-nerd, ttf-sourcecodepro-nerd, ttf-victor-mono-nerd"
```
robertkirkman, line 31: except: in Arch Linux, the command `pacman -S nerd-fonts` will install `ttf-nerd-fonts-symbols`, but that was missed/omitted here, so that will not occur in Termux since `ttf-nerd-fonts-symbols` is not in `TERMUX_PKG_DEPENDS` of `nerd-fonts` in Termux.

PR #2756, opened 2026-08-22
bump(tur-on-device/rakudo): 2026.08

rc-chuah:


licy183:
`moarvm` needs to be bumped in another PR.

Only after that PR is merged can the other two on-device packages be bumped.

PR #2757, opened 2026-08-22
add: some packages

wataamee777:


licy183:
The glibc package cannot be run directly on Termux. You can wrap it with glibc-runner, but this is not within the scope of TUR's packaging.

PR #2762, opened 2026-08-23
feat(tur/nerd-fonts): depend on ttf-nerd-fonts-symbols for Arch parity

GourangaDasSamrat:
Fixes a review comment from @robertkirkman on #2755:

> except: in Arch Linux, the command `pacman -S nerd-fonts` will
> install `ttf-nerd-fonts-symbols`, but that was missed/omitted here,
> so that will not occur in Termux since ttf-nerd-fonts-symbols is not
> in TERMUX_PKG_DEPENDS of nerd-fonts in Termux.

`ttf-nerd-fonts-symbols` is already an official Termux package
(maintained by @termux, not part of TUR), so this doesn't add a new
subpackage or build anything new - it just adds it to
`TERMUX_PKG_DEPENDS` so `pkg install nerd-fonts` pulls it in from the
main repo, same as the other 8 font subpackages.

### DIFF
--- a/tur/nerd-fonts/build.sh
+++ b/tur/nerd-fonts/build.sh
@@ -3,6 +3,7 @@ TERMUX_PKG_DESCRIPTION="Meta package that installs all Nerd Fonts patched font p
 TERMUX_PKG_LICENSE="OFL-1.1, MIT, Apache-2.0"
 TERMUX_PKG_MAINTAINER="Gouranga Das Samrat <gouranga.das.khulna@gmail.com>"
 TERMUX_PKG_VERSION=3.5.1
+TERMUX_PKG_REVISION=1
 # The first entry is for the auto-update system to fetch the latest release tarball, the rest are the upstream font zips.
 TERMUX_PKG_SRCURL=(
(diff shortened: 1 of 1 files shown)


robertkirkman:
It will need to have `TERMUX_PKG_REVISION=1` added to it so that it is a newer version than the previous version.

GourangaDasSamrat:
> It will need to have `TERMUX_PKG_REVISION=1` added to it so that it is a newer version than the previous version.

I thought the same thing, but I don't know why, without this, the ci fails in main repo, But passed here.

robertkirkman:
The CI here does not check that, so it is necessary to do it manually every time otherwise it will fail at a later stage.

GourangaDasSamrat:
I've updated it with TERMUX_PKG_REVISION=1 and pushed the changes. Thanks for clarifying!

robertkirkman:
On my screen, the change is still not there, are you sure that you pushed it already?

GourangaDasSamrat:
Ohh sorry,i forgot to push, I've pushed now.

robertkirkman:
The PR has been force pushed, but on my screen there still is no `TERMUX_PKG_REVISION=1`

image

robertkirkman (approved):

GourangaDasSamrat:
my bad, i forced push but forgot to git add

PR #2763, opened 2026-08-23
feat: add omnicli package

Manash07Bhoi:
Add OmniCLI to TUR. Tested compilation natively using termux-packages standard.


GourangaDasSamrat:
You should squish your 4 commits into 1, and commit message and pr title both should be this:

```
addpkg(tur/omnicli): 0.1.1
```

PR #2767, opened 2026-08-25
dosemu2: fix assert at startup

stsp:
Reported by @DuilioPerez

### DIFF
--- a/tur/dosemu2/build.sh
+++ b/tur/dosemu2/build.sh
@@ -3,11 +3,11 @@ TERMUX_PKG_DESCRIPTION="Run DOS programs under linux."
 TERMUX_PKG_LICENSE="GPL-2.0"
 TERMUX_PKG_MAINTAINER="@stsp"
 TERMUX_PKG_VERSION="2.0pre9-git"
-TERMUX_PKG_REVISION=10
+TERMUX_PKG_REVISION=11
 TERMUX_PKG_SRCURL=git+https://github.com/dosemu2/dosemu2
 TERMUX_PKG_GIT_BRANCH=devel
-_COMMIT=82770aba398485117c56523a1a5c261f6e37ca64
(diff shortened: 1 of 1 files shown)


robertkirkman (approved):

stsp:
@DulioPerez please check the results.
I tried under waydroid, all seems well.

stsp:
@DuilioPerez please check the results.
I tried under waydroid, all seems well.

DuilioPerez:
@stsp It's working now. But when I run Windows 3 an assertion is rised after a while:
~ $ dosemu -d Windows3/
CONF: emulated CPU forced down to real CPU: 386
ERROR: ladspa: failed to load filter.so
ERROR: soundfonts not found
ERROR: Async signal 14 from thread
/home/builder/.termux-build/dosemu2/src/src/plugin/searpc/util.c:45: char *transport_callback(void *, const char *, size_t, size_t *): assertion "!sock->in_async" failed

Aborted

stsp:
I have problems reproducing that,
but can you test this patch:
```
--- a/src/arch/linux/async/signal.c
+++ b/src/arch/linux/async/signal.c
@@ -802,7 +802,7 @@ static int sigasync0(int sig)
     char name[128];
     pthread_getname_np(tid, name, sizeof(name));
     dosemu_error("Async signal %i from thread %s\n", sig, name);
-#elif defined(__APPLE__)
+#elif defined(__APPLE__) || defined(__ANDROID__)
     /* somehow SIGALRM often gets unblocked in threads on MacOS,
        forwarding to main thread */
     pthread_kill(dosemu_pthread_self, sig);
```

Before I can install win31 under termux,
I need to add mtools package... Something
to work on later. Trying the above patch
might be easier. :)

DuilioPerez:
@stsp The assertion just happens in installation. I think it's expectable because I'm not insyalling in a HHD image but in the dosemu folder. The only remaining issue is with click

PR #2769, opened 2026-08-27
dosemu2: add the potential fix for crash on android

stsp:
@DuilioPerez reported another problem, which I have problems reproducing. But I think it should be fixed in this revision. There is no other way to find out, but to try out...


robertkirkman (approved):

stsp:
@DuilioPerez please see if the problem persists.

DuilioPerez:
@stsp I don't see any assertion anymore. The only remaining issue is with single click

stsp:
What is it?

DuilioPerez:
> What is it?

On Windows 3 mouse single click doesn't work. But long click works. Also right click doesn't work too, but I'm not sure if right click works on Windows 3. I'll do some more research and testing

stsp:
Can you check that on desktop,
rather than on termux, and fill
the ticket upstream?

PR #2774, opened 2026-08-30
bun: fall back to openat instead of openat2 on Android

### LINKED ISSUE #2773: bun: tracking issues after build-from-source packaging
## Bug list

- [ ] Installing a package with `bun add`/`bun install` that requires downloading a native binary fails with `Bad system call` unless `BUN_OPTIONS="--os=android"` is set — but testing shows this isn't simply about the variable being set:
  - On a Walton WALPAD_8G (Android 11, kernel 4.19.127, Termux 0.118.3), `bun add` works fine in `zsh` **even without** `BUN_OPTIONS` set at all. In `bash` on the same device, it fails **regardless** of whether `BUN_OPTIONS` is set or not — so this isn't a variable-dependency issue, but there is clearly a shell-specific issue on this device (bash never works, zsh always works).
  - On a Xiaomi 23127PN0CG (Android 14, kernel 6.1.57, Termux 0.119.0-beta.3), the same command fails in `zsh` too (`zsh: invalid system call`), even after `rm -rf ~/.bun` and exporting `BUN_OPTIONS="--os=android"`.
  - So there appear to be two overlapping factors: a shell-specific issue (bash consistently fails on at least one device regardless of `BUN_OPTIONS`) and a device/kernel-specific issue (zsh itself fails on a different device). The most notable differences between the two devices are Android version (11 vs 14) and kernel version (4.19 vs 6.1), which may point to a syscall being blocked by a newer kernel's seccomp filter — similar to the existing `close_range` workaround already patched for this package.
- [ ] When installing a package with bun, a `bun.lock` file is expected to be created, but it isn't being created on Termux.

## Related
- #2771 (original PR and discussion where this was reported)

GourangaDasSamrat:
Closes #2773


GourangaDasSamrat:
Closing this cause #2775 is already merged.

PR #2775, opened 2026-08-30
bun: disable openat2 and fchmodat2

licy183:


robertkirkman:
Shouldn't you tell them upstream that this (#2775) is the correct way instead of the other way that is wrong?

- https://github.com/oven-sh/bun/pull/39775
- https://github.com/oven-sh/bun/pull/39084

licy183:
To be honest, since I'm not a native English speaker, I really don't like communicating with LLM in English. It always outputs an extremely long sentence that is tedious to read.

The approach in this PR is quite different from `bun`'s routine. In Termux, an usual practice for the system calls that are only available in higher versions is to simply treat them as non-existent. However, `bun` will detect whether the call actually exists, and if not, it will perform a fallback process.

PR #2780, opened 2026-09-01
addpkg(tur/tailcat): 0.6.0

ancientcatz:


ancientcatz:
@licy183 I wonder why this can't be merged?

tur/tailcat/build.sh
```diff
@@ -0,0 +1,50 @@
...
...portal,ts_omit_capture,ts_omit_cliconndiag,ts_omit_clientmetrics,ts_omit_clientupdate,ts_omit_cloud,ts_omit_colorable,ts_omit_completion,ts_omit_completion_scripts,ts_omit_conn25,ts_omit_debug,ts_omit_debugeventbus,ts_omit_debugportmapper,ts_omit_desktop_sessions,ts_omit_dns,ts_omit_doctor,ts_omit_drive,ts_omit_favorites,ts_omit_flashappliance,ts_omit_health,ts_omit_hujsonconf,ts_omit_identityfederation,ts_omit_ipnbus,ts_omit_iptables,ts_omit_kube,ts_omit_linkspeed,ts_omit_linuxdnsfight,ts_omit_listenrawdisco,ts_omit_logtail,ts_omit_netlog,ts_omit_networkmanager,ts_omit_oauthkey,ts_omit_osrouter,ts_omit_outboundproxy,ts_omit_peerapiclient,ts_omit_peerapiserver,ts_omit_portlist,ts_omit_portmapper,ts_omit_posture,ts_omit_qrcodes,ts_omit_relayserver,ts_omit_remoteconfig,ts_omit_resolved,ts_omit_routecheck,ts_omit_runtimemetrics,ts_omit_sdnotify,ts_omit_serve,ts_omit_serviceclientprefs,ts_omit_synology,ts_omit_syslog,ts_omit_syspolicy,ts_omit_systray,ts_omit_taildrop,ts_omit_tailnetlock,ts_omit_tap,ts_omit_tpm,ts_omit_tundevstats,ts_omit_unixsocketidentity,ts_omit_useexitnode,ts_omit_useproxy,ts_omit_usermetrics,ts_omit_useroutes,ts_omit_wakeonlan,ts_omit_webbrowser,ts_omit_webclient \
```
licy183: This part seems to be in build-tags.txt. Why not just use that file directly?

tur/tailcat/build.sh
```diff
@@ -0,0 +1,50 @@
...
+		curl -s "https://api.github.com/repos/tailscale/tailcat/commits/v${TERMUX_PKG_VERSION}" \
+			| jq -r '[.sha, .commit.committer.date, (.commit.committer.date | fromdateiso8601)] | @tsv'
+	)
+
+	local overlay="${TERMUX_PKG_BUILDER_DIR}/overlay"
+	local overlay_file="${TERMUX_PKG_BUILDER_DIR}/overlay.json"
+
+	cat > "${overlay_file}" <<EOF
+{
+	"Replace": {
```
licy183: If this part of the code has been well-tested, I think it can be committed to the main repository as a patch. This way, all Go programs will receive the same function.

tur/tailcat/overlay/src/net/interface_android.go
```diff
@@ -0,0 +1,436 @@
+// Copyright 2011 The Go Authors. All rights reserved.
+// Use of this source code is governed by a BSD-style
+// license that can be found in the LICENSE file.
+
+//go:build android
```
licy183: I think a better way is not to host these golang files here, but to download them and then use diff to patch them.

licy183 (commented):
I'm currently on vacation, so I can't reply and review immediately. I will reply as soon as I see the message, but there may be a delay.

tur/tailcat/build.sh
ancientcatz: It’s not exactly the same. I’m omitting the "netgo" and "osusergo" tags.

tur/tailcat/overlay/src/net/interface_android.go
ancientcatz: Where should I host the Go files? In a Git repository?

tur/tailcat/build.sh
ancientcatz: I think it’s good already. Basically, it’s the same as using the `github.com/wlynxg/anet` module, but without the hassle. You can take a look at it when you have some free time.

ancientcatz:
> I'm currently on vacation, so I can't reply and review immediately. I will reply as soon as I see the message, but there may be a delay.

No worries at all, and sorry for bothering you while you’re on vacation. Please take your time and enjoy your break. I really appreciate you taking a look whenever you get a chance.

tur/tailcat/build.sh
robertkirkman: can you write it as a sed patch on the original file?

tur/tailcat/build.sh
robertkirkman: does the original software depend on the `github.com/wlynxg/anet` module? why is that a hassle?

tur/tailcat/overlay/src/net/interface_android.go
robertkirkman: they should be retrieved from the original place that you copied and pasted the `interface_android.so` and other files from before modifying them

tur/tailcat/build.sh
ancientcatz: No, the `tailscale.com/net/netmon` modules can't run on Android without patching them with `github.com/wlynxg/anet`. However, that's a hassle, and using an overlay is a more stable approach.

robertkirkman: What if a future version of github.com/wlynxg/anet changes the `interface_android.go` and then this version you copied and pasted does not get auomtiacally updated? I don't want that

tur/tailcat/overlay/src/net/interface_android.go
ancientcatz: Could you elaborate a bit more? `interface_android.go` and `netlink_android.go` are patched versions of their Linux counterparts. Would patching them directly in `$GOROOT` be a better solution?

robertkirkman: not in `$GOROOT` - they should be patched in the `$TERMUX_PKG_SRCDIR/vendor` folder after using the command `go mod vendor`.

tur/tailcat/build.sh
ancientcatz: > What if a future version of github.com/wlynxg/anet changes the `interface_android.go` and then this version you copied and pasted does not get auomtiacally updated? I don't want that

You can compare them if you want, but this patch is inspired by `github.com/wlynxg/anet`, not a direct copy. It is implemented against the latest stdlib APIs, whereas `anet` is built around an older stdlib baseline and is arguably more outdated.

tur/tailcat/overlay/src/net/interface_android.go
ancientcatz: > not in `$GOROOT` - they should be patched in the `$TERMUX_PKG_SRCDIR/vendor` folder after using the command `go mod vendor`.

That’s not possible. Go’s standard library can’t be vendored. That's why I'm using overlay.

tur/tailcat/build.sh
robertkirkman: if your version is newer than github.com/wlynxg/anet, please write your version as a patch onto the latest stdlib APIs globally for all packages instead of separately only for this package

tur/tailcat/build.sh
ancientcatz: > can you write it as a sed patch on the original file?

It’s done. Just needs a few adjustments, then I’ll force-push it.

tur/tailcat/build.sh
ancientcatz: > if your version is newer than github.com/wlynxg/anet, please write your version as a patch onto the latest stdlib APIs globally for all packages instead of separately only for this package

Where should I push the patch? To the upstream `termux/termux-packages` repo?

tur/tailcat/overlay/src/net/interface_android.go
robertkirkman: If that's not possible, then what is this? how does this work, and is it different from the Go standard library?

https://github.com/termux/termux-packages/blob/96a2cc7469efb42aac263b96501c50cc3f6de167/packages/golang/patch-script/fix-hardcoded-etc-resolv-conf.diff

ancientcatz: > If that's not possible, then what is this? how does this work, and is it different from the Go standard library?
> 
> https://github.com/termux/termux-packages/blob/96a2cc7469efb42aac263b96501c50cc3f6de167/packages/golang/patch-script/fix-hardcoded-etc-resolv-conf.diff

This doesn't involve `vendor` either. Vendor patching only applies to external modules and patches files in `$TERMUX_PKG_SRCDIR`. Since I need to patch the standard library directly in `$GOROOT`, I need to use an overlay instead.

robertkirkman: would it be possible for you to apply these files by modifying the contents of the `packages/golang/patch-script` folder so that they apply to all Go packages, or is the way you show that is specific to this package the only way?

ancientcatz: > would it be possible for you to apply these files by modifying the contents of the `packages/golang/patch-script` folder so that they apply to all Go packages, or is the way you show that is specific to this package the only way?

Ah, I just understood what you meant. In that case, we also need to adjust this file, right? [termux_setup_golang.sh](https://github.com/termux/termux-packages/blob/master/scripts/build/setup/termux_setup_golang.sh)

robertkirkman: Yes, I think so, and I think that is what licy183 meant by saying "it can be committed to the main repository as a patch. This way, all Go programs will receive the same function."

tur/tailcat/build.sh
robertkirkman: Yes, I think so, and I think that is what licy183 meant by saying "it can be committed to the main repository as a patch. This way, all Go programs will receive the same function."

ancientcatz: > Yes, I think so, and I think that is what licy183 meant by saying "it can be committed to the main repository as a patch. This way, all Go programs will receive the same function."

Could you review my fix and let me know if it looks good? https://github.com/termux/termux-packages/pull/31462

robertkirkman: Yes, it does, thank you.

licy183:
I just merged https://github.com/termux/termux-packages/commit/d267281163dd7cfc8ee6866f31fd7252cc6a1334, so the building script here can now be simplified.

ancientcatz:
Waiting for this PR to be merged: https://github.com/termux/termux-packages/pull/31526

licy183:
Merged in https://github.com/termux/termux-packages/commit/90e99c6daf2e1d75c97389aa1fa5f87d140ef2a4

ancientcatz:
Ready to merge. @licy183 @robertkirkman

tur/tailcat/build.sh
```diff
@@ -0,0 +1,39 @@
...
+
+	local build_tags="$(cat build-tags.txt)"
+
+	go build \
+		-ldflags "-s -w -X main.version=v${TERMUX_PKG_VERSION} -buildid=" \
+		-tags="${build_tags}" \
+		-trimpath \
+		-buildvcs=false \
+		./cmd/tailcat
+	touch -d "@${commit_epoch}" "tailcat"
```
robertkirkman: What does this do?

ancientcatz: Which one?

robertkirkman: `touch -d "@${commit_epoch}" "tailcat"`; what I mean is, most Go packages do not need that, so what happens if it is removed?

ancientcatz: Yes, it’s not really important. I just want it to produce a reproducible binary with the same timestamp. If that’s not a good approach, I can remove it.

robertkirkman: Is the modification timestamp of the binary itself important for build reproducibility? I didn't think so because isn't this set globally?

https://github.com/termux/termux-packages/blob/e8c195eb65c308c90adc82414d5f613021c40a85/scripts/build/termux_step_create_debian_package.sh#L9

ancientcatz: > Is the modification timestamp of the binary itself important for build reproducibility? I didn't think so because isn't this set globally?
> 
> https://github.com/termux/termux-packages/blob/e8c195eb65c308c90adc82414d5f613021c40a85/scripts/build/termux_step_create_debian_package.sh#L9

Not really. The checksum stays the same whether the timestamp is the same or not, so.. it’s just a cherry on top.

robertkirkman: I think that it should be removed so that it does not cause confusion.

ancientcatz: > I think that it should be removed so that it does not cause confusion.

It's done.

robertkirkman:
@ancientcatz wait for this https://github.com/termux/termux-packages/pull/31509#issuecomment-5565843008

robertkirkman:
ok now rebase it after this: https://github.com/termux-user-repository/tur/commit/d21e9c58f91d76f5b3eb8dc311a8c3f84a322c1e

robertkirkman:
Does it work without tailscale being installed? if not, do you have a version of tailscale installed that works with this?

ancientcatz:
> Does it work without tailscale being installed? if not, do you have a version of tailscale installed that works with this?

It doesn't require a Tailscale account or token at all. If you need Tailscale in Termux, you can probably use https://github.com/tailscale/ts-plug instead. It can expose local services to your tailnet and also make tailnet services accessible locally.

Actually, I initially wanted to add ts-plug to TUR instead of tailcat, but the project itself is rather unmaintained and still a work in progress, with no tagged releases yet. It also needs a vendor patch to build properly.

robertkirkman (approved):

tur/tailcat/build.sh
```diff
@@ -0,0 +1,31 @@
...
+		-ldflags "-s -w -X main.version=v${TERMUX_PKG_VERSION} -buildid=" \
+		-tags="${build_tags}" \
+		-trimpath \
+		-buildvcs=false \
+		./cmd/tailcat
+}
+
+termux_step_make_install() {
+	unset GOOS GOARCH CGO_LDFLAGS
+	unset CC CXX CFLAGS CXXFLAGS LDFLAGS
```
robertkirkman, line 29: for future reference, it seems unnecessary to unset things here.

PR #2784, opened 2026-09-03
tur-on-device/ocaml: add package at version 5.5.0 and opam 2.5.2

DevGitPit:


DevGitPit:
@robertkirkman Bun needs to be removed from here?- 

```
Fetched 18.4 kB in 1s (19.2 kB/s)
Reading package lists...
Building dependency tree...
Reading state information...
All packages are up to date.
ERROR: Duplicated package: bun
```

tur-on-device/ocaml/build.sh
```diff
@@ -0,0 +1,35 @@
+TERMUX_PKG_HOMEPAGE=https://ocaml.org/
+TERMUX_PKG_DESCRIPTION="Industrial strength programming language with an emphasis on expressive types and safety"
+TERMUX_PKG_LICENSE="LGPL-2.1"
+TERMUX_PKG_MAINTAINER="@termux-user-repository"
+TERMUX_PKG_VERSION="5.5.0"
+TERMUX_PKG_SRCURL="https://github.com/ocaml/ocaml/releases/download/${TERMUX_PKG_VERSION}/ocaml-${TERMUX_PKG_VERSION}.tar.gz"
+TERMUX_PKG_SHA256=c018052c8264a3791a8f54f84179e6bcc78ed82eb889bacc2773df445259aed3
+TERMUX_PKG_DEPENDS="binutils, clang, libandroid-shmem"
+TERMUX_PKG_BUILD_DEPENDS="which"
+TERMUX_PKG_EXCLUDED_ARCHES="arm,i686"
```
robertkirkman: what error occurs during the build for 32-bit targets? also please do not make the package depend on `which`, instead patch all instances of `which` in the code to use `command -v` instead

DevGitPit: The error: `Makefile:2289: *** The source tree was configured with --disable-native-compiler!. Stop.`

tur-on-device/ocaml/build.sh
```diff
@@ -0,0 +1,34 @@
+TERMUX_PKG_HOMEPAGE=https://ocaml.org/
+TERMUX_PKG_DESCRIPTION="Industrial strength programming language with an emphasis on expressive types and safety"
+TERMUX_PKG_LICENSE="LGPL-2.1"
+TERMUX_PKG_MAINTAINER="@termux-user-repository"
+TERMUX_PKG_VERSION="5.5.0"
+TERMUX_PKG_SRCURL="https://github.com/ocaml/ocaml/releases/download/${TERMUX_PKG_VERSION}/ocaml-${TERMUX_PKG_VERSION}.tar.gz"
+TERMUX_PKG_SHA256=c018052c8264a3791a8f54f84179e6bcc78ed82eb889bacc2773df445259aed3
+TERMUX_PKG_DEPENDS="binutils, clang, libandroid-shmem"
```
robertkirkman: which program in the `binutils` package does it depend on? that is somewhat important because it gets complicated in Termux.

robertkirkman: ideally, I think a comment should be added explaining which program in `binutils` the package depends on in order to make maintenance easy.

DevGitPit: It doesn't, apparently. Let me fix.

tur-on-device/ocaml/build.sh
robertkirkman: Thanks, actually when I test, I see a different error:

```
linker: "/data/data/com.termux/files/home/.termux-build/ocaml/src/otherlibs/unix/dllunixbyt-i686-pc-linux-android-a101.so" has text relocations (https://android.googlesource.com/platform/bionic/+/master/android-changes-for-ndk-developers.md#Text-Relocations-Enforced-for-API-level-23)
Fatal error: cannot load shared library dllunixbyt-i686-pc-linux-android-a101
Reason: dlopen failed: "/data/data/com.termux/files/home/.termux-build/ocaml/src/otherlibs/unix/dllunixbyt-i686-pc-linux-android-a101.so" has text relocations (https://android.googlesource.com/platform/bionic/+/master/android-changes-for-ndk-developers.md#Text-Relocations-Enforced-for-API-level-23)
libc: Fatal signal 6 (SIGABRT), code -6 (SI_TKILL) in tid 21822 (ocamlrun), pid 21822 (ocamlrun)
```

but using `scanelf` on the file shows this:

```
~/.../ocaml/src $ scanelf -qT otherlibs/unix/dllunixbyt-i686-pc-linux-android-a101.so 
scanelf: scanelf_file_textrels(): ELF otherlibs/unix/dllunixbyt-i686-pc-linux-android-a101.so has TEXTREL markings but doesn't appear to have any real TEXTREL's !?
  otherlibs/unix/dllunixbyt-i686-pc-linux-android-a101.so
~/.../ocaml/src $ 
```

and upstream README says that "native compilation" is not supported for 32-bit in Ocaml 5.0 and newer,

https://github.com/ocaml/ocaml#overview

I don't know what that means or if it means it should be working still or not, so it can stay disabled for 32-bit targets. The one thing you should change here is to put a space in between the `,` symbol and the `i`.

DevGitPit: You mean **comma** and `i`? Also yes after burning my life force away patching `arm` and `i686` on **Julia** if someone says no 32bit I accept it, lol.

robertkirkman: Yes comma and i

tur-on-device/ocaml/build.sh
```diff
@@ -0,0 +1,34 @@
...
+TERMUX_PKG_DESCRIPTION="Industrial strength programming language with an emphasis on expressive types and safety"
+TERMUX_PKG_LICENSE="LGPL-2.1"
+TERMUX_PKG_MAINTAINER="@termux-user-repository"
+TERMUX_PKG_VERSION="5.5.0"
+TERMUX_PKG_SRCURL="https://github.com/ocaml/ocaml/releases/download/${TERMUX_PKG_VERSION}/ocaml-${TERMUX_PKG_VERSION}.tar.gz"
+TERMUX_PKG_SHA256=c018052c8264a3791a8f54f84179e6bcc78ed82eb889bacc2773df445259aed3
+TERMUX_PKG_DEPENDS="clang, libandroid-shmem"
+TERMUX_PKG_EXCLUDED_ARCHES="arm, i686"
+TERMUX_PKG_BUILD_IN_SRC=true
+TERMUX_PKG_UNDEF_SYMBOLS_FILES="all"
```
robertkirkman: This is not good because this would happen so it's bypassing the error in an unstable way,

```
SKIP: ./lib/unix/unix.a
ERROR: ./lib/unix/unix.cmxs contains undefined symbols:
   149: 0000000000000000     0 NOTYPE  GLOBAL DEFAULT   UND modf
   165: 0000000000000000     0 NOTYPE  GLOBAL DEFAULT   UND nextafter
INFO: Done ... 6s
INFO: Found 1 files with undefined symbols after exclusion
ERROR: Refer above
```

Instead of this line, I think that `-lm` should be added to `LDFLAGS` in the same place where `-landroid-shmem` is (I checked and that does make the error stop)

robertkirkman (approved):

robertkirkman:
This is very awesome good job because nobody ever made `ocaml` a package successfully before

PR #2798, opened 2026-09-07
Python tokenizers

DevGitPit:


tur/python-hf-xet/build.sh
```diff
@@ -0,0 +1,86 @@
...
+	local _whl="$(find "$TERMUX_PKG_SRCDIR" -name "hf_xet-${TERMUX_PKG_VERSION}-*.whl" -print -quit)"
+	local _whl_dest="hf_xet-${TERMUX_PKG_VERSION}-py${TERMUX_PYTHON_VERSION/./}-none-any.whl"
+	mv "$_whl" "$_whl_dest"
+	pip install \
+		--force-reinstall \
+		--no-deps \
+		"$_whl_dest" \
+		--prefix "$TERMUX_PREFIX"
+
+	# Strip optional test dependencies from METADATA so pip debscripts are not triggered
```
robertkirkman: Do you know if there is a way to automatically detect whether all dependencies are test-only or there are real runtime dependencies in the `METADATA` file?

I did not implement checking for that globally because I don't know how to do that or if that is possible.

tur/python-hf-xet/build.sh
```diff
@@ -0,0 +1,86 @@
...
+TERMUX_PKG_SRCURL="https://files.pythonhosted.org/packages/1b/ab/522a2ab67f27971a9d48ca666d4fca85ef7d5282d142e31fd087e27b1bbe/hf_xet-$TERMUX_PKG_VERSION.tar.gz"
+TERMUX_PKG_SHA256=2e58454a340b3556dfa4972d5451aff4fba8dd42a236600ba1a1d2b1514f0fef
+TERMUX_PKG_AUTO_UPDATE=true
+TERMUX_PKG_DEPENDS="libc++, python"
+TERMUX_PKG_PYTHON_COMMON_BUILD_DEPS="'maturin>=1.7,<2.0'"
+TERMUX_PKG_BUILD_IN_SRC=true
+TERMUX_PKG_PYTHON_RUNTIME_DEPS=false
+
+_setup_rust_no_explicit_version() {
+	: "${CARGO_HOME:="$TERMUX_PKG_TMPDIR/.cargo"}"
```
robertkirkman: is this custom function copied from `python-tokenizers` actually also required for `python-hf-xet`, or would the regular `termux_setup_rust()` work?

tur/python-hf-xet/build.sh
```diff
@@ -0,0 +1,86 @@
+TERMUX_PKG_HOMEPAGE=https://github.com/huggingface/xet-core
+TERMUX_PKG_DESCRIPTION="Fast transfer of large files with the Hugging Face Hub"
+TERMUX_PKG_LICENSE="Apache-2.0"
+TERMUX_PKG_MAINTAINER="@termux-user-repository"
+TERMUX_PKG_VERSION="1.6.0"
+TERMUX_PKG_SRCURL="https://files.pythonhosted.org/packages/1b/ab/522a2ab67f27971a9d48ca666d4fca85ef7d5282d142e31fd087e27b1bbe/hf_xet-$TERMUX_PKG_VERSION.tar.gz"
```
robertkirkman: please try to build it from this URL:

https://github.com/huggingface/xet-core/releases/download/v1.6.0/hf_xet-1.6.0.tar.gz

tur/python-hf-xet/build.sh
DevGitPit: - Optional / test extras (like `tests`, `dev`, `docs`) always include an environment marker containing `extra == "..."` or `extra == '...'` in their `Requires-Dist` entries: 
```
Requires-Dist: pytest ; extra == 'tests'
Requires-Dist: coverage (>=6.0) ; extra == "dev"
```

 - Real / mandatory runtime dependencies never evaluate on extra (they either have no marker or only platform/python version markers like `; python_version < "3.11"`): 
```
Requires-Dist: numpy (>=1.20)
Requires-Dist: importlib-metadata ; python_version < "3.8"
```

robertkirkman: Do you think that if I check the `METADATA` file for `Requires-Dist:` and only proceed if the exact string `extra ==` is not present in the same line on any line that starts with `Requires-Dist:`, that will work to globally solve this?

```diff
--- a/scripts/build/termux_step_create_python_debscripts.sh
+++ b/scripts/build/termux_step_create_python_debscripts.sh
@@ -29,10 +29,10 @@ termux_step_create_python_debscripts() {
 
        # add the internal 'pip'-facing name of the package to the 'pip install' dependencies if it exists
        # in the 'METADATA' file and the 'METADATA' file also marks the package as depending
-       # on any other 'pip'-facing packages, which will install all the 'pip'-facing dependencies
+       # on any other 'pip'-facing packages, which will install all the non-optional 'pip'-facing dependencies
        # the software marks itself as depending on from PyPi that are not already installed from other Termux packages.
        # if more than one 'METADATA' file is detected, this condition will evaluate false so nothing will happen.
-       if [[ -f "$pip_metadata_file" ]] && grep -q '^Requires-Dist' "$pip_metadata_file"; then
+       if [[ -f "$pip_metadata_file" ]] && grep '^Requires-Dist' "$pip_metadata_file" | grep -v -q 'extra =='; then
                local package_pip_name="$(grep 'Name:' "$pip_metadata_file" | cut -d' ' -f2)"
                _package_python_deps+=" $package_pip_name"
        fi
```

robertkirkman: or for example is there any other syntax edge case that could fail to match like `extra==` without the space, or anything like that? That is what I worried about that prevented me from writing complicated bash parsing of the Python `METADATA` file.

DevGitPit: I am going to paste something - 

Yes, that logic is completely sound! 

Because `grep -v -q` only succeeds if there is **at least one line that does NOT match** the filter (and returns `1` if all lines match or if the input is empty), it cleanly differentiates between:
- A package that has **only** test/optional extras (all lines match `extra`, so `grep -v` yields nothing $\rightarrow$ exits `1` $\rightarrow$ debscripts skipped).
- A package that has **real runtime dependencies** (the non-extra lines don't match `extra`, so `grep -v` finds them $\rightarrow$ exits `0` $\rightarrow$ debscripts triggered).

### Syntax Edge Cases to Consider

According to [PEP 508](https://peps.python.org/pep-0508/#environment-markers):

1. **Whitespace around `==`**: While standard wheel builders (`maturin`, `flit`, `hatchling`, modern `setuptools`) typically format it with a space (`extra == '...'`), PEP 508 permits zero or more whitespace (e.g. `extra=='test'` or `extra  ==  'test'`).
2. **Quote style**: Both single `'...'` and double `"..."` quotes are valid in markers.
3. **Marker Delimiter (`;`)**: In PEP 508, environment markers strictly appear after a semicolon (`;`). Checking after `;` guarantees you are evaluating the marker section and not a package name.
4. **Alternative operators (rare)**: Very rarely, older packages used `in` (e.g. `'test' in extra` or `extra in 'test'`), but `extra ==` covers >99.9% of modern wheels.

### Recommended One-Liner

To handle arbitrary whitespace and ensure we're matching inside the marker section, you can use extended grep (`grep -E`) with POSIX character classes (which works across GNU grep, Busybox, and on-device Termux):

```diff
--- a/scripts/build/termux_step_create_python_debscripts.sh
+++ b/scripts/build/termux_step_create_python_debscripts.sh
@@ -29,10 +29,10 @@ termux_step_create_python_debscripts() {
 
        # add the internal 'pip'-facing name of the package to the 'pip install' dependencies if it exists
        # in the 'METADATA' file and the 'METADATA' file also marks the package as depending
-       # on any other 'pip'-facing packages, which will install all the 'pip'-facing dependencies
+       # on any other 'pip'-facing packages, which will install all the non-optional 'pip'-facing dependencies
        # the software marks itself as depending on from PyPi that are not already installed from other Termux packages.
        # if more than one 'METADATA' file is detected, this condition will evaluate false so nothing will happen.
-       if [[ -f "$pip_metadata_file" ]] && grep -q '^Requires-Dist' "$pip_metadata_file"; then
+       if [[ -f "$pip_metadata_file" ]] && grep '^Requires-Dist:' "$pip_metadata_file" | grep -E -v -q ';[[:space:]]*.*\<extra[[:space:]]*=='; then
                local package_pip_name="$(grep 'Name:' "$pip_metadata_file" | cut -d' ' -f2)"
                _package_python_deps+=" $package_pip_name"
        fi
```

This pattern:
- Looks strictly after `;` (the marker boundary).
- Matches `extra ==` regardless of whether spaces exist around `==` (`extra==`, `extra ==`, etc.).
- Uses standard POSIX `[[:space:]]*` so it behaves identically on host CI and on-device builds.

Test it or something.

robertkirkman: Thank you, I tested that and that's working, and after I merge this you will be able to rebase this PR to remove that line and use the global implementation to do the equivalent thing automatically, right?

- https://github.com/termux/termux-packages/pull/31552

DevGitPit: Ok.

robertkirkman: Ok now it's merged

tur/python-hf-xet/build.sh
```diff
@@ -0,0 +1,70 @@
...
+TERMUX_PKG_LICENSE="Apache-2.0"
+TERMUX_PKG_MAINTAINER="@termux-user-repository"
+TERMUX_PKG_VERSION="1.6.0"
+TERMUX_PKG_SRCURL="https://github.com/huggingface/xet-core/releases/download/v$TERMUX_PKG_VERSION/hf_xet-$TERMUX_PKG_VERSION.tar.gz"
+TERMUX_PKG_SHA256=2e58454a340b3556dfa4972d5451aff4fba8dd42a236600ba1a1d2b1514f0fef
+TERMUX_PKG_AUTO_UPDATE=true
+TERMUX_PKG_DEPENDS="libc++, python"
+TERMUX_PKG_PYTHON_COMMON_BUILD_DEPS="'maturin>=1.7,<2.0'"
+TERMUX_PKG_BUILD_IN_SRC=true
+TERMUX_PKG_PYTHON_RUNTIME_DEPS=false
```
robertkirkman: Why is this line needed here?

robertkirkman (approved):

